PR #1, opened 2015-08-31
change color of select arrow to black

jmophoto:


src/components/Select.js
```diff
@@ -212,7 +212,7 @@ const styles = {
     position: 'relative'
   },
   downArrow: {
-    color: '#999999',
+    color: '#000000',
```
bsbeeks: This should probably inherit the same color as the `selectedStyle`:
```
color: selectedStyle.color || '#000000'
```

jmophoto: Since selectedStyle is a prop, can I access it outside the Select object, though?

bsbeeks: No, so we'll need to merge the prop color onto the down arrow element here: https://github.com/jmophoto/mx-react-components/blob/jmo/select-icon-color/src/components/Select.js#L175. Something like this:
```
style={objectAssign(styles.downArrow, { color: this.props.selectedStyle.color || '#000000' })}
```

bsbeeks:
:shipit: Will merge with next release.

PR #3, opened 2015-09-23
Add donut chart component

jmophoto:
To use:

```javascript
const { DonutChart } = require('mx-react-components');

<DonutChart data=[myArray] />
```

Required params:
- data (array): Array of data objects. Example:
```javascript
[{
  name: 'Data Label Name', //string - optional, used to display the data label
  value: 10 //number - required, used to determine the pie slice size
}]
```
- activeIndex (number, -1): Only required for hover animation and data label. Used to determine which pie slice is currently active. Representing visually by making the pie slice slightly larger or zoomed in and displaying the data label.
- onMouseEnter (function): Mouse enter handler
- onMouseLeave (function): Mouse leave handler

Optional params (type, default value):
- activeOffset (number, 3): A number, in pixels, that the pie slice will increase on mouse hover.
- animateOnHover (boolean, true): Show animation on hover
- animateOnLoad (boolean, true): Show animation on load
- arcWidth (number, 80): Width of the arc in pixels
- baseArcColor (string, #E5E5E5): A css color value (hex, rgba, etc) used to set the base arc color.
- chartTotal (number): A number representing the total value of the complete pie. If no chartTotal is provided, then the sum of the data values is used. This is useful if your data values only represent a portion of the total available value.
- children (node): If children nodes are provided, they will be used in place of the default data label.
- colors (array, D3 colors): Array of colors to use for the chart
- dataPoints (array): Array of name/value pairs for data points. These will be displayed as dots along the pie.
```javascript
[{
  name: 'Data Label Name', //string - optional, not currently being used
  value: 10 //number - required, used to determine the location of the dot along the pie chart
}]
```
- dataPointColors (array, D3 colors): Array of css color values whose indexes correspond with the data array. For example, the first color value in `dataPointColors` will determine the dot color for the first item in `dataPoints`.
- dataPointRadius (number, 40): The radius, in pixels, of each `dataPoint` dot.
- height (number, 360): Height, in pixels, of the entire component.
- ~~labelStyle (object): Styling for the data label content.~~
- onClick (function): Handler for click event
- opacity (number, 1): Opacity of the pie chart
- padAngle (number, 0.01): A number, in degrees, that sets the space between pie slices.
- ~~radiusOffset (number. 0): Reduce the outer radius. Used to nest charts~~
- showBaseArc (boolean, false): Show base arc for partial data charts
- show~~Legend~~DataLabel (boolean, true): If set to `true`, then a data label will be displayed in the center of the pie on each pie slice hover, and the `zeroStateText` will be displayed.
- ~~valueFormat (object): Used to set decimal spaces and leading and trailing characters~~
- ~~valueStyle (object): Styling for the data label~~
- width (number, 360): Width of the component. Also used to calculate the radius
- ~~zeroStateText~~defaultLabel (string, 'Roll over item for details'): Text to display in legend when not hovering.

Right now, the only params needed to render the chart is the data. Without the onMouseEnter and onMouseLeave methods, the hover states do not work, though, and the legend would not be shown. Wondering if I should add that in to the component, although that would make it not stateless.

My idea about the legend is that it should be simple and only show the label and value. Anything beyond that, like what we show, can be built outside the component and positioned with the showLegend set to false.

Passing only data:
screen shot 2015-09-23 at 8 41 02 am

With other params:
screen shot 2015-09-23 at 8 40 25 am

With chart total, base arc and data points:
screen shot 2015-09-23 at 11 51 17 am


jmophoto:
TODO for Robinhood
[ ] Add base ring for 100%
[ ] Add top performer point
[ ] Animate on load

src/components/DonutChart.js
```diff
@@ -0,0 +1,234 @@
...
+    baseArcColor: React.PropTypes.string,
+    chartTotal: React.PropTypes.number,
+    children: React.PropTypes.node,
+    colors: React.PropTypes.array,
+    data: React.PropTypes.array.isRequired,
+    dataPoints: React.PropTypes.array,
+    dataPointColors: React.PropTypes.array,
+    dataPointRadius: React.PropTypes.number,
+    height: React.PropTypes.number,
+    labelStyle: React.PropTypes.object,
```
bsbeeks: We'll need to account for Radium styles.

src/components/DonutChart.js
```diff
@@ -0,0 +1,234 @@
...
+    if (this.props.showBaseArc) {
+      const radius = (Math.min(this.props.width, this.props.height) / 2) - this.props.radiusOffset;
+      const baseArc = d3.svg.arc().outerRadius(radius - this.props.activeOffset)
+                                  .innerRadius(radius - this.props.arcWidth)
+                                  .startAngle(0)
+                                  .endAngle(2 * Math.PI);
+
+      return (
+        <g
+          key={baseArc}
```
bsbeeks: Since we only have one prop, let's move this to a single line.

src/components/DonutChart.js
```diff
@@ -0,0 +1,234 @@
...
+      onMouseLeave () {},
+      opacity: 1,
+      padAngle: 0.01,
+      radiusOffset: 0,
+      showBaseArc: false,
+      showLegend: true,
+      valueFormat: {},
+      valueStyle: {},
+      width: 360,
+      zeroStateText: 'Roll over an item for details'
```
bsbeeks: Can you provide a quick explanation of each prop and what the default is what it is? This can be done with an inline comment next to each prop.

src/components/DonutChart.js
```diff
@@ -0,0 +1,234 @@
...
+
+      const endAngle = this.props.chartTotal ? valueTotal / this.props.chartTotal : 1;
+
+      const pie = d3.layout.pie().sort(null).padAngle(this.props.padAngle).endAngle(endAngle * 2 * Math.PI);
+      const values = pie(dataSets);
+      const radius = (Math.min(this.props.width, this.props.height) / 2) - this.props.radiusOffset;
+      const standardArc = d3.svg.arc().outerRadius(radius - this.props.activeOffset).innerRadius(radius - this.props.arcWidth);
+      const hoverArc = d3.svg.arc().outerRadius(radius).innerRadius(radius - this.props.arcWidth);
+
+
```
bsbeeks: Extra line break

src/components/DonutChart.js
```diff
@@ -0,0 +1,234 @@
...
+  _renderDataPoints () {
+    const dataPoints = this.props.dataPoints.map(dataPoint => {
+      return dataPoint.value;
+    });
+
+    const radius = (Math.min(this.props.width, this.props.height) / 2) - this.props.radiusOffset;
+
+    return dataPoints.map((dataPoint, index) => {
+      const endAngle = this.props.chartTotal ? dataPoint / this.props.chartTotal : dataPoint / 100;
```
bsbeeks: Instead of having a conditional here, could we set the default value of `this.props.chartTotal` to `100` in `getDefaultProps`?

src/components/DonutChart.js
jmophoto: :+1:

src/components/DonutChart.js
```diff
@@ -0,0 +1,234 @@
...
+
+      const dataPointArc = d3.svg.arc()
+        .outerRadius(radius - this.props.activeOffset)
+        .innerRadius(radius - this.props.arcWidth)
+        .startAngle(0)
+        .endAngle(endAngle * 2 * 2 * Math.PI);
+
+      const position = 'translate(' + dataPointArc.centroid() + ')';
```
bsbeeks: no need to assign this to a variable, just move the value directly into `transform` below.

src/components/DonutChart.js
```diff
@@ -0,0 +1,234 @@
...
+          <div onClick={this.props.onClick} style={styles.center}>
+            {this.props.children}
+          </div>
+        );
+      } else {
+        const labelStyle = objectAssign({
+          color: '#333',
+          fontSize: '18px',
+          fontWeight: '800'
+        }, this.props.labelStyle);
```
bsbeeks: Our components should be "Radium ready", so let's use the Radium method of combining styles. Refer to the other components to get an idea of how to do this.

src/components/DonutChart.js
jmophoto: :+1:

src/components/DonutChart.js
jmophoto: :+1:

src/components/DonutPath.js
```diff
@@ -0,0 +1,65 @@
...
+
+  getInitialState () {
+    return {
+      elapsedTime: 0,
+      endAngle: 0
+    };
+  },
+
+  componentDidMount () {
+    if (this.props.animateOnLoad) {
```
bsbeeks: This prop needs to be added to `propTypes`

src/components/DonutPath.js
```diff
@@ -0,0 +1,65 @@
...
+      clearInterval(this.interval);
+    }
+  },
+
+  _renderAnimation () {
+    let elapsedTime = this.state.elapsedTime;
+    let endAngle = this.state.endAngle;
+
+    endAngle += (elapsedTime / 5000);
```
bsbeeks: `5000` seems like a "magic number". What is it representing? Should it be a prop with a default of `5000`?

src/components/DonutPath.js
```diff
@@ -0,0 +1,65 @@
+const React = require('react');
+const d3 = require('d3');
+
+const DonutPath = React.createClass({
+  propTypes: {
+    arc: React.PropTypes.func,
+    color: React.PropTypes.string,
+    opacity: React.PropTypes.number,
+    value: React.PropTypes.object
```
bsbeeks: Can you provide explanations of these props? I also noticed there are no default props set. If that is the case, do we need to require all of the props?

src/components/DonutPath.js
jmophoto: Question on this. I am passing all the props via {...this.props}. Do I need to repeat all the propTypes, too?

bsbeeks:
This is a good first cut! After you've addressed all of my comments, be sure to run it through eslint and babel. See the Contributing section on this page for more details: http://mxenabled.github.io/mx-react-components

src/components/DonutPath.js
bsbeeks: Yes, it let's others know what props this component can expect.

src/components/DonutPath.js
jmophoto: Right now this is for the animation timing, which doesn't work. When that is finished, it will be passed as an animationDuration prop.

src/components/DonutPath.js
jmophoto: AnimateOnLoad (which is added) has a default set at the DonutChart level. Same with opacity. Would these need another default value?

The arc and value should be required. The color isn't technically needed since D3 will default to black, but since there isn't a better way to default, I would say it should be required also.

jmophoto:
Ran eslint and fixed the errors.

src/components/DonutPath.js
bsbeeks: Can we remove or comment out the animation related code until it's ready for production?

jmophoto: I removed the DonutPath component and set the DonutChart to just render the <path> itself. Also removed the animateOnLoad propType and default.

jmophoto:
I removed the animation for now and this should be good to review.

jmophoto:
- Removed prop comments
- Removed labelStyle prop
- Removed radiusOffset prop
- Removed valueStyle prop
- Removed valueFormat
- Added formatter method as a prop
- Changed showLegend to showDataLabel
- Changed zeroStateText to defaultLabelText
- Added activeIndex state
- Added handleMouseEnter to handle hover as default and then call onMouseEnter prop
- Added handleMouseLeave to handle out as default and then call onMouseLeave prop

bsbeeks:
This is ready to merge, we just need to add a DonutChart example to the demo page.

jmophoto:
Added donut chart to demo page.

jmophoto:
screen shot 2015-09-30 at 11 19 48 am 2

PR #19, opened 2015-10-01
Add Modal Component

bsbeeks:
Also passes `this.props.style` onto the Donut Chart component.

## Example
```jsx
let isOpen = true;

_handleModalClose () {
  isOpen = false;
}

<Modal
  isOpen={isOpen}
  isSmall={true}
  onRequestClose={_handleModalClose}
>
```

## Props
`isOpen` *Boolean* **optional**
Default: `false`
Determines if the modal is visible or hidden
<br/>
`isSmall` *Boolean* **optional**
Default: `false`
If set to `true`, then the modal will have a fixed width of 400px and padding of 30px.
<br/>
`onRequestClose` *Function* **optional**
A method that is called when the close button or scrim area is clicked by a user. Use this method to tell the Modal when it should close.


bsbeeks:
@jmophoto @cerinman

cerinman:
:shipit: Looks good.

PR #21, opened 2015-10-01
Move this.props.select to Select's main div

### LINKED ISSUE #6: Select Component: style prop is being applied to the wrap div element
It should be applied to the main wrapper component, but it's being applied to a child div.

bsbeeks:
Fixes #6


bsbeeks:
@jmophoto @cerinman

cerinman:
:shipit: Looks good.

PR #35, opened 2015-10-05
(fix) TypeAhead empty filter item

### LINKED ISSUE #23: TypeAhead: typing in filter name and hitting enter gives a grayed out filter with no name
Steps to duplicate:

1. Type in a letter and then hit the enter key.
2. A tag will be added to the UI with nothing in it

screenshot 2015-10-02 15 26 32

Implementation:

- [ ] Fork the repo
- [ ] run `npm install`
- [ ] run `npm run demo`. This will open up a new browser window with demos of each component. It will refresh whenever you save changes to the files in the `src` directory.
- [ ] open up `src/components/TypeAhead.js`
- [ ] On line 23, add another condition that will check if `highlightedValue` (the input value) is in `filteredItems` (the remaining available options). Hint: you'll want to use `.indexOf()`.
- [ ] Open the demo and make sure you can no longer replicate the bug.
- [ ] Do a quick run through of the TypeAhead functionality and make sure your changes didn't effect anything else.
- [ ] run `eslint src` to check for syntax and style errors. Fix the errors as needed.
- [ ] Commit your changes to the `src/` directory
- [ ] Push your changes to your fork
- [ ] Create a PR and include `Fixes #23` in the message
- [ ] Get merged and celebrate :tada:

### LINKED ISSUE #34: TypeAhead: when all options are selected if the tap key is pressed a empty filter is added
OS: Mac 0.11 (15A284)
Browser: Chrome  45.0.2454.101 (64-bit)

Steps to duplicate:
1. select all options
2. press tab to insert a blank tag (this can be done numerous times)  
screenshot 2015-10-05 15 06 50

I'll take a look and make a pull request if I can fix the issue :+1:

httpete-ire:
the TypeAhead will call the _handleItemSelect function to add
the item to the list, if the item has a value it will get pushed to
the selectedItems array, otherwise it will be ignored

closes #34

### DIFF
--- a/src/components/TypeAhead.js
+++ b/src/components/TypeAhead.js
@@ -86,9 +86,7 @@ const TypeAhead = React.createClass({
     //add to selectedItems
     const selectedItems = this.state.selectedItems;
 
-    if (item) {
-      selectedItems.push(item);
-    }
+    selectedItems.push(item);
 
     this.props.onItemSelect(item, selectedItems);
 
@@ -130,7 +128,11 @@ const TypeAhead = React.createClass({
     if (e.keyCode === 9) {
       e.preventDefault();
 
-      this._handleItemSelect(filteredItems[0]);
+      const item = filteredItems[0];
+
+      if (item) {
+        this._handleItemSelect(item);
+      }
     }
 
     //remove tag on backspace
--- a/src/components/TypeAhead.js
+++ b/src/components/TypeAhead.js
@@ -86,7 +86,9 @@ const TypeAhead = React.createClass({
     //add to selectedItems
     const selectedItems = this.state.selectedItems;
 
-    selectedItems.push(item);
+    if (item) {
+      selectedItems.push(item);
+    }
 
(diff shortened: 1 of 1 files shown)


src/components/TypeAhead.js
```diff
@@ -86,7 +86,9 @@ const TypeAhead = React.createClass({
     //add to selectedItems
     const selectedItems = this.state.selectedItems;
 
-    selectedItems.push(item);
+    if (item) {
+      selectedItems.push(item);
+    }
```
bsbeeks: Since this bug is only happening on tab, it would be better to check for the existence of `item` before we even call `_handleItemSelect`. On line 133 we are assuming that `filteredItems` has at least one value when that isn't always the case.

bsbeeks:
It looks like this is in fact a separate issue than #23. Let's leave that issue open for a first time open source contributor to complete.

httpete-ire:
as suggested, the _handleItemSelect function is only called if filteredItems[0] has a value

PR #42, opened 2015-10-05
Change default styling on donut chart

jmophoto:
New default values
- activeOffset: 0
- animateOnHover: false
- arcWidth: 10
- baseArcColor: #F5F5F5
- dataPointRadius: 5
- showBaseArc: true

Other changes
- Wrapped mouse events in conditional to only fire when animateOnHover is set to true
- Swapped position of value and text in label
- Set fontSize for label based on chart dimensions
- Set cursor to pointer only if animateOnHover is true

### DIFF
--- a/src/components/DonutChart.js
+++ b/src/components/DonutChart.js
@@ -19,6 +19,7 @@ const DonutChart = React.createClass({
     dataPointRadius: React.PropTypes.number,
     dataPoints: React.PropTypes.array,
     defaultLabelText: React.PropTypes.string,
+    defaultLabelValue: React.PropTypes.string,
     formatter: React.PropTypes.func,
     height: React.PropTypes.number,
     onClick: React.PropTypes.func,
@@ -34,14 +35,14 @@ const DonutChart = React.createClass({
   getDefaultProps () {
     return {
       activeIndex: -1,
-      activeOffset: 3,
-      animateOnHover: true,
-      arcWidth: 80,
-      baseArcColor: '#E5E5E5',
+      activeOffset: 0,
+      animateOnHover: false,
+      arcWidth: 10,
+      baseArcColor: '#F5F5F5',
       colors: d3.scale.category20().range(),
       data: [],
       dataPointColors: d3.scale.category20b().range(),
-      dataPointRadius: 40,
+      dataPointRadius: 5,
       dataPoints: [],
       defaultLabelText: 'Roll over an item for details',
       formatter (value) {
@@ -53,7 +54,7 @@ const DonutChart = React.createClass({
       onMouseLeave () {},
       opacity: 1,
       padAngle: 0.01,
-      showBaseArc: false,
+      showBaseArc: true,
       showDataLabel: true,
       width: 360
     };
@@ -80,19 +81,23 @@ const DonutChart = React.createClass({
   },
 
   _handleMouseEnter (index) {
-    this.setState({
-      activeIndex: index
-    });
+    if (this.props.animateOnHover) {
+      this.setState({
+        activeIndex: index
+      });
 
-    this.props.onMouseEnter(index);
+      this.props.onMouseEnter(index);
+    }
   },
 
   _handleMouseLeave () {
-    this.setState({
-      activeIndex: -1
-    });
+    if (this.props.animateOnHover) {
+      this.setState({
+        activeIndex: -1
+      });
 
-    this.props.onMouseLeave();
+      this.props.onMouseLeave();
+    }
   },
 
   _renderArcs () {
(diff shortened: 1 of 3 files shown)


src/components/DonutChart.js
```diff
@@ -242,16 +241,12 @@ const styles = {
     transform: 'translate(-50%, -50%)'
   },
   label: {
-    color: '#333',
-    fontSize: '14px',
-    fontWeight: '800'
-  },
-  pointer: {
-    cursor: 'pointer'
+    color: '#999',
```
bsbeeks: Can we use https://github.com/mxenabled/mx-react-components/blob/master/src/constants/Style.js#L11?

src/components/DonutChart.js
```diff
@@ -242,16 +241,12 @@ const styles = {
...
+    fontSize: '0.4em',
```
bsbeeks: Let's use 'pixels' instead of `ems`.

jmophoto: I am setting the base font size in px using the a calculation based on the chart dimensions. The ems are just used to scale down the label text. I can set them all with pixels, but I think it would make it less responsive. Thoughts?

src/components/DonutChart.js
```diff
@@ -80,19 +81,23 @@ const DonutChart = React.createClass({
...
-    this.setState({
-      activeIndex: index
-    });
+    if (this.props.animateOnHover) {
+      this.setState({
+        activeIndex: index
+      });
 
-    this.props.onMouseEnter(index);
+      this.props.onMouseEnter(index);
```
bsbeeks: Wouldn't we at least want to call this prop if animation is off? What if a user wants to get the value of the hovered slice and do something with it?

src/components/DonutChart.js
```diff
@@ -34,14 +35,14 @@ const DonutChart = React.createClass({
...
       activeIndex: -1,
-      activeOffset: 3,
-      animateOnHover: true,
-      arcWidth: 80,
-      baseArcColor: '#E5E5E5',
+      activeOffset: 0,
+      animateOnHover: false,
+      arcWidth: 10,
+      baseArcColor: '#F5F5F5',
       colors: d3.scale.category20().range(),
```
bsbeeks: Could we set the first default color as https://github.com/mxenabled/mx-react-components/blob/master/src/constants/Style.js#L3?

src/components/DonutChart.js
```diff
@@ -34,14 +35,14 @@ const DonutChart = React.createClass({
...
       data: [],
       dataPointColors: d3.scale.category20b().range(),
```
bsbeeks: Could we set the first default color as https://github.com/mxenabled/mx-react-components/blob/master/src/constants/Style.js#L7?

bsbeeks:
Since we're updated the default look of the chart, can you include screenshots?

src/components/DonutChart.js
jmophoto: Good point! Will change that.

src/components/DonutChart.js
bsbeeks: Good point :+1: .

jmophoto:
Default passing only data and dataPoints:
screen shot 2015-10-05 at 2 54 06 pm

Chart passing defaultLabelText and defaultLabelValue
screen shot 2015-10-05 at 2 56 11 pm

Chart with animation when hovering on section
screen shot 2015-10-05 at 2 56 18 pm

jmophoto:
screen shot 2015-10-05 at 3 17 54 pm

src/constants/Style.js
```diff
@@ -10,6 +10,10 @@ module.exports = {
     FONT: '#2E323F',
     LIGHT_FONT: '#999999',
 
+    BASE_ARC: '#F5F5F5',
+    DEFAULT_ARC: '#359BCF',
+    DEFAULT_DOT: '#F7B148',
```
bsbeeks: We don't need to create `DEFAULT_ARC` or `DEFAULT_DOT`. Those values are already available with `PRIMARY` and `SECONDARY`.

src/components/DonutChart.js
```diff
@@ -213,12 +210,13 @@ const DonutChart = React.createClass({
...
     const position = 'translate(' + this.props.width / 2 + ',' + this.props.height / 2 + ')';
+    const fontSize = Math.min(this.props.width, this.props.height) * 0.2 + 'px';
 
     return (
-      <div style={[styles.component, this.props.style, { height: this.props.height, width: this.props.width }]}>
+      <div style={[styles.component, this.props.style, { fontSize, height: this.props.height, width: this.props.width }]}>
         {this._renderDataLabel()}
         <svg height={this.props.height} width={this.props.width}>
-          <g style={styles.pointer} transform={position}>
+          <g style={{ cursor: this.props.animateOnHover ? 'pointer' : 'auto' }} transform={position}>
```
bsbeeks: We don't have any click events, so we shouldn't ever show a `pointer` cursor.

jmophoto: We can pass in a click event for each segment.

bsbeeks: yes, but by default we don't do anything with that click event. So it's misleading to a user if they see a `pointer` cursor, but clicking doesn't do anything. We should leave it up to the person implementing the component to determine if a `pointer` is displayed.

PR #45, opened 2015-10-06
Moved props.style to selector

jmophoto:
Moved the props.style attribute from the outer most div to the component div. As it was, there was no way to override the select border that was set in styles.component with a passed in prop.

Before:
screen shot 2015-10-06 at 3 55 54 pm

After (passing in border: none on the styles prop)
screen shot 2015-10-06 at 3 55 15 pm

### DIFF
--- a/src/components/Select.js
+++ b/src/components/Select.js
@@ -109,11 +109,11 @@ const Select = React.createClass({
     const selected = this.state.selected || this.props.selected || { displayValue: this.props.placeholderText, value: '' };
 
     return (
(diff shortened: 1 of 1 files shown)


bsbeeks:
We need to provide a way to edit the entire component's style. So we should create a new prop for the custom select styles.

jmophoto:
The outer div wraps what is shown in the component and also the fallback for the mobile version. Moving the style prop to the component div would still allow editing of the component, just not the standard mobile select. With this, are there still two props needed?

PR #49, opened 2015-10-07
Only render the DonutChart arcs if they have a value

bsbeeks:
In cases where you only 1 data object was passed in, and that data object's value was `0`, the a D3 error would occur due to the endAngle value being calculated as `Infinite`.

@jmophoto @cerinman


cerinman:
Looks good.  :shipit:

PR #50, opened 2015-10-09
Add ToggleSwitch

jmophoto:
Add ToggleSwitch component.

Available props:
- activeColor (string, '#359BCF'): Color of the active portion of the switch and the active label, if shown.
- children (node): If children nodes are provided, they will be shown in the active side of the toggle switch. For example, including an icon.
- componentStyle (array or object): Custom styling to be applied to the component
- defaultPosition (string, 'left'): Initial state of the switch. Must be either 'left' or 'right'.
- height (number 20): Height, in pixels, of the component.
- inactiveColor (string, '#F5F5F5'): Color of the inactive portion of the switch and the inactive label.
- labelStyle (array or object): Custom styling to be applied to the labels
- leftLabel (string, 'On'): Text to be displayed on the left side of the switch. Dependent on showLabels.
- onToggle (function): Method to be called on toggle. This will return the active state boolean.
- rightLabel (string, 'Off'): Text to be displayed on the right side of the switch. Dependent on showLabels.
- showLabels (boolean, true): Show left and right label text.
- togglePadding (number, 2): Amount, in pixels, of padding for the slider inside the track.
- toggleSize (number, 16): Size, in pixels of the slider. This number, plus the 2x the sliderPadding should equal the height for proper alignment. Can be bigger, with negative padding for oversized toggle.
- toggleStyle (array or object): Custom styling to be applied to the toggle.
- trackStyle (array or object): Custom styling to be applied to the track.
- width (number, 80): Width of the toggle switch, in pixels, Does not include label text.

### DIFF
--- a/src/components/ToggleSwitch.js
+++ b/src/components/ToggleSwitch.js
@@ -1,65 +1,70 @@
 const React = require('react');
 const Radium = require('radium');
-const RajaIcon = require('./RajaIcon');
 const StyleConstants = require('../constants/Style');
 
 const ToggleSwitch = React.createClass({
   propTypes: {
-    height: React.PropTypes.string,
-    iconSize: React.PropTypes.string,
-    isOn: React.PropTypes.bool,
-    offColor: React.PropTypes.string,
-    offText: React.PropTypes.string,
-    onColor: React.PropTypes.string,
-    onText: React.PropTypes.string,
+    activeByDefault: React.PropTypes.bool,
+    activeColor: React.PropTypes.string,
+    children: React.PropTypes.node,
+    height: React.PropTypes.number,
+    inactiveColor: React.PropTypes.string,
+    leftLabel: React.PropTypes.string,
     onToggle: React.PropTypes.func,
+    rightLabel: React.PropTypes.string,
+    showLabels: React.PropTypes.bool,
     style: React.PropTypes.oneOfType([
       React.PropTypes.array,
       React.PropTypes.object
     ]),
-    width: React.PropTypes.string
+    width: React.PropTypes.number
   },
 
   getDefaultProps () {
     return {
-      height: '20px',
-      iconSize: '14',
-      isOn: true,
-      offColor: StyleConstants.Colors.BASE_ARC,
-      offText: 'Off',
-      onColor: StyleConstants.Colors.PRIMARY,
-      onText: 'On',
+      activeByDefault: true,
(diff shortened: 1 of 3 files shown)


jmophoto:
- Removed font props since they can be passed with the styles prop.
- Set font size and family in component CSS.

screen shot 2015-10-09 at 1 10 05 pm

jmophoto:
I kept the activeByDefault as a boolean instead of left/right. My thought is that it may not always be left or right (up/down) and on/off is similar enough to true/false and the boolean doesn't require a === check.

This should be ready for final review. @bsbeeks @cerinman

jmophoto:
Correction. Final review for this iteration. New styling is in the works.

bsbeeks:
`activeByDefault` doesn't make sense to me because `active` is arbitrary. What if I set the right label text as `on` and the left one as `off`? Then the "active" state would display the toggle in the "off" position, which is confusing. I don't see anything wrong with using `defaultPosition` as the prop name and passing in `left` or `right` as values. We can use `propTypes` to limit the prop to just those two values.

jmophoto:
With that use case in mind, I agree. Will make that change.

jmophoto:
This now has updated default styling.
Also added styling props for component, toggle, slider and label.

Default
screen shot 2015-10-13 at 9 57 24 am

Passing icon as children
screen shot 2015-10-13 at 9 58 05 am

src/components/ToggleSwitch.js
```diff
@@ -0,0 +1,172 @@
...
+    React.PropTypes.array,
+    React.PropTypes.object
+  ]),
+  leftLabel: React.PropTypes.string,
+  onToggle: React.PropTypes.func,
+  rightLabel: React.PropTypes.string,
+  showLabels: React.PropTypes.bool,
+  sliderInset: React.PropTypes.number,
+  sliderSize: React.PropTypes.number,
+  sliderStyle: React.PropTypes.oneOfType([
```
bsbeeks: instead of `slider` let's call this `track`

jmophoto: Just for clarification, should the gray area or the circle that toggles back and forth be the track? I wondered about the naming in this.

bsbeeks: I say the toggle is the circle and the gray area is the track.

jmophoto:
Agree on the naming. Changes have been pushed up.

PR #51, opened 2015-10-13
Change bind from prop to local method

jmophoto:
There is warning thrown when binding to a function passed as a prop. From what I can tell, you can't bind 'this' between props. This adds a method to the local component which then calls the prop.

Would be interested on your thoughts on this @bsbeeks @cerinman 

http://stackoverflow.com/questions/30477042/react-js-how-to-bind-passed-in-event-handlers-this-to-child-component

### DIFF
--- a/src/components/DonutChart.js
+++ b/src/components/DonutChart.js
@@ -26,6 +26,10 @@ class DonutChart extends React.Component {
     }
   }
 
+  _handleClick (index) {
+    this.props.onClick(index);
+  }
+
   _handleMouseEnter (index) {
     if (this.props.animateOnHover) {
       this.setState({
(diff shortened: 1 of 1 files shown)


bsbeeks:
For reference: With the upgrade to React 0.14, event call back methods are not auto bound, which is why we added `.bind(this)` to all the event methods.

src/components/DonutChart.js
```diff
@@ -71,7 +71,7 @@ class DonutChart extends React.Component {
           return (
             <g
               key={i}
-              onClick={this.props.onClick.bind(this, i)}
+              onClick={this._handleClick.bind(this, i)}
```
bsbeeks: `this._handleClick` isn't defined in this component.

jmophoto: Yeah... that might be good to have. Added and pushed that up. Should be ready for review.

PR #53, opened 2015-10-15
Change toggleSize to toggleHeight and toggleWidth

jmophoto:
This will allow for rectangular toggles.

Both default to 16

@bsbeeks 

screen shot 2015-10-15 at 5 02 48 pm

### DIFF
--- a/src/components/ToggleSwitch.js
+++ b/src/components/ToggleSwitch.js
@@ -81,11 +81,11 @@ class ToggleSwitch extends React.Component {
         fontWeight: 'bold'
       },
       left: {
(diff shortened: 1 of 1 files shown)


bsbeeks:
travis is failing

jmophoto:
Would help if I knew my ABCs!

jmophoto:
Took out the height, width and padding props since they can be passed in as props.

@bsbeeks

jmophoto:
I went ahead and changed the default styles to the pattern above. Makes it less brittle.

PR #55, opened 2015-10-16
Keep the context of RangeSelector on window resize, Fixes #48

### LINKED ISSUE #48: Getting error on RangeSelector component when resizing window
Whenever you resize the window when displaying a RangeSelector component you get the following error: `Uncaught TypeError: Failed to execute 'getComputedStyle' on 'Window': parameter 1 is not of type 'Element'.`

tegon:
The callback of `window.addEventListener` changes the context of the function.
The `_bind` sends the component context to the callback, since there is no option to send the context in `_throttle`. ref https://github.com/lodash/lodash/issues/699

### DIFF
--- a/src/components/RangeSelector.js
+++ b/src/components/RangeSelector.js
@@ -2,7 +2,6 @@ const React = require('react');
 const ReactDOM = require('react-dom');
 const Radium = require('radium');
 const _throttle = require('lodash/function/throttle');
-const _bind = require('lodash/function/bind');
 
 const StyleConstants = require('../constants/Style');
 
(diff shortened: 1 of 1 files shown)


src/components/RangeSelector.js
```diff
@@ -26,11 +27,11 @@ class RangeSelector extends React.Component {
   componentDidMount () {
     this._setDefaultRangeValues();
 
-    window.addEventListener('resize', _throttle(this._setDefaultRangeValues, 300));
+    window.addEventListener('resize', _throttle(_bind(this._setDefaultRangeValues, this), 300));
```
bsbeeks: Any reason why we can't use the built in `.bind` method?
```javascript
_throttle(this._setDefaultRangeValues.bind(this), 300
```

tegon: I was gonna say that was because of browser support, since `Function.prototype.bind` doesn't exists on IE7 and IE8. 
But then I saw in the `lodash` docs that their support is for IE9 or greater, then at the end it made no difference..
I'll change to use native `.bind` instead.

bsbeeks: Also, for our components we focus on supporting IE10+

tegon: Nice! That's really better. 
Just pushed the change.

Issue #30, opened 2015-10-02
Address Accessibility


bsbeeks:

psigns:
Hi Brandon,

I see that this issue already has an assignee, but I'm really interested in learning more about accessibility.  Do you mind if I work on some of the components and make some pull requests?

bsbeeks:
@psigns we'd love some help! Feel free to put up as many PR's as you'd like :+1:

psigns:
Just checking in on this.  I've made the Range Selector, Icon, Toggle Switch, and DonutChart more accessible.  I'm going to work on the modals next, then I'll spend tomorrow working on the pull request.  I'll get to the TypeAhead component after the first pull request, because I can tell that its going to take some time.

bsbeeks:
@psigns nice! Could you put up some PRs with the changes so we can start reviewing them? If possible, separate each component into it's own PR.

cerinman:
This issue is very old and very broad.  Closing for now.  We can open individual issues for components as needed.

PR #56, opened 2015-10-19
#30 - Added aria roles and keyboard navigation to Toggle Switch.

psigns:
This pull request adds aria roles and keyboard navigation to make Toggle Switch component conform to Aria 'switch role.'  See the following links for the description of the role.  Users can use the tab button to give the Toggle Switch focus and the space bar to toggle its position.

[MDN - Aria Switch Role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_switch_role)
[MDN - Aria Checkbox Role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_checkbox_role)

## Example
```html
<p id="email-updates-label">
Would you like you receive updates?
</p>
<ToggleSwitch
  ariaCheckedPosition='right'
  ariaDescribedByElement='email-updates-label'
  leftLabel='No'
  rightLabel='Yes'
>
```

## Props
`ariaCheckedPosition` *string*  **optional**
Default: 'left'
Used to set which direction should be considered "aria-checked=true" for screen readers.  Also used to determine which label should be used to describe the switch to screen readers.

`ariaDescribedByElement` *string* **optional**
This string is the id of an html element (with no hash) that provides additional information to the user about what the switch indicating.  See the above example.

`id` *string* **optional**
An id attribute for the Toggle Switch.  If one is not provided, the component will generate its own unique id so that it does not conflict with other Toggle Switches on the page.  The id is used to associate the Toggle Switch with its labels.

### DIFF
--- a/src/components/ToggleSwitch.js
+++ b/src/components/ToggleSwitch.js
@@ -1,6 +1,8 @@
+const KeyCodes = require('../constants/KeyCodes');
 const React = require('react');
 const Radium = require('radium');
 const StyleConstants = require('../constants/Style');
+const _uniqueId = require('lodash').uniqueId;
 
 class ToggleSwitch extends React.Component {
   constructor (props) {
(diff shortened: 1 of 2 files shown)


psigns:
I'm not sure why this is failing.  Eslint checked out on my end and I didn't get any errors when I ran npm test.

bsbeeks:
No worries, we can take a look at why it's failing. It could be a configuration issue.

cerinman:

Issue #26, opened 2015-10-02
New Component: Date Picker


bsbeeks:

cerinman:
This is already done.  Just need to port it over.  Next week I should have time to get it moved.

cerinman:
This is complete but did not get closed.

PR #57, opened 2015-10-19
DatePicker component added.

tumentumurchudur:
DatePicker component added.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
   "dependencies": {
     "d3": "^3.5.6",
     "lodash": "^3.10.1",
+    "moment": "^2.10.3",
     "object-assign": "3.x.x",
     "radium": "^0.14.1",
     "react": "^0.14.0",
@@ -37,11 +38,10 @@
     "babel-core": "^5.6.15",
     "babel-eslint": "^4.1.3",
     "babel-loader": "^5.3.1",
-    "eslint": "^1.5.0",
+    "eslint": "^1.7.2",
     "eslint-plugin-react": "^3.4.2",
     "eslint-watch": "^2.1.1",
     "live-server": "^0.8.1",
-    "moment": "^2.10.3",
     "node-libs-browser": "^0.5.2",
     "webpack": "^1.10.1"
   }
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "eslint-plugin-react": "^3.4.2",
     "eslint-watch": "^2.1.1",
     "live-server": "^0.8.1",
+    "moment": "^2.10.3",
     "node-libs-browser": "^0.5.2",
     "webpack": "^1.10.1"
   }
--- a/src/components/DatePicker.js
+++ b/src/components/DatePicker.js
@@ -11,7 +11,7 @@ class DatePicker extends React.Component {
     super(props);
     this.state = {
       currentDate: null,
-      selectedDate: moment(this.props.selectedDate).unix(),
+      selectedDate: moment(this.props.defaultDate).unix(),
       showCalendar: false
     };
   }
@@ -171,7 +171,7 @@ class DatePicker extends React.Component {
         <Icon
           onClick={this._toggleCalendar.bind(this)}
           size='20px'
-          style={[styles.navIcon, styles.navRight]}
+          style={styles.caret}
           type='caret-up'
         />);
     } else {
@@ -179,7 +179,7 @@ class DatePicker extends React.Component {
         <Icon
           onClick={this._toggleCalendar.bind(this)}
           size='20px'
-          style={[styles.navIcon, styles.navRight]}
+          style={styles.caret}
           type='caret-down'
         />);
     }
@@ -229,13 +229,13 @@ class DatePicker extends React.Component {
 
 DatePicker.propTypes = {
   closeOnDateSelect: React.PropTypes.bool,
+  defaultDate: React.PropTypes.string,
   fontSize: React.PropTypes.string,
   format: React.PropTypes.string,
   locale: React.PropTypes.string,
   minimumDate: React.PropTypes.string,
   onDateSelect: React.PropTypes.func,
   scrimStyle: React.PropTypes.object,
-  selectedDate: React.PropTypes.string,
   showCaret: React.PropTypes.bool,
   showDayBorders: React.PropTypes.bool,
   title: React.PropTypes.string,
@@ -245,12 +245,12 @@ DatePicker.propTypes = {
 
 DatePicker.defaultProps = {
   closeOnDateSelect: false,
+  defaultDate: moment().unix(),
   fontSize: StyleConstants.FontSize,
   format: 'YYYY-MM-DD',
   locale: 'en',
   onDateSelect () {},
   scrimStyle: {},
-  selectedDate: moment().unix(),
   showCaret: true,
   showDayBorders: false,
   title: null,
@@ -259,6 +259,13 @@ DatePicker.defaultProps = {
 };
 
 const styles = {
+  caret: {
+    color: '#CCCCCC',
+    position: 'absolute',
+    right: '0',
+    top: '50%',
+    transform: 'translateY(-50%)'
+  },
   calendar: {
     borderTopColor: '#E5E5E5',
     borderTopStyle: 'solid',
@@ -294,12 +301,12 @@ const styles = {
   },
   calendarDayContent: {
     borderRadius: '50%',
-    height: '100%',
+    height: '32px',
     left: '50%',
     position: 'absolute',
     top: '50%',
     transform: 'translateY(-50%) translateX(-50%)',
-    width: '90%',
+    width: '32px',
 
     ':hover': {
       backgroundColor: '#359BCF',
@@ -323,6 +330,7 @@ const styles = {
     }
   },
   calendarHeader: {
+    color: '#606060',
     borderBottom: '#E5E5E5',
     borderBottomWidth: '1px',
     fontSize: '15px',
@@ -349,6 +357,7 @@ const styles = {
   component: {
     backgroundColor: '#FFFFFF',
     color: '#000000',
+    fontFamily: StyleConstants.FontFamily,
     height: '100%',
     WebkitAppearance: 'none',
     width: '100%',
@@ -396,7 +405,7 @@ const styles = {
     color: '#FFFFFF'
   },
   currentMonth: {
-    color: '#000000'
+    color: '#606060'
   },
   input: {
     backgroundColor: '#FFFFFF',
@@ -414,7 +423,6 @@ const styles = {
     }
   },
   navIcon: {
-    color: '#000000',
     cursor: 'pointer'
   },
   navLeft: {
@@ -441,7 +449,7 @@ const styles = {
     color: '#606060',
     cursor: 'pointer',
     fontSize: StyleConstants.FontSize,
-    fontWeight: 'bold',
+    padding: '5px 0 5px 5px',
     verticalAlign: 'middle',
     width: '100%',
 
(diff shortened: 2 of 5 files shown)


cerinman:
completes https://github.com/mxenabled/mx-react-components/issues/26

bsbeeks:
Please provide a writeup similar to https://github.com/mxenabled/mx-react-components/pull/3

bsbeeks:
Also, new components should use ES6 classes. See the other components for reference.

tumentumurchudur:
Sounds good. I will rewrite the component using ES6 classes and add some documentation.

tumentumurchudur:
DatePicker has been rewritten using ES6 classes. Please see the first draft of my documentation for the component and its parameters.

To use:
  const { DatePicker } = require('mx-react-components');
  <DatePicker onDateSelect={callbackFunction} selectedDate={Currently selected date}/>

There are no required parameters for this component but to make it useful, you need to provide a callback function that is called when a date is selected on the calendar as well as the parameter selectedDate to keep track of the currently selected date as a state in your component. 

Parameter Descriptions:
- calendarColumns (number, 7): A number of days to show in each column on the calendar. 
- closeOnDateSelect (boolean, false): Close the calendar automatically once a date is selected.
- divideCells (boolean, false):  Show border around each day on the calendar.
- fontSize (string, '1em'): Font size of the text and days.
- format (string, 'YYYY-MM-DD'): Date format of the calendar.
- locale (string, 'en'): Calendar locale
- minimumDate (string, null): Minimum date you can select on the calendar. There is no minimum date when it is null.
- onDateSelect (func, {}): Callback function on a date select.
- scrimStyle (object, {}): A style object that changes the styles of scrim. 
- selectedDate (number): Currently selected date on the calendar.
- useInputForSelectedDate (boolean, true): Allows the user to directly type in a date.
- useScrim (boolean, false): Show scrim

screen shot 2015-10-19 at 4 08 43 pm

package.json
```diff
@@ -41,6 +41,7 @@
     "eslint-plugin-react": "^3.4.2",
     "eslint-watch": "^2.1.1",
     "live-server": "^0.8.1",
+    "moment": "^2.10.3",
```
bsbeeks: If the component uses moment, then this should be in `dependencies`.

src/components/DatePicker.js
```diff
@@ -0,0 +1,454 @@
...
+
+  _handlePreviousClick () {
+    const selectedDate = moment.unix(this.props.selectedDate).locale(this.props.locale);
+    let currentDate = this.state.currentDate ? this.state.currentDate.locale(this.props.locale) : selectedDate;
+
+    currentDate = moment(currentDate.startOf('month').subtract(1, 'm'), this.props.format);
+
+    this.setState({
+      currentDate
+    });
```
bsbeeks: Should clicking the next/prev buttons actually change the selected date? I think a better UX would be to change the month, but not select a date. Leaving it up to the user to actually select the date they want.

src/components/DatePicker.js
```diff
@@ -0,0 +1,454 @@
...
+
+  _handleBlur () {
+    this.setState({
+      showCalendar: false
+    });
+  }
+
+  render () {
+    const selectedDate = moment.unix(this.props.selectedDate).locale(this.props.locale);
+    const currentDate = this.state.currentDate ? this.state.currentDate.locale(this.props.locale) : selectedDate;
```
bsbeeks: We are defining these two variables the same way a few times throughout the file. Perhaps we should be doing this in `constructor` or `componentDidMount`?

src/components/DatePicker.js
```diff
@@ -0,0 +1,454 @@
...
+      },
+      title: {
+        backgroundColor: '#666666',
+        color: '#f2f2f2',
+        textAlign: 'center',
+        padding: '7px 0px 7px 0px',
+        fontSize: '1em',
+        margin: '1px'
+      }
+    };
```
bsbeeks: Could we move this `styles` object outside of the React class? In the places we are using `props` to set styles, we can modify the styles object or use Radium to our advantage and do something like
```
style={[styles.calendarHeader, { borderBottomStyle: this.props.divideCells ? 'solid' : 'none'}]}
```
This would also make it so we don't have to pass the `styles` variable to the `_renderMonthlyTable` method.

src/components/DatePicker.js
```diff
@@ -0,0 +1,454 @@
...
+          <input
+            key='input'
+            onClick={this._toggleCalendar.bind(this)}
+            style={styles.input}
+            type='text'
+            value={moment.unix(this.props.selectedDate).format(this.props.format)}
+          />
+          <Icon
+            onClick={this._toggleCalendar.bind(this)}
+            size='1.75em'
```
bsbeeks: We try to avoid using `em`. Could we change this to a pixel size?

src/components/DatePicker.js
```diff
@@ -0,0 +1,454 @@
...
+        </div>
+      );
+    } else {
+      return (
+        <div
+          key='selectedDate'
+          onClick={this._toggleCalendar.bind(this)}
+          style={styles.selectedDate}
+        >
+          {moment.unix(this.props.selectedDate).format(this.props.format)}
```
bsbeeks: I've noticed that we are relying on the user to pass in the `selectedDate` as a prop. To follow suit with our other components, we should make sure the component works independent of props being passed in. In this specific case, we should be keeping track of the selected date as a property on the state. Whenever it is changed we can expose it back to the user in a method prop, then they can do what they want with it.

src/components/DatePicker.js
```diff
@@ -0,0 +1,454 @@
...
+  }
+
+  _renderTitle (styles) {
+    if (this.props.title) {
+      return (
+        <div key='title' style={styles.title}>
+          {this.props.title}
+        </div>
+      );
+    } else {
```
bsbeeks: You should be able to remove this `else` statement all together. If there is no title, then we don't return anything.

src/components/DatePicker.js
bsbeeks: I just pulled down and started using the component. It appears that this is functioning as I described above. The confusion came from the naming of the property.

src/components/DatePicker.js
```diff
@@ -0,0 +1,454 @@
...
+        borderTopWidth: '1px',
+        display: 'none',
+        padding: '10px 0'
+      },
+      calendarDay: {
+        color: '#DDDDDD',
+        float: 'left',
+        paddingBottom: 100/(this.props.calendarColumns + 2) + '%',
+        position: 'relative',
+        width: 100/(this.props.calendarColumns + 1) + 1 + '%'
```
bsbeeks: Could we somehow use css flexbox to simplify the day columns and week rows?

src/components/DatePicker.js
```diff
@@ -0,0 +1,454 @@
...
+        {this._renderScrim(styles)}
+      </div>
+    );
+  }
+}
+
+DatePicker.propTypes = {
+  calendarColumns: React.PropTypes.number,
+  closeOnDateSelect: React.PropTypes.bool,
+  divideCells: React.PropTypes.bool,
```
bsbeeks: A better name for this might be, `showDayBorders`?

src/components/DatePicker.js
```diff
@@ -0,0 +1,454 @@
...
+  _renderMonthTable (styles, currentDate, selectedDate) {
+    const days = [];
+    const startDate = moment(currentDate, this.props.format).startOf('month').startOf('week');
+    const endDate = moment(currentDate, this.props.format).endOf('month').endOf('week');
+    const minimumDate = this.props.minimumDate ? moment(this.props.minimumDate, this.props.format) : null;
+
+    while (startDate.isBefore(endDate)) {
+      const isCurrentMonth = startDate.month() === currentDate.month();
+      const isCurrentDay = startDate.format(this.props.format) === selectedDate.format(this.props.format);
+      const day = startDate.isBefore(minimumDate) ? (
```
bsbeeks: This should probably be a standard if/else conditional instead of a ternary.

src/components/DatePicker.js
```diff
@@ -0,0 +1,454 @@
...
+          </div>
+        </div>
+      );
+
+      days.push(day);
+      startDate.add(1, 'd');
+    }
+
+    //Divide calendar columns/Rows
+    if (this.props.divideCells) {
```
bsbeeks: We are already looping through the days above in the `while` statement. Is there a way we can move this logic into that statement? At first glance it looks like we could use Radium to apply the appropriate styles.

bsbeeks:
This is looking good and we're definitely on the right track. Other things that need to be addressed are:
- Add CSS classes on all the elements that would make sense to style
- Make sure eslint is passing
- Make sure it's accessible (Here is an example: http://accessibility.athena-ict.com/aria/examples/datepicker2.shtml)

src/components/DatePicker.js
tumentumurchudur: Yes, they are little repetitive. We could declare them in constructor as this.selectedDate and this.currentDate but it won't reduce the # of lines of code and these variables would become global. Dealing with global variables maybe little risky. Whereas, const variables are block-scoped and only used inside that function and destroyed immediately.

src/components/DatePicker.js
tumentumurchudur: When I change the number of calendarColumns, the calendar does not look right. I think we should remove the prop calendarColumns all together.  Most calendars have a 7 day column anyway.

bsbeeks: I agree, let's hard code it to 7 days per row

tumentumurchudur:
- Minor details have been modified as discussed
- Ran eslint and passed
- Touched up the UI and made it look as close as possible to the expected image above
- What css classes do you want added back? 

Now working on making the component accessible
screen shot 2015-10-20 at 3 54 01 pm

tumentumurchudur:
The following changes have been made:
- circle background when a date is selected
- added a caret
- used font from StyleConstants
- removed space between the calendar and input
- used shadow from StyleConstants
- added props for default selected date and caret
- centered selected date inside input

src/components/DatePicker.js
```diff
@@ -0,0 +1,462 @@
...
+
+DatePicker.propTypes = {
+  closeOnDateSelect: React.PropTypes.bool,
+  fontSize: React.PropTypes.string,
+  format: React.PropTypes.string,
+  locale: React.PropTypes.string,
+  minimumDate: React.PropTypes.string,
+  onDateSelect: React.PropTypes.func,
+  scrimStyle: React.PropTypes.object,
+  selectedDate: React.PropTypes.string,
```
bsbeeks: This should be named `defaultDate` since once a user starts interacting with the date picker we stop using it. It will also help us differentiate between the `selectedDate` parameter on the state.

bsbeeks:
We're getting closer. A couple things I'm noticing:
- The font is still a serif font. It should be a sans-serif font using `StyleConstants. FontFamily`
- The padding and height of the selected date area doesn't match the expected design
- The circle indicator is still slightly oval, it should be a perfect circle
- The font color of the selected date looks correct, but the calendar font color appears darker
- The caret color in the selected date area is too dark

tumentumurchudur:
Modified font, padding, height, font color and faded carets.
screen shot 2015-10-22 at 10 53 45 am

bsbeeks:
Looks like Travis is failing due to a trailing comma. Once that is fixed, this will be good to go.

src/constants/Style.js
```diff
@@ -19,5 +19,6 @@ module.exports = {
   },
 
   FontSize: '13px',
-  FontFamily: 'Helvetica, Arial, sans-serif'
+  FontFamily: 'Helvetica, Arial, sans-serif',
+  BoxShadow: '0 30px 30px 10px rgba(0,0,0,0.1)',
```
bsbeeks: the culprit

tumentumurchudur:
Yup, it compiles now. Thanks

PR #58, opened 2015-10-19
#30 - Make Icon component accessible

psigns:
This PR adds `<title>` and `<desc>` elements to SVG icons to make them readable by screen readers.

I followed the recommendations in [this article](http://www.sitepoint.com/tips-accessible-svg/) for accessible SVG.

## Props
`id` **string** *optional*
Used to associate the SVG with its title and desc elements through the aria-labelledby attribute.  If no id is provided, the component will generate a unique id so that multiple icons can be used on the page.


cerinman:

PR #59, opened 2015-10-19
Added aria attributes and keyboard navigation to RangeSelector.

psigns:
This PR makes the RangeSelector accessible by allowing for keyboard navigation and adding labels and aria attributes.

The range selector conforms to the [slider widget](http://www.w3.org/TR/wai-aria-practices/#slider) documentation.  The range selector presets conform to the [menu button](http://www.w3.org/TR/wai-aria-practices/#menubutton) and [menu](http://www.w3.org/TR/wai-aria-practices/#menu) documentation.

## Props
`descriptionElementId` *string* **optional**
The element that provides additional descriptive information about what the range selector selects.

`id` *string* **optional**
A unique id, used to establish aria relationships between elements in the component.  If no id is provided, the component will create its own unique id.

`pageUpDownInterval` *number* **optional**
A number that sets the interval that the Range Selector toggles should change when the user presses 'Page Up' or 'Page Down' buttons.  The w3 accessible slider widget document recommends that the 'Page Up' and 'Page Down' buttons change the slider values faster than the arrow keys.  Defaults to 10.

### DIFF
--- a/src/components/RangeSelector.js
+++ b/src/components/RangeSelector.js
@@ -2,8 +2,11 @@ const React = require('react');
 const ReactDOM = require('react-dom');
 const Radium = require('radium');
 const _throttle = require('lodash/function/throttle');
+const _uniqueId = require('lodash').uniqueId;
 
 const StyleConstants = require('../constants/Style');
+const KeyCodeConstants = require('../constants/KeyCodes');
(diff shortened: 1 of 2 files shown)


cerinman:

PR #60, opened 2015-10-20
Change select to use activeIndex instead of :hover

jmophoto:
Change the select to use activeIndex to determine hover state instead of psuedoclass.

@bsbeeks @cerinman

### DIFF
--- a/src/components/Select.js
+++ b/src/components/Select.js
@@ -11,7 +11,7 @@ class Select extends React.Component {
   constructor (props) {
     super(props);
     this.state = {
-      activeIndex: this.props.options.indexOf(this.props.selected),
+      activeIndex: false,
       isOpen: false,
       selected: false
     };
@@ -19,6 +19,7 @@ class Select extends React.Component {
 
   _handleBlur () {
     this.setState({
+      activeIndex: false,
       isOpen: false
     });
   }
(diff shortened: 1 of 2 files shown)


jmophoto:
This should also have a hover color or style. Adding that now.

jmophoto:
Default:
screen shot 2015-10-20 at 1 21 01 pm

Passing red background color and green color:
```javascript
{{ backgroundColor: 'red', color: 'green' }}
```
screen shot 2015-10-20 at 1 20 50 pm

src/components/Select.js
```diff
@@ -11,6 +11,7 @@ class Select extends React.Component {
   constructor (props) {
     super(props);
     this.state = {
+      activeIndex: this.props.options.indexOf(this.props.selected),
```
bsbeeks: I don't think we'd want to set this by default. The hover styles should only display if you are actually hovering over an item.

jmophoto:
- Set initial state of activeIndex to -1
- Set activeIndex to -1 on click
- Set activeIndex to -1 onMouseLeave

bsbeeks:
There has to be a better way to solve this. Let's spend some more time on `.blur()` and maybe using different mouse events.

bsbeeks:
@jmophoto where are we at with this?

jmophoto:
@bsbeeks: will revive this!

bsbeeks:
@jmophoto let's wrap this up or close it.

lukeschunk:
my recent PR might address this - see what you guys think

bsbeeks:
@lukeschunk agreed. Closing this in favor of your PR.

PR #61, opened 2015-10-20
Add space between RangeSelector group label and selected line

bsbeeks:
Before:
screenshot 2015-10-20 15 05 12

After:
screenshot 2015-10-20 15 04 54

@jmophoto @cerinman

### DIFF
--- a/src/components/RangeSelector.js
+++ b/src/components/RangeSelector.js
@@ -148,7 +148,7 @@ class RangeSelector extends React.Component {
         display: this.state.showPresets ? 'block' : 'none'
       },
       range: {
-        padding: '44px 0 30px',
+        padding: '45px 0',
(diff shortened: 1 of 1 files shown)


src/components/RangeSelector.js
```diff
@@ -148,7 +148,7 @@ class RangeSelector extends React.Component {
         display: this.state.showPresets ? 'block' : 'none'
       },
       range: {
-        padding: '44px 0 30px',
+        padding: '45px 0',
```
cerinman: Does this work with the new radium and short hand issue?

bsbeeks: Yes, it's only an issue if you mix shorthand and longhand in the same object.

cerinman:
:shipit:

PR #63, opened 2015-10-22
Time based split line chart

cerinman:
@bsbeeks @jmophoto

Adds new TimeBasedLineChart component.

To use:
```javascript
const  { TimeBasedLineChart } = require('mx-react-components');
```

```html
<TimeBasedLineChart 
  data={myData}
  hoverCallBack={myHoverCallBackFunction}
>
  <div>
    {Hover data from call back}
  </div>
</TimeBasedLineChart>
```

There are no required props for this component to work but the prop `data` will need to be passed if you want the graph to draw.

The data prop should be formatted as follows:

```javascript
data = [
  {
    timeStamp: 14093498, //unix time stamp for data point REQUIRED
    value: 100 //value to be displayed at the corresponding time stamp REQUIRED
  }
];
```

Props Descriptions:

-  areaBelowZeroColor (string, #EE4235): A hex color used to shade the area below zero on the line chart. 

-  alwaysShowZeroYTick (boolean, false): Zero is not always shown as a tick on the y axis.  This forces it to be drawn if it does not exist.

-  breakPointDate (unix time stamp, today's date): The location on the chart to draw the break point line and associated label

-  breakPointLabel (string, Today): The label to be drawn above the break point line on the chart.

-  children (node, null): Any child DOM elements added to the component at time of use.

-  dashedFutureLine (boolean, true): Will draw the part of the line on the graph that is in the future as a dashed line.

-  data (array, []): An array of data point objects used to draw the graph.

-  height (number, 400): The height of the graph.

-  hoverCallBack (function, () => {}): The function that is used as a means of returning the hover state to the user of the component.

-  lineColor (string, #359BCF): The color used to draw the line.

-  margin (object, { top: 20, right: 50, bottom: 20, left: 50 }): The margins used for the graph.

-  rangeType (string, day): The chart is set to render data points as days by default but you can have it render data points as months if you switch rangeType to month.

-  shadeAreaBelowZero (boolean, false): Shade the area below zero on the graph.

-  showBreakPoint (boolean, true): Hide or show the break point and it's label on the graph.

-  showTooltips (boolean, true): Hide or show tool tips on hover over data points in the graph.

-  staticXAxis (boolean, true): Makes the x axis ticks on the graph be created and drawn as static.  If set to false then the labels for the x axis are hidden until you hover over a data point on the graph.

-  width (number, 550): The width of the graph.

-  yAxisFormatter (function, (value) => {return numeral(value).format('0a')}): A function used to format y axis ticks on the graph.

Built in classes for styling:
mx-time-based-line-chart-below-zero-area: The shaded area below zero on chart
mx-time-based-line-chart-break-point-date: The break point date
mx-time-based-line-chart-break-point-label: The break point label
mx-time-based-line-chart-break-point-line: The break point line
mx-time-based-line-chart-dot: Static dots on the line
mx-time-based-line-chart-hover-dot: The dot drawn above a data point on hover
mx-time-based-line-chart-hover-rectangle: The grey rectangle shown for a data point on hover
mx-time-based-line-chart-line: Chart line
mx-time-based-line-chart-svg: Chart svg
mx-time-based-line-chart-tool-tip-wrapper: Tooltip div wrapper
mx-time-based-line-chart: Chart div wrapper

screen shot 2015-10-22 at 12 14 49 pm

### DIFF
--- a/src/components/TimeBasedLineChart.js
+++ b/src/components/TimeBasedLineChart.js
@@ -46,7 +46,7 @@ class TimeBasedLineChart extends React.Component {
   _getAreaBelowZero (data) {
     const area = d3.svg.area()
       .x(d => {
-        const currentDate = moment.unix(d.timeStamp).startOf('day').unix()
+        const currentDate = moment.unix(d.timeStamp).startOf(this.props.rangeType).unix()
 
         return this._getXScaleValue(currentDate);
       })
@@ -59,7 +59,7 @@ class TimeBasedLineChart extends React.Component {
   _getFlatLine (data) {
     const flatLine = d3.svg.line()
       .x(d => {
-        const currentDate = moment.unix(d.timeStamp).startOf('day').unix()
+        const currentDate = moment.unix(d.timeStamp).startOf(this.props.rangeType).unix()
 
         return this._getXScaleValue(currentDate);
       })
@@ -71,7 +71,7 @@ class TimeBasedLineChart extends React.Component {
   _getLine (data) {
     const line = d3.svg.line()
       .x(d => {
-        const currentDate = moment.unix(d.timeStamp).startOf('day').unix()
+        const currentDate = moment.unix(d.timeStamp).startOf(this.props.rangeType).unix()
 
         return this._getXScaleValue(currentDate);
       })
@@ -87,7 +87,7 @@ class TimeBasedLineChart extends React.Component {
   }
 
   _getSliceWidth (timeStamp) {
-    const rangeType = this.props.rangeType === 'year' ? 'month' : 'day';
+    const rangeType = this.props.rangeType;
     const startOf = timeStamp ? moment.unix(timeStamp).startOf(rangeType).unix() : moment(timeStamp).startOf(rangeType).unix();
     const endOf = timeStamp ? moment.unix(timeStamp).endOf(rangeType).unix() : moment(timeStamp).endOf(rangeType).unix();
 
@@ -99,17 +99,17 @@ class TimeBasedLineChart extends React.Component {
     const past = [];
 
     this.props.data.forEach(item => {
-      const currentDate = moment.unix(item.timeStamp).startOf('day');
+      const currentDate = moment.unix(item.timeStamp).startOf(this.props.rangeType);
 
-      if (currentDate.isBefore(moment.unix(this.props.breakPointDate), 'day')) {
+      if (currentDate.isBefore(moment.unix(this.props.breakPointDate), this.props.rangeType)) {
         past.push(item);
       }
 
-      if (currentDate.isAfter(moment.unix(this.props.breakPointDate), 'day')) {
+      if (currentDate.isAfter(moment.unix(this.props.breakPointDate), this.props.rangeType)) {
         future.push(item);
       }
 
-      if (currentDate.isSame(moment.unix(this.props.breakPointDate), 'day')) {
+      if (currentDate.isSame(moment.unix(this.props.breakPointDate), this.props.rangeType)) {
         past.push(item);
         future.push(item);
       }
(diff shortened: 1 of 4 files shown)


src/components/TimeBasedLineChart.js
```diff
@@ -0,0 +1,753 @@
...
+
+        return this._getXScaleValue(currentDate);
+      })
+      .y0(this.state.adjustedHeight - 10)
+      .y1(this._getYScaleValue(0));
+
+    return area(data);
+  }
+
+  _getFlatLine (data) {
```
bsbeeks: Is this used anywhere?

cerinman: No, I was going to use it for animations but haven't gotten that far.  I'll remove it for now.

src/components/TimeBasedLineChart.js
```diff
@@ -0,0 +1,753 @@
...
+      }
+    });
+
+    return [past, future];
+  }
+
+  _getYAxisTicks (data, numTicks = 4) {
+    const maxValue = d3.max(this.props.data, d => {
+      const value = d.above ? d.value + d.above : d.value + 1000;
+      const multiplier = value < 0 ? -1000 : 1000;
```
bsbeeks: What is the reasoning for using `1000`?

src/components/TimeBasedLineChart.js
```diff
@@ -0,0 +1,753 @@
...
+
+    return d3.scale.linear()
+      .range([this.state.adjustedHeight, 0])
+      .domain([minValue, maxValue]);
+  }
+
+  _getYScaleValue (value) {
+    const yScale = this._getYScaleFunction();
+
+    return yScale(value) - 10;
```
bsbeeks: What is the reasoning for using `10` here?

src/components/TimeBasedLineChart.js
cerinman: Rounding to the nearest 1000s so that the ticks end up being 0, 1k, 2k, 3k, etc....

src/components/TimeBasedLineChart.js
cerinman: Accounts for an addition of 10 pixels to the bottom margin when setting up the base chart

src/components/TimeBasedLineChart.js
```diff
@@ -0,0 +1,753 @@
...
+        .attr('class', 'y-axis')
+        .attr('transform', 'translate(' + (margin.left + 30) + ',' + (margin.top - 10) + ')');
+
+      chart.append('g')
+        .attr('class', 'grid-line')
+        .attr('transform', 'translate(' + (margin.left - 10) + ',' + (margin.top - 10) + ')');
+    } else {
+      chart.append('g')
+        .append('text')
+          .attr('transform', 'translate(' + ((width / 2) - 37) + ',' + (height / 2.5) + ')')
```
bsbeeks: Looks like there are some more "magic" numbers here. What's the reasoning and can we get them dynamically some how?

cerinman: inherited numbers from your line chart in insite/target line chart.  I can do some investigating and see what we can do.

src/components/TimeBasedLineChart.js
```diff
@@ -0,0 +1,753 @@
...
+        .style('stroke-dasharray', dash)
+        .style('stroke-width', 3)
+        .style('stroke', '#30B53C')
+        .style('fill', 'none');
+    }
+
+    //Debit balance line
+    if (hoveredData.below) {
+      const below = hoveredData.value + hoveredData.below;
+      const belowValue = below < -4000 ? below + 500 : below;
```
bsbeeks: Some more "magic" numbers here

src/components/TimeBasedLineChart.js
bsbeeks: :+1:

src/components/TimeBasedLineChart.js
cerinman: Used to prevent below line from crossing the bottom of the graph into the x axis tick labels.  I can see what I can do better here as well.

src/components/TimeBasedLineChart.js
bsbeeks: Why are those 10 pixels needed?

cerinman: The line doesn't line up with the y axis grid if you don't account for it.

PR #64, opened 2015-10-22
Allow users to set a specific lower bound (positive or negative) for the RangeSelector component

bsbeeks:
This will introduce a breaking change into the RangeSelector component.

- rename `range` to `upperBound`. Functionality is the same (breaking change)
- add new `lowerBound` prop. This will determine the lowest value possible on the track. It can be positive or negative
- added some comments to help other follow the logic

@jmophoto @cerinman

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -435,7 +435,8 @@ const Demo = React.createClass({
               label: 'Silent'
             }
           ]}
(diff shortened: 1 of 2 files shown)


cerinman:
:shipit:

Issue #70, opened 2015-10-22
Review and clean up magic numbers used for margins in Time Based Line Chart


cerinman:
There are a few `magic numbers` that are being used for margins and spacing with in the Time Based Line Chart.  We need to review them and see if we can set them dynamically rather than hard coding to allow for flexibility.

PR #74, opened 2015-10-25
Fix 'month' range type issues and sort out magic numbers for margin spacings

cerinman:
Completes https://github.com/mxenabled/mx-react-components/issues/70

This PR corrects some positioning issues for the y/x axis labels, gridlines, and tool tips pertaining to the `month` range type option for the TimeBasedLineChart component.  It also resolves issue #70 where we wanted to take care of the magic numbers being used to determine the positioning for the above mentioned items.

The x axis labels and the y grid lines now use the getSliceMiddle function for their offset to make sure that they line up correctly with the data points at all times.

The y axis has an offset of 20 and this number is used to position the y axis labels right or left.  Should we want to, we can expose this number as a prop so that the user can position the y axis labels to their liking.  I've left it at 20 for now as it seems to be the sweet spot.  

I've also updated the app.js file to better allow for testing the TimeBasedLineChart with data for both the `day` and `month` range types.

@bsbeeks @jmophoto 

Before this is what the`month` range type looked like.
screen shot 2015-10-24 at 8 57 51 pm

And here it is after the changes
screen shot 2015-10-25 at 11 30 58 am

And here is proof that we have not caused issues with the `day` range type
screen shot 2015-10-25 at 11 31 19 am

### DIFF
--- a/src/components/TimeBasedLineChart.js
+++ b/src/components/TimeBasedLineChart.js
@@ -205,7 +205,6 @@ class TimeBasedLineChart extends React.Component {
 
   _renderChartBase () {
     const margin = this.props.margin;
-    const marginLeftOffSet = Math.floor(this.state.adjustedWidth / this.props.data.length);
     const width = this.props.width;
     const height = this.props.height;
     const chart = d3.select(this.state.chartEl);
@@ -441,9 +440,7 @@ class TimeBasedLineChart extends React.Component {
         })
         .attr('y', -10)
         .attr('height', this.state.adjustedHeight)
-        .attr('width', d => {
-          return this._getSliceWidth();
-        })
+        .attr('width', this._getSliceWidth())
         .style('opacity', '0')
         .style('display', d => {
           const currentDate = moment.unix(d.timeStamp);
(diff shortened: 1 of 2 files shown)


src/components/TimeBasedLineChart.js
```diff
@@ -82,16 +82,12 @@ class TimeBasedLineChart extends React.Component {
...
   }
 
-  _getSliceWidth (timeStamp) {
-    const rangeType = this.props.rangeType;
-    const startOf = timeStamp ? moment.unix(timeStamp).startOf(rangeType).unix() : moment(timeStamp).startOf(rangeType).unix();
-    const endOf = timeStamp ? moment.unix(timeStamp).endOf(rangeType).unix() : moment(timeStamp).endOf(rangeType).unix();
-
-    return Math.floor(this._getXScaleValue(endOf) - this._getXScaleValue(startOf));
+  _getSliceWidth () {
+    return Math.floor(this.state.adjustedWidth / this.props.data.length);
```
bsbeeks: nice

bsbeeks:
:shipit:

Issue #75, opened 2015-10-27
TimeBasedLineChart: Tooltips don't flip if breakpoint date is not present


cerinman:
We need to flip the tooltip labels based upon the middle date of the data set rather than the break point date to ensure that the labels always get flipped.

PR #76, opened 2015-10-27
Flip tooltip on middle date rather than break point date

cerinman:
Fixes https://github.com/mxenabled/mx-react-components/issues/75

TimeBasedLineChart component

We were flipping the tool tip direction based upon the break point date.  This was working fine but failed to take into account that the graph might not always have the break point date in the range of data points supplied.  In these cases the tool tips did not get flipped and would extend off the edge of the graph.

We no take the middle data point of the set and use it's date to determine when to flip the tooltip label.

@bsbeeks @jmophoto 

Pics of range missing break point date.  See how the tool tip flips from right to left after middle point.
screen shot 2015-10-27 at 11 50 46 am
screen shot 2015-10-27 at 11 50 54 am

### DIFF
--- a/src/components/TimeBasedLineChart.js
+++ b/src/components/TimeBasedLineChart.js
@@ -476,15 +476,16 @@ class TimeBasedLineChart extends React.Component {
   }
 
   _handleDataPointMouseOver (d) {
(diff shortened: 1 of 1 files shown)


bsbeeks:
:shipit:

PR #79, opened 2015-10-28
Remove onBlur in favor of onClick in select menu

jmophoto:
Replace onBlur with an onClick event on the scrim div for the select dropdown component.

This fixes an issue where the select click was not being triggered in IE 10/11.

@bsbeeks @cerinman


cerinman:
:shipit:

PR #90, opened 2015-10-28
[WIP] Add custom date validator

jmophoto:
I created a custom date validation method to check a prop. Rough cut so far, but it works. I am thinking this could be moved into a helper maybe? Thoughts?

@bsbeeks @cerinman


bsbeeks:
I don't think this is necessary and misses the mark a bit. We should be changing `defaultDate` to be a number, not a string. Given that it should be a number, we can just use `React.PropTypes.number`.

jmophoto:
Closing to redo with above chagnes.

Issue #89, opened 2015-10-28
DatePicker: the value for `defaultDate` is set to the start of Unix time


bsbeeks:
<img width="633" alt="screenshot 2015-10-28 15 59 54" src="https://cloud.githubusercontent.com/assets/714084/10804400/f1447568-7d8c-11e5-99a2-57516fe2eb2a.png">

It looks like this is happening between we are expecting a date string to be passed into `defaultDate` and `minimumDate`. This needs to be changed to expect a Unix timestamp.

Although, if you don't pass in a `defaultDate`, it breaks again and displays the date above.

PR #92, opened 2015-10-29
Changed props defaultDate and miniumumDate to expect a Unix timestamp

tumentumurchudur:
Fixes https://github.com/mxenabled/mx-react-components/issues/89
Changed props defaultDate and miniumumDate to expect a Unix timestamp.
Cleaned up the unix date conversion code.

### DIFF
--- a/src/components/DatePicker.js
+++ b/src/components/DatePicker.js
@@ -11,7 +11,7 @@ class DatePicker extends React.Component {
     super(props);
     this.state = {
       currentDate: null,
-      selectedDate: moment(this.props.defaultDate).unix(),
+      selectedDate: moment.unix(this.props.defaultDate),
       showCalendar: false
     };
   }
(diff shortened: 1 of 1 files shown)


src/components/DatePicker.js
```diff
@@ -21,14 +21,14 @@ class DatePicker extends React.Component {
       this._handleBlur();
     }
     this.setState({
-      selectedDate: date
+      selectedDate: moment.unix(date)
     });
 
     this.props.onDateSelect(moment.unix(date).format(this.props.format));
```
jmophoto: If we send the component a unix timestamp, we should return the date in the same format.

tumentumurchudur: yes, I agree. :+1:

tumentumurchudur:
Selected date is returned in the Unix date format.

bsbeeks:
This looks good.

PR #93, opened 2015-10-29
Refactored logic for Icon

tumentumurchudur:
Refactored the logic for caret when showing/hiding the calendar.


bsbeeks:
Looks good

Issue #91, opened 2015-10-29
TypeAhead: empty string added on enter key


tumentumurchudur:
When I type a random string without selecting an option from the dropdown and hit the enter key, the component adds an empty string to the input field.

<img width="1236" alt="screen shot 2015-10-29 at 10 11 39 am" src="https://cloud.githubusercontent.com/assets/9920303/10824303/8458bb3c-7e25-11e5-99f1-219b0e61560f.png">

PR #94, opened 2015-10-29
Empty string selection bug

tumentumurchudur:
Fixes https://github.com/mxenabled/mx-react-components/issues/91

When you hit the enter key on a random string, it won't add it to the selected items.


bsbeeks:
:shipit:

Issue #85, opened 2015-10-28
DatePicker: remove `fontSize` and use the `style` prop instead


bsbeeks:
Right now the `fontSize` prop just sets the font size on the wrapping component. It doesn't change the font size of the dates, calendar header, or other elements. Let's remove this prop and use the default style prop in it's place.

PR #95, opened 2015-10-29
Removed the prop FontSize

tumentumurchudur:
Fixes https://github.com/mxenabled/mx-react-components/issues/85

Removed the prop FontSize and used the default font from Style constants.

### DIFF
--- a/src/components/DatePicker.js
+++ b/src/components/DatePicker.js
@@ -185,7 +185,7 @@ class DatePicker extends React.Component {
     return (
       <div
         onBlur={this._handleBlur.bind(this)}
-        style={[styles.component, { fontSize: this.props.fontSize }, styles.clearFix]}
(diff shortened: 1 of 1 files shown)


src/components/DatePicker.js
```diff
@@ -185,7 +185,7 @@ class DatePicker extends React.Component {
     return (
       <div
         onBlur={this._handleBlur.bind(this)}
-        style={[styles.component, { fontSize: this.props.fontSize }, styles.clearFix]}
+        style={[styles.component, styles.clearFix]}
```
bsbeeks: We need to add `this.props.style` to this array so if a user passes in custom styles, they are applied.

tumentumurchudur:
Added this.props.style

bsbeeks:
:shipit:

PR #98, opened 2015-10-30
TimeBasedLineChart: set box-sizing: content-box

### LINKED ISSUE #82: The TimeBasedLineChart should use box-sizing: border-box
The passed in width/height of this component doesn't match the rendered width/height of the component. This is because we are adding some padding to the component. For example, if I pass in a width of `400`, then the rendered with is `420`. We should set `box-sizing: border-box`, so that the padding is taken from within the component and not added to it. For example, when I pass in `400`, the rendered width would be `400` and the content area of the component would be `380`.

bsbeeks:
After investigating a little further. It made more sense to just explicitly set `boxSizing: content-box` on the wrapping element to Fix #82.

@cerinman @jmophoto

### DIFF
--- a/src/components/TimeBasedLineChart.js
+++ b/src/components/TimeBasedLineChart.js
@@ -646,7 +646,8 @@ const styles = {
   component: {
     fontFamily: StyleConstants.FontFamily,
     padding: '10px',
-    position: 'relative'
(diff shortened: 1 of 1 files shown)


cerinman:
:shipit:

PR #101, opened 2015-11-04
TwoLevelSelector component

tumentumurchudur:
TwoLevelSelector component: 
- Removed MX specific functionality

  
  Summary of props
- data (object) - accepts an object in the following format:

  {
   'option1': ['sub11', 'sub12'],
   'option2': ['sub21', 'sub22', 'sub23'],
   'option3': ['sub31'],
   'option4': ['sub41', 'sub42', 'sub43', 'sub44', 'sub45', 'sub46', 'sub47', 'sub48'],
   'option5': ['sub51', 'sub52', 'sub53', 'sub54', 'sub55']
  }
- onChange(func) - callback function that passes the selected text
- title (string) - text for the input field before an option is selected. It is 'Choose an item' by default.
- selectedItem (string) - is the default selected text  

screen shot 2015-11-04 at 10 36 54 am

### DIFF
--- a/src/components/TwoLevelSelector.js
+++ b/src/components/TwoLevelSelector.js
@@ -39,23 +39,26 @@ const TwoLevelSelector = React.createClass({
   },
 
   _getParentItems () {
-    let parentItems = [];
+    const parentItems = [];
+    const items = this.props.data;
 
-    for (let key in this.props.data) {
-      parentItems.push(key);
+    for (const key in items) {
+      if (items.hasOwnProperty(key)) {
+        parentItems.push(key);
+      }
     }
     return parentItems;
   },
 
   _flattenArray (obj) {
-    let tempArray = [];
+    const tempArray = [];
 
-    for (let key in obj) {
-     tempArray.push(key);
+    for (const key in obj) {
+      tempArray.push(key);
 
-     obj[key].map(function(val) {
-      tempArray.push(val);
-     })
+      obj[key].map(function(val) {
+        tempArray.push(val);
+      });
     }
     return tempArray;
   },
--- a/src/components/TwoLevelSelector.js
+++ b/src/components/TwoLevelSelector.js
@@ -58,9 +58,9 @@ const TwoLevelSelector = React.createClass({
         tempArray.push(key);
       }
 
-      obj[key].map(val => {
-        tempArray.push(val);
-      });
+      for (let i = 0; i < obj[key].length; i ++) {
+        tempArray.push(obj[key][i]);
+      }
     }
     return tempArray;
   },
--- a/src/components/TwoLevelSelector.js
+++ b/src/components/TwoLevelSelector.js
@@ -50,13 +50,15 @@ const TwoLevelSelector = React.createClass({
     return parentItems;
   },
 
-  _flattenArray (obj) {
+  _flattenObjectToArray (obj) {
     const tempArray = [];
 
     for (const key in obj) {
-      tempArray.push(key);
+      if (obj.hasOwnProperty(key)) {
+        tempArray.push(key);
+      }
 
-      obj[key].map(function(val) {
+      obj[key].map(val => {
         tempArray.push(val);
       });
     }
@@ -64,7 +66,7 @@ const TwoLevelSelector = React.createClass({
   },
 
   _getCategoriesByKeyword (searchKeyword) {
-    const flatList = this._flattenArray(this.props.data);
+    const flatList = this._flattenObjectToArray(this.props.data);
     const searchList = flatList.filter(option => {
       return option.toLowerCase().match(searchKeyword.toLowerCase());
     });
--- a/src/components/TwoLevelSelector.js
+++ b/src/components/TwoLevelSelector.js
@@ -58,7 +58,7 @@ const TwoLevelSelector = React.createClass({
         tempArray.push(key);
       }
 
-      for (let i = 0; i < obj[key].length; i ++) {
+      for (let i = 0; i < obj[key].length; i++) {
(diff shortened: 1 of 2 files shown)


src/components/TwoLevelSelector.js
```diff
@@ -0,0 +1,377 @@
+const React = require('react');
+const ReactDOM = require('react-dom');
+const Radium = require('radium');
+
+const Icon = require('./Icon');
+const StyleConstants = require('../constants/Style');
+
+const TwoLevelSelector = React.createClass({
```
bsbeeks: I'm not a huge fan of this name. Any other suggestions? Here are some options: `FlyoutMenu`, `TieredMenu`

cerinman: I like FlyoutMenu

src/components/TwoLevelSelector.js
```diff
@@ -0,0 +1,377 @@
...
+
+  propTypes: {
+    data: React.PropTypes.object,
```
bsbeeks: I think the expected data structure can be solidified a bit more. Here is what I propose:
```javascript
[
  {
    label: 'Has A Second Level Menu', //optional: if not set, just display the value
    value: 'menu-1',
    items: [
      {
        label: 'Second Level Menu Item', //optional: if not set, just display the value
        value: 'second-level-1',
      }
    ],
  },
  {
    label: 'No Second Level Menu', //optional: if not set, just display the value
    value: 'menu-2'
  }
]
```

src/components/TwoLevelSelector.js
```diff
@@ -0,0 +1,377 @@
...
+    onChange: React.PropTypes.func,
+    selectedItem: React.PropTypes.string,
+    title: React.PropTypes.string
+  },
+
+  getDefaultProps () {
+    return {
+      data: null,
```
bsbeeks: It seems like the data prop should be required for this component to work. If that's the case, then we shouldn't set a default prop.

src/components/TwoLevelSelector.js
```diff
@@ -0,0 +1,377 @@
...
+
+const Icon = require('./Icon');
+const StyleConstants = require('../constants/Style');
+
+const TwoLevelSelector = React.createClass({
+
+  propTypes: {
+    data: React.PropTypes.object,
+    onChange: React.PropTypes.func,
+    selectedItem: React.PropTypes.string,
```
cerinman: So, as  a user of this component, I might want more than just a string back.  Sometimes I care more about the associated id or guid of a selected item rather than it's display name.  Can we change this and the code around it to pass back more than a string.  This might require changing how the data prop is formatted but that might make this component a little more usable. 

I imagine the data prop looking like this

```javascript
[
  {
    name: 'groceries'
    id: 1232341234
    subs: [
      {
        //same layout as parent
      }
    ]
  }
]
```

src/components/TwoLevelSelector.js
tumentumurchudur: Sounds good. I will change the name to FlyoutMenu

src/components/TwoLevelSelector.js
tumentumurchudur: @cerinman @bsbeeks 
Yes, agreed. How about we change the data structure as follows:
```javascript
[
  {
    id: 1,
    value: 'groceries',
    items: [
      {
        id: 11,
        value: 'item1'
      },
      {
        id: 12,
        value: 'item2'
      }
    ]
  },
  {
    id: 2,
    value: 'books',
    items: [
      {
        id: 21,
        value: 'book1'
      }
    ]
  }
]
```

And then return an object onSelect event

src/components/TwoLevelSelector.js
tumentumurchudur: Yes, that is a good idea to require this prop.

src/components/TwoLevelSelector.js
cerinman: I still think @bsbeeks format is better.  I like the idea of the label key containing the displayed text and the value key containing an id or guid.

tumentumurchudur: Okay, I will change it. The structure is the same but the naming of properties is little different.
```javascript
[
  {
    label: 'groceries',
    value: '1xxxx',
    items: [
      {
        label: 'milk',
        value: '12xxx'
      },
      {
        label: 'eggs',
        value: '23xxx'
      }
    ]
  },
  {
    label: 'books',
    value: '31xxx',
    items: [
      {
        label: 'bookXY',
        value: '31xxx'
      }
    ]
  }
]
```

src/components/TwoLevelSelector.js
```diff
@@ -0,0 +1,377 @@
...
+  },
+
+  _handleBlur () {
+    this.setState({
+      showingOptions: false
+    });
+  },
+
+  _handleClick (selectedItem) {
+    this.props.onChange(selectedItem);
```
cerinman: And make sure the selectedItem you are returning is the entire object and not just the label as a string.

```javascript
{ label: 'groceries', value: 1 }
```

tumentumurchudur: @cerinman 
Yup, I will make sure an object is returned from the array.

tumentumurchudur:
- Changed the data structure for the data prop
- Cleaned up the styles
- Changed the onSelect callback to return an object from the passed data array

screen shot 2015-11-06 at 2 54 07 pm
screen shot 2015-11-06 at 2 55 11 pm

bsbeeks:
It doesn't look like we'll be using this anytime soon. We can reopen if needed.

PR #106, opened 2015-11-09
Add styling props to DatePicker for more flexibility

cerinman:
While using the DatePicker in our MX app, I ran into a few issues trying to style the component to match designs.  This change adds a style prop for the selectedDate and calendar wrapper divs so that we can over write the base styling when needed.

I also renamed the style object names for these two divs for clarity as well as removed a prop that is no longer supported from the demo app for the DatePicker.

@bsbeeks @jmophoto

proof that the changes did not cause any harm to standard use of the component in the demo app.
screen shot 2015-11-09 at 2 42 26 pm

### DIFF
--- a/src/components/DatePicker.js
+++ b/src/components/DatePicker.js
@@ -175,16 +175,16 @@ class DatePicker extends React.Component {
         tabIndex='0'
       >
         <div key='selectedDateWrapper' style={[
-            (this.props.selectedDateWrapperStyle || styles.selectedDateWrapper),
-            this.state.showCalendar && styles.selectedDateWrapperOpen
-          ]}>
+          (this.props.selectedDateWrapperStyle || styles.selectedDateWrapper),
+          this.state.showCalendar && styles.selectedDateWrapperOpen
+        ]}>
           {this._renderSelectedDate()}
           {this._renderCaret()}
         </div>
         <div key='calendarWrapper' style={[
-            (this.props.calendarWrapperStyle || styles.calendarWrapper),
-            this.state.showCalendar && styles.calendarShow
-          ]}>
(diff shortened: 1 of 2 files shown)


src/components/DatePicker.js
```diff
@@ -174,11 +174,17 @@ class DatePicker extends React.Component {
         style={[styles.component, styles.clearFix, this.props.style]}
         tabIndex='0'
       >
-        <div key='componentTop' style={[styles.componentTop, this.state.showCalendar && styles.componentTopOpen]}>
+        <div key='selectedDateWrapper' style={[
+          (this.props.selectedDateWrapperStyle || styles.selectedDateWrapper),
+          this.state.showCalendar && styles.selectedDateWrapperOpen
```
bsbeeks: Why don't we just add this into the array instead of using an "or" statement?
```
[
this.props.selectedDateWrapperStyle,
styles.selectedDateWrapper,
this.state.showCalendar && styles.selectedDateWrapperOpen
]
```

cerinman: I did try that but I wasn't having any luck with the prop style overwriting the component style when I did that.  I would prefer the way you have outlined so perhaps I was just doing something wrong.  I'll play with the order of styles in the array to see if I can get better results.

cerinman:
@bsbeeks The correction has made.  For future reference if you didn't already know, the order of the styles in the array matter.  I had the props style first so the component style always over wrote any key/value in both with it's value.

bsbeeks:
@cerinman you're correct. With Radium styles, the latter styles will override previous ones. Similar to common object merge methods.

bsbeeks:
:shipit:

PR #107, opened 2015-11-11
Refactor DatePicker component

cerinman:
This PR refactors the DatePicker component for the following reasons.

- Allow and handle nullifying or emptying the selectedDate input so that the component can used for a `not required` fields.
- Adds more props for styling(See new list of props below)
- Handle placeholder text for empty or nullified selectedDate input
- Fix onBlur weirdness by moving to an onClick on scrim
- Fix double caret's being drawn when using input for selectedDate
- Reorder a few functions that were not ordered alphabetically

New Props added

- inputStyle - object - used to custom style the selected date input
- placeholderText - string - used as placeholder for empty selected date input
- placeholderTextStyle - object - used to style the placeholder text

Props removed

- useScrim - boolean - was used to not render the scrim.  

Now that we rely on the scrim for closing the calendar when a user blurs off the component by clicking on the scrim we must always render it.  The user can use the provided scrimStyle prop to style the scrim but it is by default not visible.

Props with modified default values

- defaultDate - unix time stamp - number - The default value was today's date but now that we want to be able to allow null or empty values their is no longer a default value for this prop.

Screen shots:
screen shot 2015-11-10 at 8 24 34 pm
screen shot 2015-11-10 at 8 24 44 pm
screen shot 2015-11-10 at 8 25 50 pm

### DIFF
--- a/src/components/DatePicker.js
+++ b/src/components/DatePicker.js
@@ -27,7 +27,7 @@ class DatePicker extends React.Component {
       if (newDate.isValid()) {
         inputValue = newDate.format(this.props.format);
       } else {
-        inputValue = 'Invalid Date';
+        inputValue = date;
       }
     }
 
@@ -47,6 +47,7 @@ class DatePicker extends React.Component {
 
     this.setState({
       inputValue: moment.unix(date).format(this.props.format),
+      isValid: true,
       selectedDate: date
     });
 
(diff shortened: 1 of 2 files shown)


src/components/DatePicker.js
```diff
@@ -11,24 +11,81 @@ class DatePicker extends React.Component {
...
+
+  _handleInputChange (evt) {
+    this.setState({
+      inputValue: evt.target.value
+    });
+  }
+
   _handlePreviousClick () {
-    const selectedDate = moment(this.state.selectedDate).locale(this.props.locale);
+    const selectedDate = moment.unix(this._getSelectedDate()).locale(this.props.locale);
```
bsbeeks: Isn't `getSelectedDate` already converting it to a unix timestamp?

bsbeeks:
Just the one question and then this should be good to go. Be sure to update the docs branch once this gets merged.

src/components/DatePicker.js
cerinman: No, `_getSelectedDate` is returning a unix time stamp.  See line 40.  It either returns the existing unix time stamp contained in this.state.selectedDate or a new moment for today as a unix time stamp.

bsbeeks: Gotcha

bsbeeks:
:shipit:

cerinman:
@bsbeeks  Made fix to caret/text spacing here as well.

screen shot 2015-11-11 at 9 33 16 am

bsbeeks:
:shipit:

PR #108, opened 2015-11-11
Style caret icon on select component

cerinman:
The caret up/down icon on the select component used to show the user if the component is open or closed did not have a style that matched the same icons in other components.  This PR changes the styles to bring it inline for consistency.  

Select Component after change
screen shot 2015-11-10 at 9 12 28 pm

DatePicker Component for reference to what we needed to match
screen shot 2015-11-10 at 9 12 38 pm

### DIFF
--- a/src/components/Select.js
+++ b/src/components/Select.js
@@ -101,7 +101,7 @@ class Select extends React.Component {
             {selected.displayValue}
             <Icon
               size='20'
-              style={[styles.downArrow, this.props.selectedStyle && { color: this.props.selectedStyle.color }]}
+              style={styles.caret}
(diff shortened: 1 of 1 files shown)


bsbeeks:
The space to the right of the caret is smaller than the space to the left of the text. Can we make these the same? Preferably, increase the space to the right of the caret to match. If needed, let's make the adjustment to the Input component to match.

cerinman:
@bsbeeks The spacing has been fixed here.  I'll update my DatePicker PR with the same change for it's caret as well.

screen shot 2015-11-11 at 9 26 49 am

bsbeeks:
:shipit:

PR #109, opened 2015-11-13
A few more bug fixes for the DatePicker component

cerinman:
This PR does the following.

- When using an input for the selected date, we no longer remove the bottom border for the input wrapper when the calendar is showing.  This feature was causing weird behavior when a user would style the input wrapper to not have any border.  This also now has the added benefit of mimicking the other components that have a similar open closed state.

- Refactor the `_handleInputBlur` function.  Before we where trying to validate the date via the string value from the input and updating state and calling the `onDateSelect` callback.  This was causing several call backs to happen and also would allow the user to cause unexpected behavior by doing something like typing `Nov` rather than a complete date.  We now check the value of the input to see if it has length.  If it doesn't then we just set things to null and move on.  If it does have length then we restore the current selected date that is in state.  This allows the user to null out the input but forces them to use the calendar to select a date.

- Wraps the caret icon in a div and then allows a prop to style it.  Positioning the caret to match up with other inputs with similar carets was becoming difficult and their was no good way of just effecting the caret icon.  

- Change moment to use utc for dates so that they get formatted using utc rather than with the user time zone offset.

Props Added

- caretWrapperStyle: object - allows for styling the div wrapper around the caret icon

### DIFF
--- a/src/components/DatePicker.js
+++ b/src/components/DatePicker.js
@@ -55,26 +55,17 @@ class DatePicker extends React.Component {
   }
 
   _handleInputBlur (evt) {
-    let inputValue = null;
-    let isValid = true;
-    let selectedDate = null;
+    if (evt.target.value.length === 0) {
+      this.props.onDateSelect(null);
 
-    if (evt.target.value && evt.target.value.length) {
-      const newDate = moment(new Date(evt.target.value));
-
-      inputValue = this._getInputValueByDate(newDate.isValid() ? newDate.unix() : evt.target.value);
-      isValid = newDate.isValid();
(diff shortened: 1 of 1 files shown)


bsbeeks:
Why are we NOT wanting to use the user's local time zone offset?

cerinman:
@bsbeeks As it pertains to the date picker, it is more important that the date is right than the time of day.  The component takes in and returns a unix time stamp but moment will try to offset the time stamp when we format it to be displayed to the user.  By switching to UTC we ensure that a date that is set to the beginning of the day doesn't get offset to the previous day by mistake.  This is the situation that I'm currently dealing with internally when using this component.

bsbeeks:
makes sense. :shipit:

PR #110, opened 2015-11-16
Revert utc changes and fix input placeholder for DatePicker component

cerinman:
After an off line discussion, it was decided to revert the UTC changes made to the DatePicker component.  This PR does just that.  It also fixes the placeholder text for the input.


bsbeeks:
:shipit:

PR #111, opened 2015-11-18
Add new FormField component

cerinman:
This PR adds a new component FormField.  This component is a wrapper for form fields that provides a nice clean layout of the forms label, is required error message, and placeholder text.

screen shot 2015-11-18 at 11 03 54 am
screen shot 2015-11-18 at 11 04 13 am

It can wrap just about anything that requires a label and some basic validation.

Props

- fieldWrapperStyle: (array | object) - Style object or Radium array of style objects used to style the field wrapper.
- height: number(default 40) - Height of component.
- isRequired: boolean(default false) - Used to show or hide the `* Required` message in top right corner if field.
- label: string - The label to rendered next to the field.
- labelStyle: (array | object) - Style object or Radium array of style objects used to style the label.
- placeholder: string - The placeholder text to be drawn below the field.
- placeholderStyle: (array | object) - Style object or Radium array of style objects used to style the placeholder.
- requiredStyle: (array | object) - Style object or Radium array of style objects used to style the required error message.
- width: number(default 350) - The width of the component

Classes available for styling the component

- mx-form-field-component : used to style component as a whole.
- mx-form-field-label: used to style label.
- mx-form-field-wrapper: used to style the field wrapper.
- mx-form-field-required: used to style the required error message.
- mx-form-field-placeholder: used to style the placeholder.

### DIFF
--- a/src/components/FormField.js
+++ b/src/components/FormField.js
@@ -1,23 +1,9 @@
 const React = require('react');
 const Radium = require('radium');
-const moment = require('moment');
 
 const StyleConstants = require('../constants/Style');
 
 class FormField extends React.Component {
-  constructor (props) {
-    super(props);
-    this.state = {
-      focused: false
-    };
-  }
-
-  _toggleFieldFocus () {
-    this.setState({
-      focused: !this.state.focused
-    });
-  }
-
   render () {
     return (
       <div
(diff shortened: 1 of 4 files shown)


src/components/FormField.js
```diff
@@ -0,0 +1,121 @@
...
+const StyleConstants = require('../constants/Style');
+
+class FormField extends React.Component {
+  render () {
+    return (
+      <div
+        className='mx-form-field-component'
+        style={[
+          styles.component, this.props.style,
+          { height: this.props.height + 'px', width: this.props.width + 'px' }
```
tumentumurchudur: Not sure if we always have to append 'px' but it defaults to pixels.
`{ height: this.props.height, width: this.props.width }`

src/components/FormField.js
```diff
@@ -0,0 +1,121 @@
...
+        ]}
+      >
+        <div
+          className='mx-form-field-label'
+          style={[
+            styles.label, this.props.labelStyle,
+            { marginTop: this.props.height - 16 }
```
tumentumurchudur: Trying to understand this line here, so I set the height of the component to 300px and then the label would have a top margin of 284px?

cerinman:
After an offline discussion, we have decided to drop this component from the open source for now.

PR #118, opened 2015-11-30
TimeBasedLineChart: Dont render zero axis if to close to another tick.

cerinman:
This prevents the addition of the zero y axis tick if zero is to close to an existing tick.  This stops the behavior where the zero y axis would some times overlap another y axis tick.  The shading of the area below zero will still happen.

@bsbeeks @jmophoto @malcolmwebdev @tumentumurchudur 

Before:
screen shot 2015-11-30 at 1 37 53 pm
 
After:
screen shot 2015-11-30 at 1 49 42 pm


bsbeeks:
:shipit:

PR #119, opened 2015-11-30
Update the way we center Modal content, add an isRelative prop

bsbeeks:
This also sets a max height on the modal content to avoid the modal displaying off the screen.

@mxenabled/frontend-engineers

### DIFF
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -88,11 +88,11 @@ class Modal extends React.Component {
   render () {
     if (this.props.isOpen)
       return (
-        <div className='mx-modal' style={styles.scrim}>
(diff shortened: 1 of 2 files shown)


src/components/Modal.js
```diff
@@ -166,7 +161,9 @@ const styles = {
     textTransform: 'uppercase'
   },
   content: {
-    position: 'relative'
+    position: 'relative',
+    maxHeight: 'calc(100% - 140px)',
```
cerinman: I didn't know about calc.  Interesting.

cerinman:
:shipit:

src/components/Modal.js
bsbeeks: yep, it's now supported by all modern browsers without any prefixes.

PR #120, opened 2015-11-30
Fix positioning and text alignment on Modal content

bsbeeks:
Also, with the new changes to how we position the modal content, we no longer need the `isSmall` prop, so that is now deprecated.

@cerinman


cerinman:
nice :+1:

PR #125, opened 2015-12-08
Fix modal content height issue.

cerinman:
This fixes an issue with the max height on the content div of the modal component.  We are using the the css utility `calc` for the default value but it is causing issues in some edge case where the content of the modal would force the bottom of the modal out of it's parent.

To fix this, we now allow a contentStyle prop where users can over ride the default and set their own height/max-height for the content div.

New Prop:

contentStyle (Object) : Style object used to style the content div of the modal 

@mxenabled/frontend-engineers 

Screen shots of the demo app using the 100% max-height default for default and small.
screen shot 2015-12-07 at 5 17 40 pm
screen shot 2015-12-07 at 5 17 54 pm

### DIFF
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -163,7 +163,7 @@ const styles = {
   },
   content: {
     position: 'relative',
-    maxHeight: '100%',
+    maxHeight: '100% - 140px',
     overflow: 'auto'
   },
   footer: {
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -102,7 +102,7 @@ class Modal extends React.Component {
               type='close-solid'
             />
             {this._renderTitleBar()}
-            <div className='mx-modal-content' style={styles.content}>
+            <div className='mx-modal-content' style={[styles.content, this.props.contentStyle]}>
               {this.props.children}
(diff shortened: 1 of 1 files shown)


cerinman:
eslint is passing locally.  Not sure why it's failing in github.

src/components/Modal.js
```diff
@@ -163,7 +163,7 @@ const styles = {
   },
   content: {
     position: 'relative',
-    maxHeight: 'calc(100% - 140px)',
+    maxHeight: '100%',
```
bsbeeks: For the default I think we should leave the `-140px`, it accounts for the header and footer elements.

src/components/Modal.js
```diff
@@ -163,7 +163,7 @@ const styles = {
   },
   content: {
     position: 'relative',
-    maxHeight: 'calc(100% - 140px)',
+    maxHeight: '100% - 140px',
```
cerinman: @bsbeeks is this what you meant?  Or did you mean with the calc call as well?

cerinman:
@bsbeeks ready for another review

bsbeeks:
:shipit: We'll need to look into the eslint issue

PR #126, opened 2015-12-09
Add selectedDateColor prop to DatePicker component

cerinman:
We have a need to be able to set the color used for the selected date and hover state of days on the calendar.  This PR adds that ability via a new prop.  

Also update the DatePicker component to use the Style constants file for colors/font sizes and stripped the px unit where possible.

New Prop

selectedDateColor Type(string) - Default('#359BCF') - Used to change the color of the selected date and hover state of days on the calendar.

@mxenabled/frontend-engineers 

screen shot 2015-12-09 at 4 20 11 pm

### DIFF
--- a/src/components/DatePicker.js
+++ b/src/components/DatePicker.js
@@ -134,8 +134,8 @@ class DatePicker extends React.Component {
             key={startDate.format('DDDD')}
             style={[
               styles.calendarDayContent,
-              noSelectDay ? styles.calendarDayDisabled : isCurrentDay && { backgroundColor: this.props.selectedDateColor, color: '#FFFFFF' },
-              { ':hover': { backgroundColor: this.props.selectedDateColor, color: '#FFFFFF' } }
+              noSelectDay ? styles.calendarDayDisabled : isCurrentDay && { backgroundColor: this.props.selectedDateColor, color: StyleConstants.Colors.INVERSE_PRIMARY },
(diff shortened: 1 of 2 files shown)


cerinman:
Again, eslint is passing locally but failing here.

bsbeeks:
you can take out the datepicker fullscreen component. I'm in the middle of a complete rewrite of it.

src/components/DatePicker.js
```diff
@@ -132,7 +132,11 @@ class DatePicker extends React.Component {
         >
           <div
             key={startDate.format('DDDD')}
-            style={[styles.calendarDayContent, noSelectDay ? styles.calendarDayDisabled : isCurrentDay && styles.currentDay]}
+            style={[
+              styles.calendarDayContent,
+              noSelectDay ? styles.calendarDayDisabled : isCurrentDay && { backgroundColor: this.props.selectedDateColor, color: StyleConstants.Colors.INVERSE_PRIMARY },
+              { ':hover': { backgroundColor: this.props.selectedDateColor, color: StyleConstants.Colors.INVERSE_PRIMARY, cursor: 'pointer' } }
```
bsbeeks: could we define this in a variable above so we can format it with line breaks?

cerinman:
@bsbeeks ready for another review

bsbeeks:
:shipit:

PR #127, opened 2015-12-10
Update eslint modules, lock down eslint version used in travis, remove rule for multiple react components in same file

bsbeeks:
The eslint rule that checks to make sure only one React component is defined per file, doesn't like it when jsx is used outside of the ReactComponent class. We are using jsx when defining one of the default props in SelectFullScreen. Removing the rule for now since it's one we rarely (if ever) break. 

Also updated the eslint modules to correspond with our other internal projects at MX.

@mxenabled/frontend-engineers


cerinman:
:shipit:

Issue #112, opened 2015-11-18
Select: Cannot scroll using the up/down keys


tumentumurchudur:
When I click the select component and try to scroll up and down using the up/down arrows but it does not work. 

<img width="1232" alt="screen shot 2015-11-18 at 4 11 28 pm" src="https://cloud.githubusercontent.com/assets/9920303/11257499/176e933a-8e0f-11e5-89bc-21c1deeae80e.png">

PR #128, opened 2015-12-15
Select: Up/down keys allow scrolling the list

tumentumurchudur:
Fixes https://github.com/mxenabled/mx-react-components/issues/112

- Added the ability to scroll using the up/down keys
- Select an item by hitting the Enter key


src/components/Select.js
```diff
@@ -156,6 +217,11 @@ const styles = {
   selected: {
     position: 'relative'
   },
+  activeItem: {
+    backgroundColor: StyleConstants.Colors.PRIMARY,
+    color: StyleConstants.Colors.INVERSE_PRIMARY,
```
bsbeeks: I'm updating the colors in another PR, to avoid conflict we'll want to change this to `WHITE` instead of `INVERSE_PRIMARY`.

src/components/Select.js
```diff
@@ -156,6 +217,11 @@ const styles = {
...
+    opacity: 1
```
bsbeeks: Is this needed? It doesn't look like we're setting opacity anywhere else.

bsbeeks:
Just a couple minor changes, then this is good to go. Looks good!

src/components/Select.js
tumentumurchudur: I will remove it.

tumentumurchudur:
@bsbeeks Ready for another review.
- Renamed `INVERSE_PRIMARY` to `WHITE`
- Removed `opacity: 1`

bsbeeks:
:shipit:

bsbeeks:
I'll release this with my changes.

PR #131, opened 2015-12-21
Fix nuances with TimeBasedLineChart spacing

bsbeeks:
- Allowing a user to set the styles, specifically `padding` and `box-sizing` was breaking the chart width and spacing. This removes the ability to set the styles on the wrapping `div`. Users should still be able to accomplish the same results by wrapping this component with a div that hold their custom styles.
- Now that the spacing is cleaned up, the best default right margin is `0`.
- This also moves the zeroState out of D3 and exposes a `zeroState` prop that allows users to set the chart zero state using anything that is renderable (string, number, jsx, etc).

@mxenabled/frontend-engineers

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -579,18 +579,18 @@ const Demo = React.createClass({
         />
 
         <br/><br/>
-        <TimeBasedLineChart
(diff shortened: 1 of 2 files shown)


cerinman:
:shipit:

PR #132, opened 2015-12-22
Fix TimeBasedLineChart when receiving new data props

bsbeeks:
My last changes worked if the data is passed into the initial props. If data was passed into the prop anytime after initial mount, then the chart wouldn't display. This was due to the fact that we were not initially rendering out the svg element that D3 uses. So, after the initial load, the reference to the svg didn't exist.

This also does the following:
- Updates the demo to mimic the behavior of the data coming in after initial mount
- Moves the mouseLeave event handler onto the svg element
- Updates object-assign module (was getting some dependency conflicts, all npm commands still work)

@mxenabled/frontend-engineers

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -296,188 +296,132 @@ const icons = [
   }
 ];
 
-const lineChartData = {
-  day: [
-    {
-      timeStamp: moment().subtract(15, 'days').startOf('day').unix(),
-      value: 800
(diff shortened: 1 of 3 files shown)


cerinman:
Nice!  :shipit:

jmophoto:
:shipit:

PR #133, opened 2016-01-01
Luke/range selector click functionality

lukeschunk:
# Summary of Features Added
* Click above the upper selector on/near the track and the upper toggle will move
* Click below the lower selector on/near the track and the lower toggle will move
* Click between the two toggles, and the closest one will move

### Added a ` <div className='track-holder'> ` Element
* This element was added because the `range` element already was handling mouse up events, and the track is too thin to reliably handle clicks
* The element now contains the track, and extends several pixels beyond the track so the user clicking doesn't have to be so exact

#####  `_handleClickStartOnTrack()` function 
* Function determines which toggle should move by setting `this.state.dragging` example:
``` 	
//click point is less than lower selector
	if (newPixels < this.state.lowerPixels) {
	  this.setState({
		dragging: 'Lower'
	  });
	}
```

#####  `_handleClickOnTrack() ` function
* Moves the proper toggle

### Misc:
* Padding on `range` div was adjusted to compensate for the new `track-holder` containing element that I added
* I really enjoyed working on this - let me know if you see any edits/issues
* I'm not sure why the formatting looks weird in the changes shown on github - I matched the spacing and indents and everything in Atom and it looked fine there...

### DIFF
--- a/src/components/RangeSelector.js
+++ b/src/components/RangeSelector.js
@@ -20,7 +20,8 @@ class RangeSelector extends React.Component {
       selectedLabel: this._getSelectedLabel(lowerValue, upperValue),
       showPresets: !!this.props.presets.length && !lowerValue && !upperValue,
       upperPixels: 1,
-      upperValue
+      upperValue,
+      trackClicked: false
     };
   }
 
@@ -92,7 +93,11 @@ class RangeSelector extends React.Component {
     });
   }
 
-  _handleClickStartOnTrack (e) {
+  _handleTrackMouseDown (e) {
+    this.setState({
+      trackClicked: true
+    });
+    
     const clientX = e.touches ? e.touches[0].clientX : e.clientX;
 
     const newPixels = clientX - ReactDOM.findDOMNode(this.refs.rangeSelector).getBoundingClientRect().left;
@@ -125,35 +130,8 @@ class RangeSelector extends React.Component {
       });
     }
   }
-
-  _handleClickOnTrack (e) {
-    if (this.state.dragging) {
-      const clientX = e.touches ? e.touches[0].clientX : e.clientX;
-      const pixelInterval = this.props.interval * this.state.width / this.state.range;
-      const newState = {
-        selectedLabel: null
-      };
-
-      let newPixels = clientX - ReactDOM.findDOMNode(this.refs.rangeSelector).getBoundingClientRect().left;
-
-      //make sure we snap to our interval
-      newPixels = Math.round(newPixels / pixelInterval) * pixelInterval;
-
-      //convert our pixels to a 0-based scale
-      const newPosition = (newPixels * this.state.range / this.state.width) + this.props.lowerBound;
-
-      //covert our 0-based value to actual value
-      const newValue = Math.round(newPosition / this.props.interval) * this.props.interval;
-
-      newState[this.state.dragging.toLowerCase() + 'Pixels'] = newPixels;
-      newState[this.state.dragging.toLowerCase() + 'Value'] = newValue;
-
-      this.setState(newState);
-
-      e.preventDefault();
-    }
-  }
-
+  
+  //this method now handles both the dragging of the toggle, and moving it when track is clicked
   _handleDragging (e) {
     if (this.state.dragging) {
       const clientX = e.touches ? e.touches[0].clientX : e.clientX;
@@ -198,7 +176,11 @@ class RangeSelector extends React.Component {
     }
   }
 
-  _handleDragEnd () {
+  _handleDragEnd (e) {
+    if (this.state.trackClicked) {
+      this._handleDragging(e);
+    }
+    
     if (this.state.dragging) {
       this.props['on' + this.state.dragging + 'DragStop'](this.state[this.state.dragging.toLowerCase() + 'Value']);
     }
@@ -358,8 +340,7 @@ class RangeSelector extends React.Component {
           {this.props.presets.length ? <div className='mx-rangeselector-toggle' onClick={this._handleToggleViews.bind(this)} style={styles.showPresets}>Groups</div> : null}
 	<div
           className='mx-rangeselector-track-holder'
-          onMouseDown={this._handleClickStartOnTrack.bind(this)}
-          onMouseUp={this._handleClickOnTrack.bind(this)}
+          onMouseDown={this._handleTrackMouseDown.bind(this)}
           style={styles.trackHolder}
 	>
           <div className='mx-rangeselector-track' style={styles.track}></div>
--- a/src/components/RangeSelector.js
+++ b/src/components/RangeSelector.js
@@ -93,66 +93,66 @@ class RangeSelector extends React.Component {
   }
 
   _handleClickStartOnTrack (e) {
-	const clientX = e.touches ? e.touches[0].clientX : e.clientX;
+    const clientX = e.touches ? e.touches[0].clientX : e.clientX;
 
-	let newPixels = clientX - ReactDOM.findDOMNode(this.refs.rangeSelector).getBoundingClientRect().left;
+    let newPixels = clientX - ReactDOM.findDOMNode(this.refs.rangeSelector).getBoundingClientRect().left;
 
-	//click point is less than lower selector
-	if (newPixels < this.state.lowerPixels) {
-	  this.setState({
-		dragging: 'Lower'
-	  });
-	}
+    //click point is less than lower selector
+    if (newPixels < this.state.lowerPixels) {
+      this.setState({
+        dragging: 'Lower'
+      });
+    }
 
-	//click point is higher than upper selector
-	if (newPixels > this.state.upperPixels) {
-	  this.setState({
-		dragging: 'Upper'
-	  });
-	}
+    //click point is higher than upper selector
+    if (newPixels > this.state.upperPixels) {
+      this.setState({
+	dragging: 'Upper'
+      });
+    }
 
-	//click point is higher than lower selector && also lower than midway point between selectors
-	if (newPixels > this.state.lowerPixels && newPixels < (this.state.lowerPixels + (this.state.upperPixels - this.state.lowerPixels) / 2)) {
-	  this.setState({
-		dragging: 'Lower'
-	  });
-	}
+    //click point is higher than lower selector && also lower than midway point between selectors
+    if (newPixels > this.state.lowerPixels && newPixels < (this.state.lowerPixels + (this.state.upperPixels - this.state.lowerPixels) / 2)) {
+      this.setState({
+	dragging: 'Lower'
+      });
+    }
 
-	//click point is lower than upper selector && also greater than midway point between selectors
-	if (newPixels < this.state.upperPixels && newPixels > (this.state.upperPixels - (this.state.upperPixels-this.state.lowerPixels) / 2)) {
-	  this.setState({
-	    dragging: 'Upper'
-	  });
-	}
+    //click point is lower than upper selector && also greater than midway point between selectors
+    if (newPixels < this.state.upperPixels && newPixels > (this.state.upperPixels - (this.state.upperPixels-this.state.lowerPixels) / 2)) {
+      this.setState({
+        dragging: 'Upper'
+      });
+    }
   }
 
   _handleClickOnTrack (e) {
     if (this.state.dragging) {
+ 
+      const clientX = e.touches ? e.touches[0].clientX : e.clientX;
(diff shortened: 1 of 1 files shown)


cerinman:
Hi @lukeschunk.  Thanks for the contribution.  Could I get you to take some screen shots and post so that we can see the selector after your changes?  I'm reviewing your code now but screen shots of UI changes are always helpful in these situations!  Thanks again.

src/components/RangeSelector.js
```diff
@@ -92,6 +92,68 @@ class RangeSelector extends React.Component {
     });
   }
 
+  _handleClickStartOnTrack (e) {
+    const clientX = e.touches ? e.touches[0].clientX : e.clientX;
```
cerinman: Is the touches array supported in all browsers?  Maybe link to the documentation if you would.

cerinman:
So my main question is about the touches array.  I'm not sure if it's intended use matches the use case here.  What are your thoughts @bsbeeks

lukeschunk:
Yes in terms of the screenshots, here is one where you can see the new 'track holder' element I added outlined in blue to handle the clicks, and the range component outlined in red. (obviously outlines just to show divs). Here is a video to display the actual clicking better: https://www.youtube.com/watch?v=B0npsL5HhOM&feature=youtu.be 


screen shot 2016-01-04 at 9 11 03 am

src/components/RangeSelector.js
bsbeeks: We should be ok with using `touches`. Kudos on considering it for touch devices.

src/components/RangeSelector.js
```diff
@@ -290,12 +356,19 @@ class RangeSelector extends React.Component {
           style={styles.range}
         >
           {this.props.presets.length ? <div className='mx-rangeselector-toggle' onClick={this._handleToggleViews.bind(this)} style={styles.showPresets}>Groups</div> : null}
+	<div
+          className='mx-rangeselector-track-holder'
+          onMouseDown={this._handleClickStartOnTrack.bind(this)}
+          onMouseUp={this._handleClickOnTrack.bind(this)}
```
bsbeeks: I think a better name for these methods would be:
`_handleTrackMouseDown`
`_handleTrackMouseUp`

src/components/RangeSelector.js
```diff
@@ -92,6 +92,68 @@ class RangeSelector extends React.Component {
...
+
+      //covert our 0-based value to actual value
+      const newValue = Math.round(newPosition / this.props.interval) * this.props.interval;
+
+      newState[this.state.dragging.toLowerCase() + 'Pixels'] = newPixels;
+      newState[this.state.dragging.toLowerCase() + 'Value'] = newValue;
+
+      this.setState(newState);
+
+      e.preventDefault();
```
bsbeeks: What is this preventing?

src/components/RangeSelector.js
cerinman: :+1: on the naming suggestion.  I like those better.

src/components/RangeSelector.js
```diff
@@ -92,6 +92,68 @@ class RangeSelector extends React.Component {
...
+
+    //click point is lower than upper selector && also greater than midway point between selectors
+    if (newPixels < this.state.upperPixels && newPixels > (this.state.upperPixels - (this.state.upperPixels - this.state.lowerPixels) / 2)) {
+      this.setState({
+        dragging: 'Upper'
+      });
+    }
+  }
+
+  _handleClickOnTrack (e) {
```
bsbeeks: This method looks almost identical to `_handleDragging`. Is there any way we can extract some of the shared logic into helper methods?

src/components/RangeSelector.js
bsbeeks: Somewhere in this method we need to trigger the `onLowerDragStop` or `onLowerDragStop` methods so parent components know about the update value.

src/components/RangeSelector.js
lukeschunk: will change!

src/components/RangeSelector.js
lukeschunk: yes the `_handleClickOnTrack` method is very similar to `_handleDragging` - i'll see what I can do

src/components/RangeSelector.js
lukeschunk: I just tested, and `_handleDragEnd` does fire on/after the mouseUp. I'm a bit confused - can you clarify what the `onLowerDragStop` and `onUpperDragStop` methods would do?

bsbeeks: So what's happening is the event is bubbling up, so the track `onMouseUp` event fires, then the `mx-rangeselector-range` div `onMouseUp` event fires. The result is we're setting state twice for one event (which we want to avoid). One way to avoid this would be to eliminate this method all together and rely on the `mx-rangeselector-range` div `onMouseUp` event to handle this logic. If you do this, then you'll want to set some type of state attribute in `_handleClickStartOnTrack` to indicate a track click happened.

The `onLowerDragStop` and `onUpperDragStop` methods are passed to this component through it's props. So they are defined by the parent component. When called, the parent component will be "notified" of the toggle value change. @cerinman can you think of a better way to explain this?

cerinman: Yep, `onLowerDragStop` and `onUpperDragStop` are call back functions supplied via props to this component by the parent.  These call back functions should be called when ever the lower or upper values change.  You pass the value to the callback function so that the parent is aware that the user changed the slider.

cerinman: Not that it's any better than your description.

lukeschunk: I definitely see how we're changing state twice on one event which isn't great. Thanks for the explanations - I'm still wrapping my head around React, but I think that all makes sense. I'll take another pass at everything? Should I open a new PR?

bsbeeks: No, you can go ahead and just modify/update this one.

lukeschunk:
### Updates:

* removed `onMouseUp` event from the `track holder` div, now `range` element handles the mouseUp
* added  `trackClicked` to state that is set to true when track is clicked
* `_handleDragEnd` method fires only once now, and thus `onLowerDragStop` or `onUpperDragStop` also fire

##### Eliminated `_handleClickOnTrack` method
* now `_handleDragging` handles the toggle movement when it is dragged, and when the track is clicked. Because I eliminated the method that was very similar to `_handleDragging`, I didn't break out anything to helper methods, however, this means that some of the code in `_handleDragging` is 'unused' when the track is clicked - is this an issue?
* Since this method is now handling more than just dragging, should it be renamed to something like: `_handleToggleMove` ?

src/components/RangeSelector.js
```diff
@@ -92,6 +93,45 @@ class RangeSelector extends React.Component {
     });
   }
 
+  _handleTrackMouseDown (e) {
+    this.setState({
+      trackClicked: true
+    });
```
bsbeeks: Setting state this way will cause 2 state updates with the conditionals below. A way we've gotten around this is to put the updated state into a `const`, then only call `setState` once at the end of the method:
```javascript
_handleTrackMouseDown (e) {
  const updatedState = {
    trackClicked: true
  };

  // ...stuff

  if (newPixels < this.state.lowerPixels) {
    updatedState.dragging = 'Lower';
  }

  // ..other stuff

  this.setState(updatedState);
}
```

bsbeeks:
I pulled this down to test it and it looks like the `onLowerDragStop` and `onUpperDragStop` methods are being passed the previous value of the corresponding toggle.

Default RangeSelector loaded
screenshot 2016-01-06 14 18 58

After clicking to the right of the upper toggle. You'll notice it changed to `50` in the UI, but the value passed to `onUpperDragStop` was the previous value of `30`.
screenshot 2016-01-06 14 19 06

lukeschunk:
ahhh you're right, I should have caught that `onLowerDragStop` and `onUpperDragStop` are being passed the previous value of the toggle. Can I take another stab?

bsbeeks:
For sure, I think this is really close to being merged! :checkered_flag:

lukeschunk:
great! :ok_hand:

lukeschunk:
### `onLowerDragStop` and `onUpperDragStop` are now getting passed the correct value of the toggle both when the track is clicked and when the toggle is dragged

* I moved the passing of `upper/lowerValue` to the props into a callback within `_handleDragging()`, this way the state updates before it is passed to props 

```
this.setState(newState, () => {
        this.props['on' + this.state.dragging + 'DragStop'](this.state[this.state.dragging.toLowerCase() + 'Value']);
        
        if (this.state.trackClicked) {
          this.setState({
            dragging: false,
            trackClicked: false
          });
        }
      });
```

* Since `_handleDragEnd` serves double duty - ie. it ends the toggle movement if you're dragging, but starts the toggle movement if you're clicking the track, I added conditionals to account for this:
```
  _handleDragEnd (e) {
    if (this.state.trackClicked) {
      this._handleDragging(e);
    } else {
      this.setState({
        dragging: false,
        trackClicked: false
      });
    }
  }
```
:octocat: let me know what you guys think!

src/components/RangeSelector.js
```diff
@@ -92,6 +93,47 @@ class RangeSelector extends React.Component {
     });
   }
 
+  _handleTrackMouseDown (e) {
```
bsbeeks: I played around with this method a bit and came up with a more condensed version of your changes that I think we should go with:
```javascript
_handleTrackMouseDown (e) {
  const clientX = e.touches ? e.touches[0].clientX : e.clientX;
  const newPixels = clientX - ReactDOM.findDOMNode(this.refs.rangeSelector).getBoundingClientRect().left;
  const updatedState = {
    trackClicked: true
  };
  const clickBelowLower = newPixels < this.state.lowerPixels;
  const clickAboveUpper = newPixels > this.state.upperPixels;
  const clickCloserToLower = newPixels > this.state.lowerPixels && newPixels < (this.state.lowerPixels + (this.state.upperPixels - this.state.lowerPixels) / 2);
  const clickCloserToUpper = newPixels < this.state.upperPixels && newPixels > (this.state.upperPixels - (this.state.upperPixels - this.state.lowerPixels) / 2);

  if (clickBelowLower || clickCloserToLower) {
    updatedState.dragging = 'Lower';
  }

  if (clickAboveUpper || clickCloserToUpper) {
    updatedState.dragging = 'Upper';
  }

  this.setState(updatedState);
}
```

src/components/RangeSelector.js
```diff
@@ -130,20 +172,30 @@ class RangeSelector extends React.Component {
...
+      this.setState(newState, () => {
+        this.props['on' + this.state.dragging + 'DragStop'](this.state[this.state.dragging.toLowerCase() + 'Value']);
+
+        if (this.state.trackClicked) {
+          this.setState({
+            dragging: false,
+            trackClicked: false
+          });
+        }
+      });
```
bsbeeks: Played around with this as well, and we can get rid of the `setState` callback method. Not only do we get rid of the callback, but it makes it a little more readable:
```javascript
if (this.state.trackClicked) {
  newState.dragging = false;
  newState.trackClicked = false;
}

this.props['on' + this.state.dragging + 'DragStop'](newState[this.state.dragging.toLowerCase() + 'Value']);

this.setState(newState);
```

lukeschunk: oh nice - I was thinking that since `setState()` is async (right?) the callback would be necessary. Thanks for the feedback I'll fix this today.

lukeschunk: I tried this, and the logs you see below are logging the previous value of the upper and lower toggle when the track is clicked - is that an issue? 

```
	  if (this.state.trackClicked) {
  		newState.dragging = false;
  		newState.trackClicked = false;
	  }
	  
	  //Logs previous value of the toggle, when track is clicked
	  console.log("this is this.state.upperValue", this.state.upperValue); 
	  console.log("this is this.state.lowerValue", this.state.lowerValue);
	  
	  this.props['on' + this.state.dragging + 'DragStop'](newState[this.state.dragging.toLowerCase() + 'Value']);

	  this.setState(newState);
```
screen shot 2016-01-14 at 11 24 59 am

bsbeeks: You're console logging `this.state.upperValue`, you should be console logging `newState.upperValue`. That's the updated value and what we're passing to the props event handler method.

lukeschunk: Great, thanks for clarifying :+1:

src/components/RangeSelector.js
lukeschunk: Great, will change.

PR #135, opened 2016-01-06
Allow for adding content to footer of Modal component

cerinman:
This PR adds a new prop to the Modal component to allow for passing in jsx for populating the footer with content.  This was done after a need to allow text links in the footer.

New Prop:

footerContent: type(node) - used to add jsx to footer of the component.

@mxenabled/frontend-engineers 

Large Modal

With tool tip:
screen shot 2016-01-06 at 1 35 36 pm

With out tool tip:
screen shot 2016-01-06 at 1 36 03 pm

No footer content:
screen shot 2016-01-06 at 1 40 05 pm

Small Modal

With tool tip:
screen shot 2016-01-06 at 1 36 25 pm

With out tool tip:
screen shot 2016-01-06 at 1 36 42 pm

No footer content:
screen shot 2016-01-06 at 1 40 19 pm


src/components/Modal.js
```diff
@@ -73,7 +82,7 @@ class Modal extends React.Component {
   _renderTooltipIconAndLabel () {
     if (this.props.tooltip) {
       return (
-        <div>
+        <div className='mx-modal-tooltip-label' style={styles.tooltipLabel}>
```
bsbeeks: Somewhat related...we don't need the else statement below. If there is no tooltip then just don't do anything in this method.

cerinman: will remove the else block

bsbeeks:
after you remove the else block, this is good to :shipit:

jmophoto:
:+1:

PR #136, opened 2016-01-08
Loader text

lukeschunk:
### Added Ability to Pass in Custom Text to Loader
* I might be oversimplifying this or missing something, but:
* I replaced the hard coded 'LOADING...' string with `{this.props.loaderText}` in the `render()` method
* One thing I noticed when I put a console log at the top of the `render()` function is it's logging twice - probably due to the 'live reload enabled', but I don't know why it was reloading?
* Right now the ellipses `...` is not hard coded, but is that something you'll want no matter what the text is?
screen shot 2016-01-08 at 3 22 29 pm

### DIFF
--- a/src/components/Loader.js
+++ b/src/components/Loader.js
@@ -6,6 +6,7 @@ const StyleConstants = require('../constants/Style');
 
 class Loader extends React.Component {
   render () {
+	console.log("this is this.props.loaderText", this.props.loaderText);
     if (this.props.isLoading) {
       const styles = {
         component: {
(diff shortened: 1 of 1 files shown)


src/components/Loader.js
```diff
@@ -75,14 +75,16 @@ Loader.propTypes = {
   color: React.PropTypes.string,
   isLoading: React.PropTypes.bool,
   isRelative: React.PropTypes.bool,
-  isSmall: React.PropTypes.bool
+  isSmall: React.PropTypes.bool,
+  loaderText: React.PropTypes.string
```
bsbeeks: Since we already have the context of the component name, `<Loader>`, I think we can rename this to just `text`, or better yet we can use `this.props.children` and let the user pass in whatever they want (text, numbers, jsx).

lukeschunk: sounds good - I'll get to this soon.

PR #137, opened 2016-01-12
Fix TimeBasedLineChart Tooltip position, Hover State, and style issues

cerinman:
This PR corrects a few bugs with the TimeBasedLineChart.

- The component was updating two to three times on hover of a slice.  The multiple renders where causing some funkyness.  This fixes that issue my implementing the `shouldUpdateComponent` lifecycle method.  We now only update the chart if this.props.data or this.state.hoverData is not equal to the previous state.
- The hovered rectangle was no longer being rendered on mouse over.  This fixes that by moving the `this._renderSvgTooltipComponents` call from `this._renderTooltip` function to the `_handleDataPointMouseOver` function where it actually has correct reference to everything it needs to render properly.
- Tooltips where not positioned correctly next to the hovered rectangle slice since are last set of fixes where we fixed spacing issues.  This fixes that issue by removing the magic `10` numbers that were part of the equation for figuring out the left and right position of the tooltip
- And finally, the styles for the chart needed adjusted for text color/font size to match designs.  While I was updating the text color and font size I went ahead and cleaned up the styles objectby removing px and using style constants where required.

@mxenabled/frontend-engineers 

screen shot 2016-01-11 at 4 59 19 pm
screen shot 2016-01-11 at 4 59 26 pm

### DIFF
--- a/src/components/TimeBasedLineChart.js
+++ b/src/components/TimeBasedLineChart.js
@@ -280,14 +280,14 @@ class TimeBasedLineChart extends React.Component {
       const isPast = i === 0;
 
       const group = chart.append('g')
-        .attr('class', 'group-' + i)
+        .attr('class', 'group')
         .attr('transform', 'translate(' + this.props.margin.left + ',' + this.props.margin.top + ')');
 
       //Lines ================================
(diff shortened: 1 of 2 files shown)


bsbeeks:
:shipit:

PR #139, opened 2016-01-13
Add slide and fade transitions

jmophoto:
This adds slide up/down and fade in/out transitions components.

To use, wrap the content to be toggled in <Transition> tags wit isShown, type and duration params. Defaults to false, slide and 500ms.

This is ported from the Robinhood repo where it is used to toggle tables open and closed. It uses the react animation life-cycle and removes the component on unmount, which is good.

It uses TransitionGroup, which like all react addons is listed as experimental, so that could be an issue.

I've looked into other ways to animate and here are some options;
1. Not use the TransitionGroup and instead alter the state at an interval, changing a specified value
2. Use a outside module such as react-tween-state or react-motion, which pretty much does the above

I am open to any thoughts on this.

@mxenabled/frontend-engineers

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -588,7 +588,7 @@ const Demo = React.createClass({
           <span style={styles.button} onClick={this._handleTransitionClick.bind(null, 'slide')}>Slide Transition</span>
         </div>
         <br/><br/>
-        <Transition shouldAnimate={this.state.showSlideTransition} type='slide'>
+        <Transition isShown={this.state.showSlideTransition} type='slide'>
             <div style={{
               backgroundColor: '#F7F8F8',
               border: '1px solid #E3E6E7' ,
@@ -605,7 +605,7 @@ const Demo = React.createClass({
           <span style={styles.button} onClick={this._handleTransitionClick.bind(null, 'fade')}>Fade Transition</span>
         </div>
         <br/><br/>
-        <Transition shouldAnimate={this.state.showFadeTransition} type='fade'>
+        <Transition isShown={this.state.showFadeTransition} type='fade'>
             <div style={{
               backgroundColor: '#F7F8F8',
               border: '1px solid #E3E6E7' ,
--- a/demo/app.js
+++ b/demo/app.js
@@ -12,6 +12,7 @@ const {
   SelectFullScreen,
   TimeBasedLineChart,
   ToggleSwitch,
+  Transition,
   TypeAhead,
   DatePicker,
   DatePickerFullScreen
@@ -447,6 +448,9 @@ const Demo = React.createClass({
       selectedDatePickerDate: moment().unix(),
       showModal: false,
       showSmallModal: false,
+      showFadeTransition: false,
+      showSlideTransition: false,
+      transitionType: 'slide',
       windowWidth: document.documentElement.clientWidth || document.body.clientWidth
     }
   },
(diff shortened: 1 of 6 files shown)


demo/app.js
```diff
@@ -569,6 +584,40 @@ const Demo = React.createClass({
         </div>
 
         <br/><br/>
+        <div style={{ textAlign: 'center', fontFamily: 'Helvetica, Arial, sans-serif' }}>
```
tumentumurchudur: I would extract this out into a style object.

demo/app.js
```diff
@@ -569,6 +584,40 @@ const Demo = React.createClass({
...
+          <span style={styles.button} onClick={this._handleTransitionClick.bind(null, 'slide')}>Slide Transition</span>
+        </div>
+        <br/><br/>
+        <Transition isShown={this.state.showSlideTransition} type='slide'>
+            <div style={{
```
tumentumurchudur: Same here. I would move these into a style object.

demo/app.js
```diff
@@ -569,6 +584,40 @@ const Demo = React.createClass({
...
+              boxSizing: 'border-box',
+              height: 400,
+              padding: '10px',
+              position: 'relative',
+              width: '100%'
+            }}></div>
+        </Transition>
+
+        <br/><br/>
+        <div style={{ textAlign: 'center', fontFamily: 'Helvetica, Arial, sans-serif' }}>
```
tumentumurchudur: Same here

demo/app.js
```diff
@@ -569,6 +584,40 @@ const Demo = React.createClass({
...
+          <span style={styles.button} onClick={this._handleTransitionClick.bind(null, 'fade')}>Fade Transition</span>
+        </div>
+        <br/><br/>
+        <Transition isShown={this.state.showFadeTransition} type='fade'>
+            <div style={{
```
tumentumurchudur: Same here

demo/app.js
```diff
@@ -483,10 +487,21 @@ const Demo = React.createClass({
     })
   },
 
+  _handleTransitionClick (transitionType) {
```
tumentumurchudur: I think this is more of a toggle function. I would break it up into two functions because I feel like this function is doing two things. A function should only do one thing.
```javascript
_toggleSlideTransition () {
  this.setState({
    showSlideTransition: !this.state.showSlideTransition;
  })
},

_toggleFadeTransition () {
  this.setState({
    showFadeTransition: !this.state.showFadeTransition;
  })
}
```

src/components/Transition.js
```diff
@@ -0,0 +1,46 @@
...
+
+class Transition extends React.Component {
+  constructor (props) {
+    super(props);
+  }
+
+  _renderTransition () {
+    switch (this.props.type) {
+      case 'slide':
+        return (<Slide {...this.props} />);
```
tumentumurchudur: Not sure if you need `()`

src/components/Transition.js
```diff
@@ -0,0 +1,46 @@
...
+      case 'fade':
+        return (<Fade {...this.props} />);
```
tumentumurchudur: Probably don't need `()`

bsbeeks:
Recording an offline discussion I had with @jmophoto. Before we commit any code we are going to explore a couple different ways of handling animations in React. Once we decide on a general direction I'd like to test it out internally before releasing anything to this repo.

PR #145, opened 2016-01-14
Fix Modal component button alignment on no footerContent

cerinman:
@mxenabled/frontend-engineers 

When I added the footerContent prop to the Modal component I caused an issue with the buttons alignment in the footer when no footerContent was provided.  This fixes the issue by always rendering the footerContent div even if it is empty to ensure that flex box aligns things correctly in the footer.

 Before:
screen shot 2016-01-14 at 3 42 14 pm

After:
screen shot 2016-01-14 at 3 42 32 pm
screen shot 2016-01-14 at 3 43 05 pm
screen shot 2016-01-14 at 3 43 19 pm

### DIFF
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -57,7 +57,7 @@ class Modal extends React.Component {
   _renderFooterContent () {
     return (
       <div className='mx-modal-footer-content' style={styles.footerContent}>
-        {this.props.footerContent ? this.props.footerContent : null}
+        {this.props.footerContent}
       </div>
     );
   }
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -227,7 +227,6 @@ const styles = {
     textAlign: 'left'
   },
   buttons: {
-    alignSelf: 'flex-end',
     textAlign: 'right'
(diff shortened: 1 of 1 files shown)


paniclater:
:shipit:

src/components/Modal.js
```diff
@@ -55,13 +55,11 @@ class Modal extends React.Component {
...
-    if (this.props.footerContent) {
-      return (
-        <div className='mx-modal-footer-content' style={styles.footerContent}>
-          {this.props.footerContent}
-        </div>
-      );
-    }
+    return (
+      <div className='mx-modal-footer-content' style={styles.footerContent}>
+        {this.props.footerContent ? this.props.footerContent : null}
```
cerinman: haha I don't have to do the check do I.  If's the prop is null already who cares.  Let me try it with out the check.

bsbeeks: yep, you're right

cerinman:
Removed the check around `footerContent`.  Ready to review again.

bsbeeks:
:shipit:

PR #146, opened 2016-01-15
Button: Add icon prop and outline types

bsbeeks:
screenshot 2016-01-15 08 38 09

### propTypes
- `icon` *String*: can be any value available in the Icon component

// @mxenabled/frontend-engineers

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -3,10 +3,13 @@ const Radium = require('radium');
 
 const StyleConstants = require('../constants/Style');
 
+const Icon = require('../components/Icon');
+
 const StyleUtils = require('../utils/Style');
 
 const Button = React.createClass({
   propTypes: {
+    icon: React.PropTypes.string,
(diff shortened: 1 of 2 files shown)


src/components/Button.js
```diff
@@ -22,8 +33,11 @@ const Button = React.createClass({
     const styles = {
       component: {
         borderRadius: 2,
+        borderStyle: 'solid',
+        borderWidth: 1,
+        borderColor: 'transparent',
         display: 'inline-block',
-        padding: '7px 14px',
+        padding: '10px 14px',
```
cerinman: so this doesn't line up with the style guide now.  https://app.frontify.com/d/KHQZnlj1XMIO/ui-library-web#/ui-components/buttons  Is that ok?

bsbeeks: cleared it with product - we'll be changing the style guide

cerinman:
:shipit:

PR #148, opened 2016-01-25
changed loader to react.createclass

lukeschunk:
### Changed Loader to `React.createClass` syntax
* moved `getDefaultProps()` and `propTypes` inside of React.createClass
* changed it to `const Loader` 
* followed these guidelines for order: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-comp.md

### DIFF
--- a/src/components/Loader.js
+++ b/src/components/Loader.js
@@ -4,7 +4,26 @@ const Spin = require('./Spin');
 
 const StyleConstants = require('../constants/Style');
 
-class Loader extends React.Component {
+const Loader = React.createClass({
+
+  propTypes: {
+	  color: React.PropTypes.string,
+	  isLoading: React.PropTypes.bool,
+	  isRelative: React.PropTypes.bool,
+	  isSmall: React.PropTypes.bool
+  },
+
+  getDefaultProps() {
+	  return {
+		  color: StyleConstants.Colors.PRIMARY,
+		  isLoading: false,
+		  isRelative: false,
+		  isSmall: false,
+		  children: 'LOADING...'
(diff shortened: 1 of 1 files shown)


src/components/Loader.js
```diff
@@ -4,7 +4,26 @@ const Spin = require('./Spin');
 
 const StyleConstants = require('../constants/Style');
 
-class Loader extends React.Component {
+const Loader = React.createClass({
+
```
bsbeeks: Remove this line break

src/components/Loader.js
```diff
@@ -69,21 +88,7 @@ class Loader extends React.Component {
...
 
-Loader.defaultProps = {
-  color: StyleConstants.Colors.PRIMARY,
-  isLoading: false,
-  isRelative: false,
-  isSmall: false,
-  children: 'LOADING...'
-};
 
 module.exports = Loader;
```
bsbeeks: Indentation is off with all your changes. We use 2 spaces for indentation.

lukeschunk: Sorry about that. I always have that issue for some reason even though I switched it in my atom. I usually just go back through on github and fix the changes...

bsbeeks:
Make sure it passes Travis.

src/components/Loader.js
bsbeeks: You should be able to set your default spacing in Atom to 2 spaces (no tabs).

lukeschunk:
@bsbeeks I will :ok_hand:

Issue #142, opened 2016-01-14
Convert components back to using createClass instead of ES6 classes


bsbeeks:
We are currently using `createClass` in all of our internal projects at MX. Jumping between the 2 syntaxes is a bit confusing...especially for newbies. For not let's move everything back to `createClass` until we convert our internal projects to ES6.

lukeschunk:
I can work on this once the craziness of this week is over for me. I'll check the docs, but when we make that change, we move all the default prop types and validation inside of the class, right?

jmophoto:
Components to be converted
- [ ] DatePicker
- [ ] DatePickerFullScreen
- [ ] DonutChart
- [ ] Icon
- [ ] Loader
- [ ] Modal
- [ ] RajaIcon
- [ ] RangeSelector
- [ ] Select
- [ ] SelectFullScreen
- [ ] Spin
- [ ] TimeBasedLineChart
- [ ] ToggleSwitch
- [ ] TypeAhead

cerinman:
I'm revamping the `TimeBasedLineChart` component now so I can handle that one with the large changes that I'm already making.

iheartrachie:
Can this be closed? I think we're reverted everything.

cerinman:
Yes. This appears to be done.

PR #151, opened 2016-01-27
Fix setState calls when using props

jmophoto:
Calling this.props in a component constructor in other browsers works, but in IE10 it returns null. This was causing an error for any of those components in IE10. 

This PR changes this.props to props in the constructors.

There were also a few that called a method that referenced this.props, so I have set props to default to this.props if not included and then included props in the constructor. (See DatePicker.js)

There was also a duplicate method in app.js for _handleSelectChange, so removed that.

The demo now runs in IE10 and all of the components changed were tested in Chrome, Safari and IE11.

@mxenabled/frontend-engineers

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -717,12 +717,6 @@ const Demo = React.createClass({
     });
   },
 
-  _handleSelectChange (option) {
-    this.setState({
-      icon: option
-    });
-  },
-
   _handleModalClick () {
     this.setState({
       showModal: true,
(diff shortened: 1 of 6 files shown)


jmophoto:
On a test, it does appear that on converting components to the createClass pattern also fixes this. It may be better to move ahead with #142 instead of this approach.

bsbeeks:
I think that's a better solution since we already want to move down that path.

jmophoto:
@bsbeeks: agreed. Closing this and will head that direction.

PR #153, opened 2016-01-27
Revert ToggleSwitch to createClass

jmophoto:
Part of #142 

- Revert to createClass
- Move propTypes and defaultProps
- setState
- Remove .bind

@mxenabled/frontend-engineers


cerinman:
:shipit:

PR #156, opened 2016-01-27
Revert DatePickerFullScreen to createClass

jmophoto:
Part of #142 

Revert to createClass
Moved propTypes and defaultProps
SetState
Remove .bind(this)
Change .bind(this, to .bind(null,

@mxenabled/frontend-engineers


cerinman:
:shipit:

PR #159, opened 2016-01-28
Revert RajaIcon to createClass

jmophoto:
Part of #142

Revert to createClass
Moved propTypes and defaultProps

@mxenabled/frontend-engineers


cerinman:
:shipit:

PR #166, opened 2016-01-29
Select: Removed opacity of options

tumentumurchudur:
- Removed opacity of options in the Select component

Before
screen shot 2016-01-29 at 10 03 53 am

After
screen shot 2016-01-29 at 10 04 31 am


jmophoto:
:shipit:

PR #167, opened 2016-01-31
create new Input component

JordanSpencerWu:
This add a new `Input` component that support the following types `text`, `number`, `email`, and `password`. 

## Goal 
Make it simple and easy to use. 

## States
`value` _String_
This `input` component only has one state which is the value.

## Props
`placeholderText` _String_ **optional**
A string that will be used as a placeholder in the `input` component. 

`label` _String_ **optional**
A string that will be used as a label for the `input` component.

`onTextChange` _Callback_ **required**
A callback that will be invoked when the value of the `input` component is changed. 

`type` _String_ **optional**
A string that determines one of four types of input, `text`, `number`, `password`, or `email`.

`icon` _String_ **optional**
A string that determines which icon will be used with this input component, this will create an icon from the `icon` component (must be a valid type of icon).

`options` _Array_ **optional**
A array of strings that will be used for input attributes `disabled` or `required`.

`autofocus` _Boolean_ **optional**
A boolean that determines whether or not the `input` component will be autofocus.

`prefix` _String_ **optional** (forgot to add to PropTypes)
A string that adds a prefix in front of the `input` component.

`suffix` _String_ **optional** (forgot to add to PropTypes)
A string that adds a suffix at end of the `input` component.

`button` _String_ **optional**
A string that determines if a `submit` or `clear` button will be added to the `input` component.

`onSubmit` _Callback_ **optional**
A callback that will be used for the submit button, this will handle the submit.

## Examples
default:
```
<Input onTextChange={this._handleTextInputChange} />
```
final:
```
<Input 
  icon='user'
  type='number'
  prefix='$'
  suffix='.00'
  label='Final Example'
  onTextChange={this._handleMoneyInputChange}
  button='submit'
  onSubmit={this._handleMoneySubmit}
  placeholderText='Enter Amount'
/>
```

More examples will be in `demo/app.js` file.

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -15,7 +15,8 @@ const {
   ToggleSwitch,
   TypeAhead,
   DatePicker,
-  DatePickerFullScreen
+  DatePickerFullScreen,
+  Input
 } = require('../src/Index');
 
 const styles = {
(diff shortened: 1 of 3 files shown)


JordanSpencerWu:
I saw that my code didn't pass the npm test and make another commit to pass the test.

PR #174, opened 2016-02-05
Select: No shadow onClick when there is no options

tumentumurchudur:
https://git.moneydesktop.com/product/MoneyDesktop/issues/327

- No shadow when there is no options to show onClick

Before
screen shot 2016-02-05 at 3 11 29 pm

After
screen shot 2016-02-05 at 3 12 01 pm


jmophoto:
:shipit:

PR #175, opened 2016-02-08
Tommy/revamp timebasedlinechart

cerinman:
*BREAKING CHANGE TO TimeBasedLineChart*

This PR attempts to address the issues around how we were handling D3 inside of React.  This change addresses the following

- Componentizes the various pieces of the chart for reuse
- Lets React do the rendering while still using D3 to figure positioning/sizing/styling.  
- Resolves some of the sticky issues that were preventing us from doing transition animations via D3.
- Addresses style changes from product.

I've spent the weekend testing these changes both in the demo app as well as in our internal application.  I'm satisfied that this change has a very positive outcome for for how we write D3 charts and their ability to scale.

`---------------------------------------------------------------------------`

Component: TimeBasedLineChart 

Props

- `breakPointDate` - Type(`number(unix time stamp)`) - Default (`Today's Date as a unix time stamp`) - A date used to draw a vertical line on the graph to indicate one section of the graph from the other.
- `breakPointLabel` - Type(`string`) - Default(`Today`) - A text label drawn to the right of the break point line.
- `data` - Type(`array(objects)` - isRequired - Array of objects that define the shape of the line and x/y axis ranges.  Each object must have at least an x key/value and and y key/value as shown in the example.

Example of an object in the `data` array

```javascript
{
  x: 1447890110,  // Unix time stamp for date along x axis
  y: 800 // Number value for y axis
}
```

- `height` - Type(`number`) - Default(`400`) - The height of the chart.
- `hoveredDataPointDetails` - Type(`array(objects)`) - An array of objects used to build the the hovered data point tool tip box below the chart.

Example Object for `hoveredDataPointDetails`
```javascript
{
  format: 'MMM DD, YYYY', // Format used to format the value
  key: 'x', // The key used to get the appropriate value from the data set
  label: 'Date', // The label to be used for the value on the tool tip
  type: 'date' // Type determine the js library to be used to format the value
}
```

- `lineColor` - Type(`string`) - Default(`#359BCF`) - The color used for the line on the chart.
- `margin` - Type(`object`) - Default(`{ top: 20, right: 20, bottom: 20, left: 75 }`) - The margins used to determine the spacing around the chart.
- `rangeType` - Type(`string ('day', 'month')`) - Default(`day`) - A string used to normalize the data points to month or day intervals.
- `shadeFutureOnGraph` - Type(`boolean`) - Default(`true`) - Boolean used to determine if any future part of the graph(past today) should be shaded.
- `showBreakPoint` - Type(`boolean`) - Default(`true`) - Boolean used to determine if the break point line and label should be shown.
- `width` - Type(`number`) - Default(`550`) - The width of the chart.
- `yAxisFormatter` - Type(`function`) - Default(`(d) { return numeral(d).format('0.0a'); }`) - Callback function used to format the y axis tick labels on the cart.
- `zeroState` - Type('node`) - Default(```html<div style={styles.zeroState}>No Data Found</div>```) - Child jsx node used in place of the chart of the supplied data has no length.

Example usage

```javascript
<TimeBasedLineChart
  breakPointDate={moment().startOf('day').unix()}
  breakPointLabel={'Today'}
  data={this.state.lineChartData}
  height={400}
  hoveredDataPointDetails={[
    {
      format: 'MMM DD, YYYY',
      key: 'x',
      label: 'Date',
      type: 'date'
    },
    {
      format: '$0,0',
      key: 'y',
      label: 'Value',
      type: 'number'
    }
  ]}
  rangeType={'day'}
  showBreakPoint={true}
  width={700}
/>
```

@mxenabled/frontend-engineers

Screen shots - showing two resolutions to show how the chart now scales much better

400 x 700 Chart
screen shot 2016-02-07 at 6 51 52 pm

200 x 500 Chart
screen shot 2016-02-07 at 7 43 07 pm

`---------------------------------------------------------------------------`

Component: AxisGroup

Used to draw an axis for a chart with it's associated ticks and tick labels

Props

`axis` - Type(`string(x, y)`) - `isRequired` - Used to determine if the axis will be the x or y axis.
`axisFormatFunction` - Type(`function`) - `isRequired` - A call back function used to format the axis tick labels.
`data` - Type(`array(objects)`) - `isRequired` - Data array that matches the same layout as described in the TimeBasedLineChart component.
`orientation` - Type(`string`) - `isRequired` - A string used to determine which side of the axis to draw the ticks and tick labels.
`scaleFunction` - Type('function`) - `isRequired` - A D3 scale function used to determine the scale of the axis.
`translation` - Type(`string`) - Default(`translate(0,0)`) - A string used to transform the position of the axis

Example

```javascript
<AxisGroup
  axis='y'
  axisFormatFunction={this.props.yAxisFormatter}
  data={this.props.data}
  orientation='left'
  scaleFunction={this._getYScaleFunction}
  translation={this._getYAxisTranslation()}
/>
```

Screenshot

screen shot 2016-02-08 at 8 44 09 am

`---------------------------------------------------------------------------`

Component: CirclesGroup

Props

`adjustedHeight` -Type(`number`) - `isRequired` - A height value used to determine the base of the line graph or 0 point of the y axis.
`circleRadius` - Type(`number`) - Default(`3`) - A number value used to determine the radius of each circle drawn.
`data - Type(`array(objects)`) - `isRequired` - Data array that matches the layout described in the TimeBasedLineChart component.
`shouldAnimate` - Type(`boolean`) - Default(`true`) -  A boolean used to determine of the circles should be animated on load from the y axis 0 point to their final height on the chart.
`translation` - Type(`string`) - Default(`translate(0,0)`) - A string used to transform the circles.
`xScaleValueFunction` - Type(`function`) - `isRequired` - A D3 scale function used to determine where to place a circle on the x axis of the graph.
`yScaleValueFunction` - Type(`function`) - `isRequired` - A D3 scale function used to determine where to place a circle on the y axis of the graph.

Example

```javascript
<CirclesGroup
  adjustedHeight={this.state.adjustedHeight}
  data={this.props.data}
  translation={this._getLineTranslation()}
  xScaleValueFunction={this._getXScaleValue}
  yScaleValueFunction={this._getYScaleValue}
/>
```

Screenshot

screen shot 2016-02-08 at 8 57 06 am

`---------------------------------------------------------------------------`

Component: GridLinesGroup

Props

`axis` - Type(`string`) - `isRequired` - Used to determine if the grid lines will be based upon the x or y axis.
`data` - Type(`array(objects)`) - `isRequired` - Data array that matches the layout described in the TimeBasedLineChart component.
`orientation` - Type(`string`) - Default(`left`) - A string used to determine which side of the given axis to draw the grid lines.
`scaleFunction` - Type(`function`) - `isRequired` - A D3 scale function used to determine the scale of the grid lines.
`tickSize` - Type(`number`) - `isRequired` - A numeric value used to determine the lines width or height depending on the axis used.
`translation` - Type(`string`) - Default(`translate(0,0)`) - A string used to transform the grid lines if needed.

Example

```javascript
 <GridLinesGroup
   axis='y'
   data={this.props.data}
   orientation='left'
   scaleFunction={this._getYScaleFunction}
   tickSize={this.state.adjustedWidth * -1}
   translation={this._getYAxisTranslation()}
/>
```

Screenshot
screen shot 2016-02-08 at 9 24 01 am

`---------------------------------------------------------------------------`

Component: BreakPointGroup

Props

`adjustedHeight` - Type(`number`) - `isRequired` - A height value used to determine the base of the line graph or 0 point of the y axis.
`breakPointDate` - Type(`number`) - `isRequired` - A unix time stamp that determines where on the chart to draw the break point vertical line.
`breakPointLabel` - Type(`string`) - `isRequired` - A string used for drawing the break point label next to the break point line on the chart.
`margin` - Type(`object`) - `isRequired` - An object with top, right, bottom, and left margin values of the chart.  Used to further place the break point line correctly on the chart.
`translation` - Type(`string`) - Default(`translate(0,0)`) - A string used to transform the break point line and label.
`xScaleValueFunction` - Type(`function`) - `isRequired` - A D3 scale function used to determine where to place the break point line and label on the x axis of the graph.

Example

```javascript
<BreakPointGroup
  adjustedHeight={this.state.adjustedHeight}
  breakPointDate={this.props.breakPointDate}
  breakPointLabel={this.props.breakPointLabel}
  margin={this.props.margin}
  translation={this._getVerticalLineTranslation()}
  xScaleValueFunction={this._getXScaleValue}
/>
```

Screenshot

screen shot 2016-02-08 at 9 52 21 am

`---------------------------------------------------------------------------`

Component: LineGroup

Props

`adjustedHeight` - Type(`number`) - `isRequired` - A height value used to determine the base of the line graph or 0 point of the y axis.
`data` - Type(`array(objects)`) - `isRequired` - Data array that matches the layout described in the TimeBasedLineChart component.
`lineColor` - Type(`string`) - Default(`#56595A` CHARCOAL) - A string(hex color) used to determine the color of line.
`shouldAnimate` - Type(`boolean`) - Default(`true`) - A boolean used to determine if the line should be animated on load from the zero point of the y axis up to the lines actual path.
`translation` - Type(`string`) - Default(`translate(0,0)`) - A string used to transform the line.
`xScaleValueFunction` - Type(`function`) - `isRequired` - A D3 scale function used in determining the lines path.
`yScaleValueFunction` - Type(`function`) - `isRequired` - A D3 scale function used in determining the lines path.

Example

```javascript
<LineGroup
  adjustedHeight={this.state.adjustedHeight}
  data={this.props.data}
  lineColor={this.props.lineColor}
  translation={this._getLineTranslation()}
  xScaleValueFunction={this._getXScaleValue}
  yScaleValueFunction={this._getYScaleValue}
/>
```

Screenshot

screen shot 2016-02-08 at 10 02 39 am

`---------------------------------------------------------------------------`

Component: ShadedRectangleGroup

Props

`fillColor` - Type(`string`) - Default(`#E3E6E7` FOG) - A string(hex color) to be used for the shade color of the rectangle.
`height` - Type(`number`) - `isRequired` - A numerical value for the height of the rectangle.
`translation` - Type(`string`) - Default(`translate(0,0)`) -A string used to transform the rectangle.
`width` - Type(`number`) - `isRequired` - A numerical value for the width of the rectangle.
`x` - Type(`number`) - `isRequired` - A numerical value for placing the rectangle on the chart at the correct position on the x axis.
`y` - Type(`number`) - `isRequired` - A numerical value for placing the rectangle on the chart at the correct position on the y axis.

Example

```javascript
<ShadedRectangleGroup
  height={this.state.adjustedHeight}
  translation={this._getLineTranslation()}
  width={this._getShadedRectangleWidth()}
  x={this._getShadedRectangleXValue()}
  y={0}
/>
```

Screenshot

screen shot 2016-02-08 at 10 28 23 am

`---------------------------------------------------------------------------`

Component: TimeAxisGroup

Props

`numberOfTicks` - Type(`number`) - Default(`10`) - A numerical value used to determine the number of axis ticks.  Does not guarantee 10 ticks but will result in rough number being around 10.
`tickSize` - Type(`number`) - `isRequired` - React.PropTypes.number,
`timeAxisFormat` - Type(`string`) - Default(`6`) - A numerical value used to determine the length of each axis tick.
`translation` - Type(`string`) - Default(`translate(0,0)`) - A string used to transform the axis.
`xScaleFunction` - Type(`function`) - `isRequired` - A D3 scale function used in determining the axis tick and label spacing on the axis.

Example

```javascript
<TimeXAxisGroup
  data={this.props.data}
  timeAxisFormat={'MMM D'}
  translation={this._getTimeAxisTranslation()}
  xScaleFunction={this._getXScaleFunction}
/>
```

Screenshot

screen shot 2016-02-08 at 10 35 55 am

`---------------------------------------------------------------------------`

Component: SlicesGroup

A group of invisible rectangle slices that can be placed over a chart for use as a mouseover placeholder for each data point in the graph. 

Props

`adjustedHeight` - Type(`number`) - `isRequired` - A height value used to determine the base of the line graph or 0 point of the y axis.
`data` - Type(`array(objects)`) - `isRequired` -Data array that matches the layout described in the TimeBasedLineChart component.
`handleChartMouseOver` - Type(`function`) - `isRequired` - A call back function used to pass back to the parent component the details about the hovered slice.
`sliceWidth` - Type(`number`) - `isRequired` - A numeric value to determine the width of each slice in the group.
`translation` - Type(`string`) - `isRequired` - A string used to transform the slices group.
`xScaleValueFunction` - Type(`function`) - `isRequired` - A D3 scale function used in determining the placement of each slice on the x axis of the chart.

Example

```javascript
<SlicesGroup
  adjustedHeight={this.state.adjustedHeight}
  data={this.props.data}
  handleChartMouseOver={this._handleChartMouseOver}
  sliceWidth={this._getSliceWidth()}
  translation={this._getLineTranslation()}
  xScaleValueFunction={this._getXScaleValue}
/>
```

Screenshot

No screen shot for this one as it's invisible.


src/utils/Chart.js
```diff
@@ -0,0 +1,35 @@
+const d3 = require('d3');
+
+const Chart = {
+  getDataMinMaxValues (data, axis) {
+    const max = d3.max(data, d => {
+      return Math.ceil(d[axis] / 1000) * 1000;
```
tumentumurchudur: Isn't this mathematically same as `return Math.ceil(d[axis])`?

src/utils/Chart.js
```diff
@@ -0,0 +1,35 @@
...
+    });
+
+    const min = d3.min(data, d => {
+      return Math.floor(d[axis] / 1000) * 1000;
```
tumentumurchudur: Same here

src/utils/Chart.js
```diff
@@ -0,0 +1,35 @@
...
+    });
+
+    return { min, max };
+  },
+
+  getAxisTickValues (data, axis) {
+    const minMaxValues = this.getDataMinMaxValues(data, axis);
+    const range = minMaxValues.max - minMaxValues.min;
+    const tempStep = range / 6;
```
tumentumurchudur: What is `6`?

src/components/TimeBasedLineChart.js
```diff
@@ -1,32 +1,231 @@
...
+  },
+
+  render () {
+    return (
+      <g className='hover-state' ref='hoverState'>
+        <g className='hover-state-line' ref='hoverStateLine' transform={this.props.translation}>
+          <line
+            className='hovered-data-point-line'
+            x1={this.props.xScaleValueFunction(this.props.hoveredDataPoint.x)}
+            x2={this.props.xScaleValueFunction(this.props.hoveredDataPoint.x)}
```
tumentumurchudur: Just an idea I'm throwing out here. Instead of repeating `this.props.something`, I wonder if you could grab these props at the top:
```javascript
{ hoverDataPoint, rangeType } = this.props;
```
And then use these `{xScaleValueFunction(hoveredDataPoint.x)}`
`{moment.unix(hoveredDataPoint.x).format(rangeType === 'day' ? 'MMM DD' : 'MMM')}`

src/components/TimeBasedLineChart.js
```diff
@@ -1,32 +1,231 @@
...
+            y2={this.props.yScaleValueFunction(this.props.hoveredDataPoint.y)}
+          />
+        </g>
+        <g className='hover-state-date-rect' ref='hoverStateDateRect' transform={this.props.translation}>
+          <rect
+            className='hovered-data-point-date'
+            height={30}
+            width={60}
+            x={this.props.xScaleValueFunction(this.props.hoveredDataPoint.x) - 30}
+            y={this.props.adjustedHeight}
```
tumentumurchudur: What is `30`?

src/components/TimeBasedLineChart.js
```diff
@@ -1,32 +1,231 @@
...
+    zeroState: React.PropTypes.node
+  },
+
+  getDefaultProps () {
+    return {
+      breakPointDate: moment().startOf('day').unix(),
+      breakPointLabel: 'Today',
+      height: 400,
+      lineColor: StyleConstants.Colors.PRIMARY,
+      margin: { top: 20, right: 20, bottom: 20, left: 75 },
```
tumentumurchudur: Can you move this down inside the styles object?

src/components/TimeBasedLineChart.js
```diff
@@ -42,713 +241,291 @@ class TimeBasedLineChart extends React.Component {
...
+    this.setState({
+      hoveredDataPoint
     });
+  },
 
-    let minValue = d3.min(this.props.data, d => {
-      const value = d.below ? d.value + d.below : d.value - 1000;
-      const multiplier = value < 0 ? -1000 : 1000;
+  // Helper Functions
+  _getFormatedValue (value, type, format) {
```
tumentumurchudur: No biggie but you may want to fix the spelling here. It should be `_getFormattedValue`

src/components/TimeBasedLineChart.js
```diff
@@ -42,713 +241,291 @@ class TimeBasedLineChart extends React.Component {
...
+    let formattedValue = '';
+
+    switch (type) {
+      case 'date':
+        formattedValue = moment.unix(value).format(format);
+        break;
+      case 'number':
+        formattedValue = numeral(value).format(format);
+        break;
+      default:
```
tumentumurchudur: Glad, you chose to include the `default` case here. :+1:

src/components/d3/BreakPointGroup.js
```diff
@@ -0,0 +1,43 @@
...
+
+  render () {
+    const breakPointLabelOffSet = 10;
+
+    return (
+      <g className='break-point-items' ref='breakPointItems' transform={this.props.translation}>
+        <line
+          className='break-point-line'
+          x1={this.props.xScaleValueFunction(this.props.breakPointDate)}
+          x2={this.props.xScaleValueFunction(this.props.breakPointDate)}
```
tumentumurchudur: Did you mean to call `{this.props.xScaleValueFunction(this.props.breakPointDate)}` twice here for x1 and x2? If so, you could call it outside the `render` function and assign it to a variable and use it here.

src/components/d3/BreakPointGroup.js
```diff
@@ -0,0 +1,43 @@
...
+          y1={this.props.margin.top}
+          y2={this.props.adjustedHeight + this.props.margin.bottom}
+        />
+        <text
+          className='break-point-label'
+          x={this.props.xScaleValueFunction(this.props.breakPointDate) + breakPointLabelOffSet}
```
tumentumurchudur: Looks like you are calling `{this.props.xScaleValueFunction(this.props.breakPointDate)}` three times including this one.

src/components/d3/CirclesGroup.js
```diff
@@ -0,0 +1,66 @@
...
+      d3.select(this.refs.circleGroup).selectAll('.circle').data(this.props.data).transition().attr('cy', d => {
+        return this.props.yScaleValueFunction(d.y);
+      });
+    }
+  },
+
+  render () {
+    return (
+      <g className='circle-group' ref='circleGroup' transform={this.props.translation}>
+        {this.props.data.length <= 45 ? (
```
tumentumurchudur: So you cannot pass an array that has more than 45 elements? What is the limitation here?

src/components/d3/LineGroup.js
```diff
@@ -0,0 +1,78 @@
...
+
+  render () {
+    return (
+      <g className='chart-line-group' ref='chartLineGroup' transform={this.props.translation}>
+        <path
+          d={this.props.shouldAnimate ? this.state.flatLine(this.props.data) : this.state.line(this.props.data) }
+          fill='none'
+          ref='chartLine'
+          stroke={this.props.lineColor}
+          strokeWidth={2}
```
tumentumurchudur: May want to receive `strokeWidth` as a prop from the parent component.

src/components/TimeBasedLineChart.js
jmophoto: I go back and forth on my thinking on that pattern. Typically I use a const if I am reusing the same value, not for props in general. But Tumen's suggestion does feel cleaner.

src/components/d3/SlicesGroup.js
```diff
@@ -0,0 +1,35 @@
...
+        {this.props.data.map((dataPoint, index) => {
+          return (
+            <rect
+              height={this.props.adjustedHeight}
+              key={'slice-' + index}
+              onMouseOver={this.props.handleChartMouseOver.bind(null, dataPoint)}
+              opacity={0}
+              transform={this.props.translation}
+              width={this.props.sliceWidth}
+              x={this.props.xScaleValueFunction(dataPoint.x) - (this.props.sliceWidth / 2)}
```
tumentumurchudur: No need for `()`
`x={this.props.xScaleValueFunction(dataPoint.x) - this.props.sliceWidth / 2}`

src/components/d3/TimeXAxisGroup.js
```diff
@@ -0,0 +1,54 @@
...
+    this._renderAxis();
+  },
+
+  _renderAxis () {
+    const timeAxisFunction = d3.svg.axis()
+    .scale(this.props.xScaleFunction())
+    .tickFormat(d => {
+      return moment.unix(d).format(this.props.timeAxisFormat);
+    })
+    .tickSize(this.props.tickSize, this.props.tickSize)
```
tumentumurchudur: Did you mean to include `this.props.tickSize` twice?

tumentumurchudur:
Overall, looks really good. :+1:

jmophoto:
A lot of good stuff in there. I like the idea of the d3 specific components. I think they could be reused for other chart types.

src/utils/Chart.js
cerinman: It is not.  The dividing by and then multiplying by a thousand is a way to give our selves a cushion for the min and max values.  Other wise the data points end up being to close to the top and bottom of the graph.  This ensures that the data ends up being more centralized in the graph.

src/utils/Chart.js
cerinman: It is not.  The dividing by and then multiplying by a thousand is a way to give our selves a cushion for the min and max values.  Other wise the data points end up being to close to the top and bottom of the graph.  This ensures that the data ends up being more centralized in the graph.

src/utils/Chart.js
cerinman: 6 is the number of ticks that we would like displayed.  I'll move this to a variable so it is discoverable.

src/components/TimeBasedLineChart.js
cerinman: I agree, I'm making multiple calls for the same value.  I'll make the change.

src/utils/Chart.js
tumentumurchudur: Not sure why d3 works that way. It just sounds really strange to divide a number by 1000 and then multiplying it by 1000 gives a different result. 

I know Math.ceil returns an integer. And you are trying to get a smaller number by diving it by 1000. But anytime you have a number less than 1, `Math.ceil` always returns 1. How does that change anything? Can you give me an example? Sorry, I'm just curious to know.

src/components/TimeBasedLineChart.js
cerinman: :+1:

src/components/d3/BreakPointGroup.js
cerinman: Yeah I'll pull the value out into a variable rather than making three calls

src/components/d3/CirclesGroup.js
cerinman: The idea here is that more than 45 circles for a small enough width would result in circles overlapping each other.  45 was a number arrived at after talking to product.  In cases where there are more than 45 circles we don't want to draw the circles at all.  I'll move 45 to a descriptive variable.

src/components/d3/TimeXAxisGroup.js
cerinman: Yes, the D3 tickSize function takes two arguments.  Inner and Outer tick sizes.  I just want them to be the same size in this case.

src/utils/Chart.js
cerinman: Math.ceil will round up the number to the nearest decimal place so 569 becomes 569.  In this case I want it to round up to 1000 so I have to do the extra math.

screen shot 2016-02-10 at 1 24 22 pm

cerinman:
@tumentumurchudur @jmophoto this is ready for another review.

src/utils/Chart.js
tumentumurchudur: @cerinman I thought about this some more. Let's say we have numbers like `35156.25` and `35352.25`, 
`Math.ceil(d[axis] / 1000) * 1000` returns `36000` for both of them. As a result, the chart will be much smoother. But I'm not sure if that is what you want. If you want the chart more accurate, I would do this to increase the value just a little in relative to its current value, so it wouldn't be too close to the top and bottom
```javascript
return d[axis]*1.001
```

cerinman: So I tried your suggestion but the chart then doesn't display in a way where the line is more centralized in the graph.  

Using `d[axis]*1.001`
screen shot 2016-02-10 at 1 32 29 pm


Using `Math.ceil(d[axis] / 1000) * 1000`
screen shot 2016-02-10 at 1 32 52 pm


I should note that I found the `Math.ceil(d[axis] / 1000) * 1000` equation from doing research on D3 chart scaling.  It seems to be a pretty normal approach to padding space on the chart to get your line to be more centralized.

jmophoto: If I read this right, it isn't changing the value of the data points, but setting the top and bottom values for the entire chart. Otherwise the chart would go to 569 and the top data point would be on the edge. 

The divide and multiply by x just rounds up to the nearest xth. The example rounds 569 to the nearest 1000th. If it was 1569 it would return 2000. 

If x was 100 it would round 569 to 600. if x was 10 it would round 569 to 570.

My question would be if we always want to round to 1000. What if the numbers range from 1-99?

cerinman: Valid point @jmophoto.  Let me play with the math some more and see if I can make the `1000` more dynamic.

src/utils/Chart.js
```diff
@@ -0,0 +1,42 @@
...
+    });
+
+    const maxDigits = Math.ceil(max).toString().length - 1;
+    const multiplier = Math.pow(10, maxDigits);
+
+    max = Math.ceil(max / multiplier) * multiplier;
+    min = Math.floor(min / multiplier) * multiplier;
+
+    return { min, max };
+  },
```
cerinman: @tumentumurchudur @jmophoto Ok I revamped this a little.  we now figure out the correct multiplier based upon the number of digits in the max value after it has been rounded to a whole number(int).  We then use the `Math.pow` function that gets the correct multiplier to use.  Tested with several data sets with small and large numbers and everything is working correctly.

cerinman:
Ready for another review

jmophoto:
All the items appear to be addressed. I like the multiplier flexibility. I give this a :+1: and hand it off to @tumentumurchudur for his approval.

src/utils/Chart.js
```diff
@@ -0,0 +1,42 @@
...
+
+    let min = d3.min(data, d => {
+      return d[axis];
+    });
+
+    const maxDigits = Math.ceil(max).toString().length - 1;
+    const multiplier = Math.pow(10, maxDigits);
+
+    max = Math.ceil(max / multiplier) * multiplier;
+    min = Math.floor(min / multiplier) * multiplier;
```
tumentumurchudur: Do you need a custom multiplier for the min too? Let's say the min value was 959 and max value was 36254. Then you would have 40000 and 10000. Wouldn't you want 959 be 1000?

cerinman: In the case of the min value, your example would actually end up being 0 because the min value gets floored.  But I can see who if the min number ended up being a negative number how that could result in a large negative min value.  I'll add a custom min multiplier.

cerinman:
@tumentumurchudur added the custom multiplier for the min value.  Ready to review again.

src/components/TimeBasedLineChart.js
```diff
@@ -1,32 +1,245 @@
...
+    fontFamily: StyleConstants.FontFamily,
+    position: 'relative',
+    boxSizing: 'content-box',
+    display: 'inline-block'
+  },
+
+  // Chart
+  breakPointLabel: {
+    fill: StyleConstants.Colors.ASH,
+    'font-family': StyleConstants.Fonts.REGULAR,
```
tumentumurchudur: I'm sure there is a reason for not using `fontFamily`. Does this have to do with d3 charts?

cerinman: Yes, I noted in the comment above the objects.  D3 cas to have them as `font-family` or things don't work.

src/components/TimeBasedLineChart.js
```diff
@@ -1,32 +1,245 @@
...
+            width={dateRectangleWidth}
+            x={hoveredDataPointXScaleValue - dateRectangleMiddle}
+            y={adjustedHeight}
+          />
+        </g>
+        <g className='hover-state-circle' ref='hoverStateCircle' transform={translation}>
+          <circle
+            className='circle'
+            cx={this.props.xScaleValueFunction(this.props.hoveredDataPoint.x)}
+            cy={this.props.yScaleValueFunction(this.props.hoveredDataPoint.y)}
```
tumentumurchudur: It should be `{yScaleValueFunction(hoveredDataPoint.y)}`

cerinman: yep missed these.

src/components/TimeBasedLineChart.js
```diff
@@ -42,713 +255,294 @@ class TimeBasedLineChart extends React.Component {
...
-  }
+  },
 
   render () {
+    const { breakPointDate, breakPointLabel, data, height, lineColor, margin, rangeType, width, zeroState, yAxisFormatter } = this.props;
+    const { adjustedHeight, adjustedWidth, hoveredDataPoint } = this.state;
+
     return (
       <div className='mx-time-based-line-chart' style={[styles.component, { height: this.props.height + 'px', width: this.props.width + 'px' }]}>
         {this.props.data.length ? (
```
tumentumurchudur: You can drop the `px` here.
`{ height: this.props.height, width: this.props.width}`

cerinman: yep

cerinman:
@tumentumurchudur ready for another review if you're not already looking.

src/components/TimeBasedLineChart.js
tumentumurchudur: :+1:

src/components/TimeBasedLineChart.js
```diff
@@ -42,713 +255,294 @@ class TimeBasedLineChart extends React.Component {
...
-  }
+  },
 
   render () {
+    const { breakPointDate, breakPointLabel, data, height, lineColor, margin, rangeType, width, zeroState, yAxisFormatter } = this.props;
+    const { adjustedHeight, adjustedWidth, hoveredDataPoint } = this.state;
+
     return (
-      <div className='mx-time-based-line-chart' style={[styles.component, { height: this.props.height + 'px', width: this.props.width + 'px' }]}>
+      <div className='mx-time-based-line-chart' style={[styles.component, { height: this.props.height, width: this.props.width }]}>
```
tumentumurchudur: One last change here
`{ height: height, width: width }`, no need for `this.props`

src/components/TimeBasedLineChart.js
```diff
@@ -42,713 +255,294 @@ class TimeBasedLineChart extends React.Component {
...
         {this.props.data.length ? (
```
tumentumurchudur: same here. No need for `this.props`

src/components/TimeBasedLineChart.js
cerinman: I'll do you one better.  I bet I can just do `{ height, width }`

src/components/TimeBasedLineChart.js
cerinman: yep

src/components/TimeBasedLineChart.js
```diff
@@ -42,713 +255,294 @@ class TimeBasedLineChart extends React.Component {
...
           <div>
-            {this._renderTooltip()}
-            {this._renderChart()}
+            <svg
+              height={height}
+              onMouseLeave={this._handleChartMouseLeave}
+              ref='chart'
+              width={width}
+            >
+              {this.props.shadeFutureOnGraph ? (
```
tumentumurchudur: might as well pull this prop at the top to be consistent?

src/components/TimeBasedLineChart.js
```diff
@@ -42,713 +255,294 @@ class TimeBasedLineChart extends React.Component {
...
+                yScaleValueFunction={this._getYScaleValue}
+              />
+              <CirclesGroup
+                adjustedHeight={adjustedHeight}
+                data={data}
+                translation={this._getLineTranslation()}
+                xScaleValueFunction={this._getXScaleValue}
+                yScaleValueFunction={this._getYScaleValue}
+              />
+              {this.state.hoveredDataPoint ? (
```
tumentumurchudur: same here

src/components/d3/GridLinesGroup.js
```diff
@@ -0,0 +1,51 @@
...
+      .tickSize(this.props.tickSize, 0, 0)
+      .tickFormat('')
+      .ticks(tickValues.length)
+      .tickValues(tickValues);
+
+    d3.select(this.refs[this.props.axis + 'GridLines']).call(gridLinesFunction);
+  },
+
+  render () {
+    return <g className={this.props.axis + '-grid-line'} ref={this.props.axis + 'GridLines'} transform={this.props.translation} />;
```
tumentumurchudur: may want to put props on a new line to make it easy on the eye

src/utils/Chart.js
```diff
@@ -0,0 +1,44 @@
...
+      return d[axis];
+    });
+
+    const maxDigits = Math.ceil(max).toString().length - 1;
+    const minDigits = Math.floor(min).toString().length - 1;
+    const maxMultiplier = Math.pow(10, maxDigits);
+    const minMultiplier = Math.pow(10, minDigits);
+
+    max = Math.ceil(max / maxMultiplier) * maxMultiplier;
+    min = Math.floor(min / minMultiplier) * minMultiplier;
```
tumentumurchudur: :+1:

cerinman:
Everything fixed again @tumentumurchudur .  Thank you for such a in depth PR review.  :+1:

src/components/d3/AxisGroup.js
```diff
@@ -0,0 +1,47 @@
...
+  },
+
+  componentDidUpdate () {
+    this._renderAxis();
+  },
+
+  _renderAxis () {
+    const tickValues = ChartUtils.getAxisTickValues(this.props.data, this .props.axis);
+    const axisFunction = d3.svg.axis()
+      .scale(this.props.scaleFunction())
```
tumentumurchudur: Did you want to hoist these `this.props` at the top? Or just the long ones? Asking because we want to be consistent.

src/components/d3/AxisGroup.js
```diff
@@ -0,0 +1,47 @@
...
+      .orient(this.props.orientation)
+      .tickFormat(this.props.axisFormatFunction)
+      .ticks(tickValues.length)
+      .tickValues(tickValues);
+
+    d3.select(this.refs[this.props.axis + 'Axis']).call(axisFunction);
+  },
+
+  render () {
+    return <g className={this.props.axis + '-axis'} ref={this.props.axis + 'Axis'} transform={this.props.translation} />;
```
tumentumurchudur: I would indent and put each prop on a new line.

src/components/d3/AxisGroup.js
cerinman: I didn't on these shorter files but I'm more than happy to do it if we want to remain consistent.

src/components/d3/CirclesGroup.js
```diff
@@ -0,0 +1,69 @@
...
+  componentDidMount () {
+    this._animateCircles();
+  },
+
+  componentDidUpdate () {
+    this._animateCircles();
+  },
+
+  _animateCircles () {
+    if (this.props.shouldAnimate) {
```
tumentumurchudur: Same here. Did you want to make it consistent with the `render` function? I'm fine either way where you have the `this.props` assigned to variables at the top.

src/components/d3/AxisGroup.js
cerinman: Lets leave these shorter components as is for now.

src/components/d3/CirclesGroup.js
cerinman: Lets leave these shorter components as is for now.

tumentumurchudur:
Looks good to me. :shipit:

cerinman:
:bow:

cerinman:
These are breaking changes so this will be released as 2.0.0-rc.0

PR #177, opened 2016-02-11
Fix renderOptions conditional in Select component

cerinman:
We recently mad a change to the select component to fix an issue where the drop shadow was be rendered for the options div even when there was no options.  We fixed that issue be adding a conditional check to see if this.props.options had a length before rendering the options.  

While that worked for most use cases, we allow users of Select to pass options via children and in those cases the conditional was preventing the options from rendering.  This fixes that issue by adding `this.props.children` to the check.

Before Select with children
screen shot 2016-02-11 at 11 42 55 am

After Select with children
screen shot 2016-02-11 at 11 43 09 am

@tumentumurchudur @jmophoto @paniclater @bsbeeks @lukeschunk

### DIFF
--- a/src/components/Select.js
+++ b/src/components/Select.js
@@ -188,7 +188,7 @@ const Select = React.createClass({
               type={this.state.isOpen ? 'caret-up' : 'caret-down'}
             />
           </div>
-          {this.props.options.length ? this._renderOptions() : null}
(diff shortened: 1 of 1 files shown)


jmophoto:
:shipit:

PR #179, opened 2016-02-16
Do not show break point label if to close to right edge of graph

cerinman:
We ran into a situation where the break point label was being extended past the edge of the graph on the right side if it's position was to close to the edge.  This PR fixes that issue by choosing to NOT draw the svg text label if it's position is less than 100 pixels from the right edge of the graph.

@mxenabled/frontend-engineers 

Break Point in center
screen shot 2016-02-16 at 2 43 24 pm

Break Point close to right edge
screen shot 2016-02-16 at 2 43 04 pm


bsbeeks:
:shipit:

PR #180, opened 2016-02-16
Time Based Line Chart - Fix y axis min max padding

cerinman:
This PR addresses an issue where attempting to round a min or max value for the y axis that was small was ending up with a rounded number not large or small enough.  This was causing the min or max tick value to be weird for the given scale/step interval.  This fixes the issue by always defaulting to at least a multiplier of 100.

@mxenabled/frontend-engineers 

Y Axis Before
screen shot 2016-02-16 at 3 22 24 pm

Y Axis After
screen shot 2016-02-16 at 3 22 50 pm


bsbeeks:
:shipit:

PR #182, opened 2016-02-17
DatePicker: Allows selectedDate prop

tumentumurchudur:
This PR allows users of the DatePicker component to be able to pass `selectedDate` prop and have it be selected in the calendar. The DatePicker component had an internal state that kept track of the selected date but it shouldn't do that, simply because the consumers wouldn't have control over the selected date.

### DIFF
--- a/src/components/DatePicker.js
+++ b/src/components/DatePicker.js
@@ -11,7 +11,6 @@ const DatePicker = React.createClass({
     calendarWrapperStyle: React.PropTypes.object,
     caretWrapperStyle: React.PropTypes.object,
     closeOnDateSelect: React.PropTypes.bool,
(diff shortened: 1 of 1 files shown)


bsbeeks:
I disagree, I think it's ok for the component to keep track of it's selected date. By passing that selected data back to the consumers via a method prop, the consumers can take action on the changed date. The default date is there to know what the initial state on mount should be, but after that the component manages it's own state.

Were we seeing bugs with the original approach?

bsbeeks:
Talked with @tumentumurchudur offline and I had misunderstood the approach. This is good to go :shipit:

tumentumurchudur:
Some things are not working quite right. I thought of a better way to approach this issue. I will follow up with a PR. Closing.

PR #183, opened 2016-02-18
DatePicker: Expose a method to Datepicker, so selected date can be changed

tumentumurchudur:
This PR allows users to change its internal `selectedDate` state through a method. We need to be able to change the selected date in the calendar because I have two instances of the DatePicker component that  need to communicate and change each other's selected date.

### DIFF
--- a/src/components/DatePicker.js
+++ b/src/components/DatePicker.js
@@ -54,6 +54,15 @@ const DatePicker = React.createClass({
     };
   },
 
+  componentWillReceiveProps (nextProps) {
+    if (nextProps.defaultDate !== this.props.defaultDate) {
+      this.setState({
+        selectedDate: nextProps.defaultDate,
+        inputValue: moment.unix(nextProps.defaultDate).format(this.props.format)
+      });
+    }
+  },
+
   _getInputValueByDate (date) {
     let inputValue = null;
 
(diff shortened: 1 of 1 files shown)


cerinman:
So I get what we're trying to accomplish here but I'm concerned about changing internal state via an external method call.  I did use this approach for validating a form but in that case we were just asking the form to validate itself, we weren't changing the form's state.  

I would feel better if others looked at this before shipping this just to make sure we're all comfortable with this pattern.  It seems like this could open the door to a pattern that is anti-react if we're not careful.  @tumentumurchudur use case here sounds valid to me

@mxenabled/frontend-engineers what is everyone's thoughts on this?

See https://facebook.github.io/react/tips/expose-component-functions.html for better context.  We would basically reaching into the DatePicker via a ref and calling this new function to set the selectedDate.

jmophoto:
Wouldn't you be able to do the same thing by passing a new defaultDate in as a prop and then using a componentWillReceiveProps to check?

```
    if (newProps.defaultDate !== this.props.defaultDate) {
        this.setState({
           selectedDate: newProps.defaultDate
       });
    }
}
```

tumentumurchudur:
@jmophoto let me play with that.

tumentumurchudur:
@jmophoto `componentWillReceiveProps` works great and is a better solution. :+1: 
Ready for review again.

jmophoto:
Looks good. Much more React-y-er.

:shipit:

cerinman:
Awesome, thanks for the help @jmophoto.  A much better solution.

bsbeeks:
A little late, but wanted to chime in that I think this was an excellent solution. Good work!

tumentumurchudur:
@bsbeeks so you like the other solution where changing the state through an exposed method?

bsbeeks:
No, I like this one that got merged =). Just giving props!

PR #186, opened 2016-02-25
Return dataSet object on hover or click instead of index

jmophoto:
This changes the onClick and onMouseEnter callbacks to return the active dataset object instead of an index. This helps with the accuracy for the user since the data could change and not match with the index in the chart.

@mxenabled/frontend-engineers


bsbeeks:
:shipit:

PR #188, opened 2016-02-26
Added ImageUploader component

tumentumurchudur:
This component allows users to upload an image by dragging and dropping the image file onto the drop zone. 

Props
- allowedFileTypes: React.PropTypes.array - (Allowed image types. Default value: ['png', 'jpg', 'gif'])
- color: React.PropTypes.string - (Icon color when file is being dragged in browser. Default value: PRIMARY) 
- fileBeingDragged: React.PropTypes.bool - (Indicates that a file being dragged)
- maxFileSize: React.PropTypes.number - (Max file size, default value: 150 KB)
- onFileError: React.PropTypes.func - (Called when there is an error loading a file)
- onFileRemove: React.PropTypes.func - (Called when a file is removed)
- onFileSuccess: React.PropTypes.func.isRequired - (Called when a file is successfully loaded and it also returns the file loaded both in encoded base64 string and the File format)

screen shot 2016-02-26 at 4 37 22 pm

### DIFF
--- a/src/components/ImageUploader.js
+++ b/src/components/ImageUploader.js
@@ -1,0 +1,257 @@
+const React = require('react');
+const Radium = require('radium');
+const numeral = require('numeral');
+
+const StyleConstants = require('../constants/Style');
+
+const Icon = require('./Icon');
+const Button = require('./Button');
+
+const ImageUploader = React.createClass({
+  propTypes: {
+    allowedFileTypes: React.PropTypes.array,
+    brandColor: React.PropTypes.string,
+    fileBeingDragged: React.PropTypes.bool,
+    maxFileSize: React.PropTypes.number,
+    onFileError: React.PropTypes.func,
+    onFileRemove: React.PropTypes.func,
+    onFileSuccess: React.PropTypes.func.isRequired
+  },
+
+  getDefaultProps () {
+    return {
+      allowedFileTypes: ['png', 'jpg', 'gif'],
+      brandColor: StyleConstants.Colors.PRIMARY,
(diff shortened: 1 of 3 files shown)


bsbeeks:
@tumentumurchudur I know you were running into some issues getting this component to send the image to one of our internal app's API. Were you able to figure those out? I'd assume if we were having problems, others would run into the same thing.

src/components/ImageUploader.js
```diff
@@ -0,0 +1,257 @@
...
+  getInitialState () {
+    return {
+      fileContent: null,
+      invalidMessage: null,
+      uploadedFile: null
+    };
+  },
+
+  _onDropzoneClick () {
+    this.refs.hiddenInput.click();
```
bsbeeks: Per React docs, using a ref string is "mostly legacy at this point" (https://facebook.github.io/react/docs/more-about-refs.html#the-ref-string-attribute). Could we adjust this to use the ref callback approach?

src/components/ImageUploader.js
```diff
@@ -0,0 +1,257 @@
...
+          Max file size - <span>{this.props.maxFileSize}</span> KB
+        </div>
+      </div>
+    );
+  },
+
+  render () {
+    return (
+      <div>
+        {this._renderDropzone()}
```
bsbeeks: Pulling this into a separate method seems a bit overkill. I suggest eliminating `_renderDropzone` and moving it's content into the `render` method.

bsbeeks:
Closing in favor of @jmophoto's PR.

PR #190, opened 2016-03-02
Rename npm test script to eslint

cerinman:
This was done to avoid confusion where we use the script name `test` for travis config stuff in other applications.


paniclater:
:shipit:

PR #191, opened 2016-03-02
updated -8 to -30 for background color hover

lukeschunk:
Derek came over and we discussed how the button component hover effect (in the demo app ) is not noticeable enough. We decided that changing the value passed in from `-8` to `-30` would be a good change.

The only thing I'm worried about with this PR is it having an adverse effect on all the apps? Perhaps some of them (due to color difference, etc) have a hover effect that is actually dialed in nicely and this will push it overboard?

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -51,8 +51,8 @@ const Button = React.createClass({
         transition: 'all .2s ease-in',
 
         ':hover': {
-          backgroundColor: StyleConstants.adjustColor(this.props.primaryColor, -8),
-          borderColor: StyleConstants.adjustColor(this.props.primaryColor, -8),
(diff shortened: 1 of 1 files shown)


bsbeeks:
We're good to make this change wholesale in all our MX projects

lukeschunk:
@bsbeeks are you saying we won't merge this PR, but rather go through each App and make the change?

bsbeeks:
No, go ahead and merge it. I'm saying we are good with roling out the change to all the apps after its merged.

lukeschunk:
Great - sounds good :+1: 

I don't believe I have authority to merge it - could you add my personal account to the org?

bsbeeks:
:shipit: and added to the repo.

PR #192, opened 2016-03-03
Add DropZone component

jmophoto:
Adds drop zone component that will accept a file, either by dragging or through a file browser, and return that file.

Props:
- onFileAdd: function returning the file dropped or selected
- onFileRemove: function called on remove click
- uploadedFile: file that is passed back to the component on drop/select

This incorporates @tumentumurchudur's hover effect, changing the border to dashed blue and a white background when dragging onto the window.

Validation and file processing is left up to the parent component. 

DropZone name could use some suggestions. ImageUpload doesn't seem to work because any file can be dropped. FileUpload might work, but the component doesn't actually upload. Thoughts @mxenabled/frontend-engineers?

Initial State:
screen shot 2016-03-02 at 10 05 50 pm

Drag State:
screen shot 2016-03-02 at 10 06 35 pm

Loaded State:
screen shot 2016-03-02 at 10 06 50 pm

### DIFF
--- a/src/components/DropZone.js
+++ b/src/components/DropZone.js
@@ -1,0 +1,168 @@
+const React = require('react');
+const Radium = require('Radium');
+const numeral = require('numeral');
+
+const StyleConstants = require('../constants/Style');
+
+const Button = require('./Button');
+
+const DropZone = React.createClass({
+  propTypes: {
+    dragging: React.PropTypes.bool,
+    onFileAdd: React.PropTypes.func,
+    onFileRemove: React.PropTypes.func,
+    uploadedFile: React.PropTypes.any,
+    validationMessage: React.PropTypes.string
+  },
+
+  getInitialState () {
+    return {
+      dragging: false,
+      imageSource: null
+    };
+  },
+
+  componentDidMount () {
+    window.addEventListener('mouseover', this._handleMouseOver);
+    window.addEventListener('mouseout', this._handleMouseOut);
+  },
+
+  componentWillUnmount () {
+    window.removeEventListener('mouseover', this._handleMouseOver);
+    window.removeEventListener('mouseout', this._handleMouseOut);
+  },
+
+  _handleMouseOver (e) {
+    if (e.buttons === 1) {
+      this.setState({
+        dragging: true
+      });
+    }
+  },
+
+  _handleMouseOut () {
+    this.setState({
+      dragging: false
+    });
+  },
+
+  _handleFileSelect (e) {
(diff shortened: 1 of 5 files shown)


bsbeeks:
For the name, I vote `FileUpload`.

src/components/DropZone.js
```diff
@@ -0,0 +1,165 @@
...
+
+    reader.readAsDataURL(file);
+
+    reader.onload = function () {
+      this.setState({
+        dragging: false,
+        imageSource: reader.result
+      });
+      this.props.onFileAdd(file);
+    }.bind(this);
```
bsbeeks: instead of binding, you can use a fat arrow method.

src/components/DropZone.js
```diff
@@ -0,0 +1,165 @@
...
+          ref='hiddenInput'
+          style={styles.hiddenInput}
+          type='file'
+        />
+      </div>
+    );
+  }
+});
+
+const styles = {
```
bsbeeks: let's move this into a method called `styles` to follow our new internal pattern.

bsbeeks:
Overall, this is a solid. I think @tumentumurchudur 's component has some elements that yours is missing (https://github.com/mxenabled/mx-react-components/pull/188). Specifically I think the validation (file type, size, etc) should be included in the component. I suggest combining the PRs into one.

jmophoto:
This has been cleaned up and combined with items from @tumentumurchudur's PR.

- Added validation for type and size (optional)
- Added validation message on error
- Validation message is removed on new drag
- Drag indicator was changed from window to the drop zone
- On drag
-- Expand height of the drop zone
-- Change background to white
-- Add dashed blue border
-- Add import icon
-- Change text
- Add conditional to show preview if image and document icon for others
- Renamed to FileUpload

@mxenabled/frontend-engineers

src/components/FileUpload.js
```diff
@@ -0,0 +1,220 @@
...
+const React = require('react');
+
+const StyleConstants = require('../constants/Style');
+
+const Button = require('./Button');
+const Icon = require('./Icon');
+
+const FileUpload = React.createClass({
+  propTypes: {
+    allowedfileTypes: React.PropTypes.array,
```
bsbeeks: should be `allowedFileTypes`...the "f" needs to be capitalized.

jmophoto: :+1: Fixed and pushed.

bsbeeks:
:shipit:

PR #197, opened 2016-03-07
Lowercase radium

jmophoto:
- Lowercase radium require to fix duplicate module error.

@mxenabled/frontend-engineers


cerinman:
:shipit:

PR #199, opened 2016-03-07
Fix Internet Explorer 11/Edge issues with TimeBasedLineChart

cerinman:
Lines were not rendering correctly or at all in TimeBasedLineChart in IE 11/Edge

- Updates eslint-config-mx to version 1.0.4 and resolves new peer dependency on `babel-plugin-transform-object-assign`
- `Math.log10` is not supported in IE.  Dropped it's use in favor of the suggested equation provided by Mozilla in the docs for `Math.log10`.
- Only draw the zero dashed line if zero is contained within the y range.
- Use `none` in favor of `0, 0` for `strokeDasharray` property when asking for a dashed line via the `dashLine` prop in `LineGroup` component.  `0, 0` is not supported in IE and results in no line being rendered.

@mxenabled/frontend-engineers 

IE 11
screen shot 2016-03-07 at 2 32 18 pm

Edge
screen shot 2016-03-07 at 2 31 57 pm


src/components/TimeBasedLineChart.js
```diff
@@ -553,7 +559,7 @@ const TimeBasedLineChart = React.createClass({
                   xScaleValueFunction={this._getXScaleValue}
                 />
               ) : null}
-              {alwaysShowZeroLine ? (
+              {alwaysShowZeroLine && this._yRangeContainsZero() ? (
```
paniclater: alwaysShowZeroLine implies to me if it is true the zero line should always be displayed, but there is a second check here in the logic -- can we rename alwaysShowZeroLine to something else, maybe just `showZeroLine`. I don't know, maybe there isn't a better option. But it does feel a little confusing here.

cerinman: Agreed, that is a little confusing.  I'll make the change.

cerinman:
@paniclater prop renamed.  Ready for another review.

paniclater:
:shipit:

src/components/TimeBasedLineChart.js
```diff
@@ -227,6 +226,7 @@ const TimeBasedLineChart = React.createClass({
       shadeBelowZero: false,
       shadeFutureOnGraph: true,
       showBreakPoint: true,
+      showZeroLine: false,
```
bsbeeks: This is a breaking change. Since we aren't providing any backwards compatibility, we need to make sure we bump the second number in the version.

cerinman: Yes you are correct. I already released but I can bump the version and release again.

Issue #200, opened 2016-03-08
Default Value/Validation in Select Component doesn't match.


lukeschunk:
I noticed that the PropTypes in the Select component states: `selected: React.PropTypes.object,` validating for an object. Yet in the default, it's setting the value to a boolean: 
```
getDefaultProps () {
    return {
      onChange () {},
      options: [],
      placeholderText: 'Select One',
      selected: false,
      valid: true
    };
  },
```

In the demo app, it is getting passed an object for this value... so perhaps the default value should not be a boolean?

bsbeeks:
Good catch. We should just remove the default value all together.

scotttball:
Close? @lukeschunk

PR #201, opened 2016-03-08
removed default prop for selected on select component

lukeschunk:
The `selected` prop type is an Object, but the default value was a boolean. Removing the default value here so there is no validation conflict.

### DIFF
--- a/src/components/Select.js
+++ b/src/components/Select.js
@@ -34,8 +34,7 @@ const Select = React.createClass({
   getInitialState () {
     return {
       highlightedValue: null,
-      isOpen: false,
-      selected: false
+      isOpen: false
     };
   },
 
--- a/src/components/Select.js
+++ b/src/components/Select.js
@@ -27,7 +27,6 @@ const Select = React.createClass({
       onChange () {},
       options: [],
       placeholderText: 'Select One',
-      selected: false,
       valid: true
     };
   },


scotttball:
:shipit:

paniclater:
It would be better to set the default value to match the propType or change the propType to match the default value. If the prop is sometimes an object and sometimes a boolean then the code should be fixed to make sure the type stays the same.

bsbeeks:
@paniclater not in this case. The code is already handling an empty or null `selected` prop, so we can omit the default prop type.

paniclater:
@bsbeeks @lukeschunk I really think we should at least change the value in getInitialState as well then to either be an empty object or null. Doesn't make sense to me to remove false in one place and not the other.
```
  getInitialState () {
    return {
      highlightedValue: null,
      isOpen: false,
      selected: false
    };
  },
```

bsbeeks:
@paniclater I agree on the state attribute. I thought you were referring to setting a default on the prop.

paniclater:
:+1: Yeah my comment was a little rambly. Glad I cleared it up.

scotttball:
`selected: null` would be correct then? @paniclater

paniclater:
I personally prefer `selected: {}` because it is more descriptive of what the expected data will be like. I did a quick search and found examples of both empty objects and null. Not sure if we have a preferred pattern for this.

I just don't want selected to sometimes be a boolean and sometimes be an object.

paniclater:
After reviewing the code `null` is the best option here because of how the value gets determined in the render:
```
const selected = this.state.selected || this.props.selected || { displayValue: this.props.placeholderText, value: '' };
```

lukeschunk:
so just to clarify @paniclater @bsbeeks @scotttball I'm going to simply remove `selected` from `getInitialState`?

paniclater:
yeah:
```
  getInitialState () {
    return {
      highlightedValue: null,
      isOpen: false
    };
  },
```
Then when the const selected is used:
```
const selected = this.state.selected || this.props.selected || { displayValue: this.props.placeholderText, value: '' };
```
`this.state.selected` will be falsey

paniclater:
:shipit:

bsbeeks:
Fixes #200

PR #202, opened 2016-03-08
Add isRequired to onFileAdd and onFileRemove

jmophoto:
- Add isRequired to onFileAdd and onFileRemove
- Add check for allowFileTypes before validation
- Add textAlign center to component style

@mxenabled/frontend-engineers


cerinman:
Code looks good.  There is now a Components folder in the root of the project that has a doc file for the FileUpload component.  Was that added by accident when the docs where updated on the gh-pages branch?  Can we kill that folder and file as part of this PR?  If not can we kill it as part of another PR?  I keep clicking into it by mistake.

jmophoto:
I saw that but hoped it was just local. I removed it and re-pushed.

cerinman:
:shipit:

PR #204, opened 2016-03-09
Add ability to send existing imageUrl to FileUpload

jmophoto:
- Add imageUrl to FileUpload props
- If imageUrl is present, show preview of existing file on load
- Replace preview of existing file with uploaded file on drop
- Restore preview of existing file if uploadedFile is removed
- Fade preview image on drag
- Keep upload instructions in view even with image preview
- Add attention icon to invalid message
- Move invalid message rendering into a method since it is reused

@mxenabled/frontend-engineers 

Empty drop zone:
screen shot 2016-03-09 at 1 46 25 pm

Loaded drop zone
screen shot 2016-03-09 at 1 21 58 pm

Existing image drop zone
screen shot 2016-03-09 at 1 22 12 pm

Faded preview on fade
screen shot 2016-03-09 at 1 23 10 pm

### DIFF
--- a/src/components/FileUpload.js
+++ b/src/components/FileUpload.js
@@ -10,6 +10,7 @@ const Icon = require('./Icon');
 const FileUpload = React.createClass({
   propTypes: {
     allowedFileTypes: React.PropTypes.array,
+    imageUrl: React.PropTypes.string,
     maxFileSize: React.PropTypes.number,
     onFileAdd: React.PropTypes.func.isRequired,
     onFileRemove: React.PropTypes.func.isRequired,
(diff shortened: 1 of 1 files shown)


src/components/FileUpload.js
```diff
@@ -87,44 +86,82 @@ const FileUpload = React.createClass({
...
+    } else {
+      return null;
+    }
+  },
+
+  _validateFile (file) {
+    const isTooBig = this.props.maxFileSize < file.size / 1000;
+    const isInvalidType = this.props.allowedFileTypes && this.props.allowedFileTypes.indexOf(file.type) < 0;
+
+    if (isTooBig || isInvalidType) {
```
paniclater: I'm finding the nested ternary inside of the if a little complex. What if you use a let statement to set the message in each case and then setState after with whatever the value is?

paniclater: I misunderstood what was going on here and sorted it out offline. :shipit:

PR #205, opened 2016-03-10
PieChart: Pie chart with animations

tumentumurchudur:
This PR adds a PieChart/DonutChart component that has customizable animations. 

Props
- animateDuration: specifies the duration of animation on load and hover events
- animationTypeOnLoad: specifies the type of animation on load. It is either roll or pop. More animation types can be added in the future
- arcWidth: the width of the donut chart if inner and outer radius are defined.
- colors: defines the color of each slice in the pie chart
- data: raw data 
- height: height of the chart
- hoverExpandDistance: specifies how far each slice grows outward on hover
- onLabelClick: function called when the label is clicked
- onMouseEnter: function called when mouse enters the slice in the pie chart
- onMouseLeave: function called when mouse leaves the slice
- onPieClick: function called when a slice in the piechart is clicked
- showDataLabels: determines whether to show the label on the center of the chart
- style: styles passed to the chart from a parent component
- width: specifies the width of the chart

screen shot 2016-03-10 at 11 20 53 am

@mxenabled/frontend-engineers

### DIFF
--- a/src/components/PieChart.js
+++ b/src/components/PieChart.js
@@ -14,6 +14,7 @@ const PieChart = React.createClass({
     onMouseEnter: React.PropTypes.func,
     onMouseLeave: React.PropTypes.func,
     showDataLabel: React.PropTypes.bool,
+    style: React.PropTypes.object,
     title: React.PropTypes.string,
     width: React.PropTypes.number
   },
@@ -27,15 +28,27 @@ const PieChart = React.createClass({
       onMouseEnter () {},
       onMouseLeave () {},
       showDataLabel: true,
-      title: '',
       width: 300
     };
   },
 
   componentDidMount () {
-    const dom = ReactDOM.findDOMNode(this);
+    const dom = ReactDOM.findDOMNode(this.refs.chart);
 
     this._renderChart(dom);
+  },
+
+  _renderShowLabels () {
+    const styles = this.styles();
+
+    return (
+      <div
+        className='mx-donutchart-data'
+        style={styles.center}
+      >
+        {this.props.children}
+      </div>
+    );
   },
 
   _renderChart (dom) {
@@ -44,6 +57,13 @@ const PieChart = React.createClass({
     const expandWidth = 10;
     const strokeWidth = 3;
     const radius = Math.min(width, height) / 2 - expandWidth;
+    const arc = d3.svg.arc()
+      .outerRadius(radius)
+      .innerRadius(radius / 2);
+
+    const arcOver = d3.svg.arc()
+      .outerRadius(radius + expandWidth)
+      .innerRadius(radius / 2);
 
     const pieChart = d3.select(dom)
       .append('svg')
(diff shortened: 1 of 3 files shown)


src/components/PieChart.js
```diff
@@ -0,0 +1,182 @@
...
+      showDataLabel: true,
+      width: 300
+    };
+  },
+
+  componentDidMount () {
+    const dom = ReactDOM.findDOMNode(this.refs.chart);
+
+    this._renderChart(dom);
+  },
```
cerinman: I'm only seeing `this._renderChart` called in `componentDidMount`.  What happens if the chart data changes?  Will the chart update and reanimate?

tumentumurchudur: @cerinman You are correct. The chart won't update because it is already mounted. I'm not sure if there is any situation we would change the chart data on the fly. In the Investment widget, we don't change the chart data once it is mounted. Do you have any recommendations on how we could update the chart as its data changes and still keep the animations?

cerinman: It will need to update when we use it in the Spending widget react rewrite as a user can dig into transactions and change transaction categories.  Something like this should work

```javascript
componentWillReceiveProps (nextProps) {
  if (!_isEgual(this.props.data, nextProps.data) {
     this._renderChart();
  }
}
```

tumentumurchudur: Had an offline discussion with @cerinman and we both agreed that the suggestion above wouldn't work since the chart is already mounted in the DOM. Calling `this._renderChart` would probably create another instance of the PieChart.

src/components/PieChart.js
```diff
@@ -0,0 +1,182 @@
...
+        style={styles.center}
+      >
+        {this.props.children}
+      </div>
+    );
+  },
+
+  _renderChart (dom) {
+    const { animateDuration, arcWidth, height, width, hoverExpandDistance, data, colors } = this.props;
+    const thisGlobal = this;
```
bsbeeks: why are we doing this instead of using fat arrow methods?

bsbeeks:
I'm worried that we are taking a step back with this PR. We've established a good pattern of using d3 for calculations and React to render out elements to the DOM. These changes hand the DOM manipulation back to D3. @cerinman @jmophoto you've two have been working a lot with d3 lately. I'd like to hear your thoughts on this approach. And more importantly, if we can accomplish these same animations without giving d3 control of the DOM.

src/components/PieChart.js
tumentumurchudur: Yeah, that is a good question. I tried doing that but the whole issue around passing global `this` and local `this` to different functions make this component very hard to read and confusing. It is more of an issue with D3 and how it handles animations. I thought it would be best to leave it as, so it is easer to read it from top to bottom.
@cerinman @jmophoto Any thoughts?

cerinman:
@bsbeeks I had an offline conversation with @tumentumurchudur and we're going to work together tomorrow to see if we can figure out a solution because I have the same concerns.  We'll see what we can do.

bsbeeks:
:+1:

src/components/PieChart.js
tumentumurchudur: @bsbeeks Fat arrow methods are not suitable in some situations where "lexical scoping" points `this` to the wrong context. And there is no way to override `this` in arrow methods to point it at the right context. For example, if I use the fat arrow function here 
```javascript
.on('mouseenter', function (d, i) {
        d3.select(this)
          .transition()
          .duration(animateDuration)
          .attr('d', arcOver);

        thisGlobal.props.onMouseEnter(d.value, i);
      })
```

```javascript
.on('mouseenter', (d, i) => {
        d3.select(this)
          .transition()
          .duration(animateDuration)
          .attr('d', arcOver);

        thisGlobal.props.onMouseEnter(d.value, i);
      })
```
In this case, `this` is pointing at the global scope instead of the object where the `mouseenter` event is being defined.

bsbeeks: Make sense :+1:

PR #206, opened 2016-03-10
Moore/simple drawer component

paniclater:
Behold The Humble Beginnings of \<Drawer /\>
--
screen shot 2016-03-10 at 1 40 05 pm

--
This adds a new animated `Drawer` component to be used to replacing stacking modals.
Example
--
```
<Drawer 
  isOpen={this.state.showDrawer}
  onClose={this._toggleDrawer}
  siblingContent={navContent}>
    <div>
      Insert Custom Content Here
    </div>
</Drawer>
```
Props
--
-`duration` _Number_ **optional**
Number of miliseconds for opening animation to take. Defaults to 500 if not provided.

-`easing` _Array_ **optional**
Bezier line curve array for easing. Defaults to [0.28, 0.14, 0.34, 1.04] if not provided.
See [here](http://cubic-bezier.com/#.17,.67,.83,.67) for more information on crafting this prop.

- `isOpen` _Boolean_ **optional**
Simple flag to be used to tell the drawer whether or not it should animate or remain open or closed

- `onClose` _Function_ **optional** 
Function to be called when drawer is closed

- `navContent` _Object_ **Optional**
Object passed in to render sibling content navigation and handlers. If `siblingContent` is passed in, it must adhere to the following `shape` propType requirement. The next sibling navigation will not be displayed unless this object is passed in with all required types.
  - `duration` _Number_ **Optional**
 Optional duration for nav animations, if not provided, top level duration will be used.

  - `label` _String_ **Required**
  Text and other elements to be displayed in navContent

  - `onNextClick` _Function_ **Required**
  Function to be called when the next sibling caret is clicked

  - `onPreviousClick` _Function_ **Required**
  Function to be called when the previous sibling caret is clicked

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -523,10 +523,10 @@ const Demo = React.createClass({
     });
   },
 
-  _nextSibling () {
+  _handleNextSiblingClick () {
   },
 
-  _previousSibling () {
+  _handlePreviousSiblingClick () {
   },
 
   _toggleDrawer () {
@@ -538,9 +538,9 @@ const Demo = React.createClass({
   render () {
     const navConfig = {
       duration: 200,
-      onNextClick: this._nextSibling,
+      onNextClick: this._handleNextSiblingClick,
       label: _find(this.state.drawerSiblings, { selected: true }).id + ' of ' + this.state.drawerSiblings.length,
-      onPreviousClick: this._previousSibling
+      onPreviousClick: this._handlePreviousSiblingClick
     };
 
     return (
@@ -549,7 +549,6 @@ const Demo = React.createClass({
           <Button onClick={this._toggleDrawer} style={{ position: 'absolute', left: 15, top: 15 }}>Toggle Drawer</Button>
           {this.state.showDrawer ?
           <div style={{ textAlign: 'center', width: '80%', margin: 'auto', position: 'relative', height: '200', overflow: 'hidden' }}>
-            <p>Drawer</p>
             <Drawer navConfig={navConfig} onClose={this._toggleDrawer}>
               <div style={{ padding: 20, fontFamily: 'Helvetica, Arial, sans-serif' }}>Insert Custom Content Here</div>
             </Drawer>
--- a/demo/app.js
+++ b/demo/app.js
@@ -450,7 +450,7 @@ const Demo = React.createClass({
       },
       lineChartData: [],
       selectedDatePickerDate: moment().unix(),
-      showDrawer: false,
+      showDrawer: true,
       showModal: false,
       showSmallModal: false,
       uploadedFile: null,
@@ -524,11 +524,9 @@ const Demo = React.createClass({
   },
 
   _nextSibling () {
-    console.log('next');
   },
 
   _previousSibling () {
-    console.log('previous');
   },
 
   _toggleDrawer () {
@@ -548,12 +546,14 @@ const Demo = React.createClass({
     return (
       <div>
         <br/><br/>
-        <div style={{ textAlign: 'center', width: '80%', margin: 'auto', position: 'relative', height: '200', overflow: 'hidden' }}>
           <Button onClick={this._toggleDrawer} style={{ position: 'absolute', left: 15, top: 15 }}>Toggle Drawer</Button>
-          <Drawer isOpen={this.state.showDrawer} navContent={navContent} onClose={this._toggleDrawer}>
-            <div style={{ padding: 20, fontFamily: 'Helvetica, Arial, sans-serif' }}>Insert Custom Content Here</div>
-          </Drawer>
-        </div>
+          {this.state.showDrawer ?
+          <div style={{ textAlign: 'center', width: '80%', margin: 'auto', position: 'relative', height: '200', overflow: 'hidden' }}>
+            <p>Drawer</p>
+            <Drawer navContent={navContent} onClose={this._toggleDrawer}>
+              <div style={{ padding: 20, fontFamily: 'Helvetica, Arial, sans-serif' }}>Insert Custom Content Here</div>
+            </Drawer>
+          </div> : null}
         <br/><br/>
         <div style={{ textAlign: 'center', width: '80%', margin: 'auto' }}>
           <FileUpload
--- a/demo/app.js
+++ b/demo/app.js
@@ -550,7 +550,9 @@ const Demo = React.createClass({
         <br/><br/>
         <div style={{ textAlign: 'center', width: '80%', margin: 'auto', position: 'relative', height: '200', overflow: 'hidden' }}>
           <Button onClick={this._toggleDrawer} style={{ position: 'absolute', left: 15, top: 15 }}>Toggle Drawer</Button>
-          <Drawer isOpen={this.state.showDrawer} onClose={this._toggleDrawer} siblingContent={drawerSiblingContent}/>
+          <Drawer isOpen={this.state.showDrawer} onClose={this._toggleDrawer} siblingContent={drawerSiblingContent}>
+            <div style={{ padding: 20, fontFamily: 'Helvetica, Arial, sans-serif'}}>Insert Custom Content Here</div>
+          </Drawer>
         </div>
         <br/><br/>
         <div style={{ textAlign: 'center', width: '80%', margin: 'auto' }}>
--- a/demo/app.js
+++ b/demo/app.js
@@ -536,7 +536,7 @@ const Demo = React.createClass({
   },
 
   render () {
-    const navContent = {
+    const navConfig = {
       duration: 200,
       onNextClick: this._nextSibling,
       label: _find(this.state.drawerSiblings, { selected: true }).id + ' of ' + this.state.drawerSiblings.length,
@@ -550,7 +550,7 @@ const Demo = React.createClass({
           {this.state.showDrawer ?
           <div style={{ textAlign: 'center', width: '80%', margin: 'auto', position: 'relative', height: '200', overflow: 'hidden' }}>
             <p>Drawer</p>
-            <Drawer navContent={navContent} onClose={this._toggleDrawer}>
+            <Drawer navConfig={navConfig} onClose={this._toggleDrawer}>
               <div style={{ padding: 20, fontFamily: 'Helvetica, Arial, sans-serif' }}>Insert Custom Content Here</div>
             </Drawer>
           </div> : null}
--- a/demo/app.js
+++ b/demo/app.js
@@ -538,11 +538,10 @@ const Demo = React.createClass({
   },
 
   render () {
-    const drawerSiblingContent = {
-      currentSiblingIndex: _find(this.state.drawerSiblings, { selected: true }).id,
-      nextSibling: this._nextSibling,
-      previousSibling: this._previousSibling,
-      totalSiblings: this.state.drawerSiblings.length
+    const navContent = {
+      onNextClick: this._nextSibling,
+      label: _find(this.state.drawerSiblings, { selected: true }).id + ' of ' + this.state.drawerSiblings.length,
+      onPreviousClick: this._previousSibling
     };
 
     return (
@@ -550,7 +549,7 @@ const Demo = React.createClass({
         <br/><br/>
         <div style={{ textAlign: 'center', width: '80%', margin: 'auto', position: 'relative', height: '200', overflow: 'hidden' }}>
           <Button onClick={this._toggleDrawer} style={{ position: 'absolute', left: 15, top: 15 }}>Toggle Drawer</Button>
-          <Drawer isOpen={this.state.showDrawer} onClose={this._toggleDrawer} siblingContent={drawerSiblingContent}>
+          <Drawer isOpen={this.state.showDrawer} navContent={navContent} onClose={this._toggleDrawer}>
             <div style={{ padding: 20, fontFamily: 'Helvetica, Arial, sans-serif' }}>Insert Custom Content Here</div>
           </Drawer>
         </div>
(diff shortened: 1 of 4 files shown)


src/components/Drawer.js
```diff
@@ -0,0 +1,146 @@
...
+
+const Icon = require('../components/Icon');
+
+const Drawer = React.createClass({
+  propTypes: {
+    duration: React.PropTypes.number,
+    easing: React.PropTypes.array,
+    isOpen: React.PropTypes.bool,
+    onClose: React.PropTypes.func.isRequired,
+    siblingContent: React.PropTypes.shape({
```
lukeschunk: never seen the `shape` prop type before

src/components/Drawer.js
```diff
@@ -0,0 +1,146 @@
...
+
+const StyleConstants = require('../constants/Style');
+
+const Icon = require('../components/Icon');
+
+const Drawer = React.createClass({
+  propTypes: {
+    duration: React.PropTypes.number,
+    easing: React.PropTypes.array,
+    isOpen: React.PropTypes.bool,
```
cerinman: We've decided to drop the `isOpen` prop on modals and modal like components in favor of letting the parent decide when to render.  Using a prop like this prevents the component from unmounting when `isOpen` is false.

cerinman: Just noticing that your're not using this prop to decide if the component should be shown but for animation.  So my above comment is pointless

src/components/Drawer.js
```diff
@@ -0,0 +1,146 @@
...
+  _slideArrowAndSiblingContent (isOpen) {
+    this._slideArrow(isOpen);
+    this._slideSiblingContent(isOpen);
+  },
+
+  _slideArrow (isOpen) {
+    const el = this._arrow;
+    const transition = isOpen ? { left: -25 } : { left: 25 };
+    const options = {
+      duration: 200,
```
cerinman: Is there a reason why we use a hard coded value for duration here but use `this.props.duration` elsewhere?

demo/app.js
```diff
@@ -502,10 +523,38 @@ const Demo = React.createClass({
     });
   },
 
+  _nextSibling () {
+    console.log('next');
+  },
+
+  _previousSibling () {
+    console.log('previous');
```
cerinman: console logs left over in the demo here.  can we kill them, or are they serving a purpose?

src/components/Drawer.js
```diff
@@ -0,0 +1,146 @@
...
+    if (newProps.isOpen !== this.props.isOpen) {
+      this._renderTransition(newProps.isOpen);
+    }
+  },
+
+  _renderSiblingContent () {
+    const styles = this.styles();
+
+    return (
+      <div ref={(ref) => (this._siblingContent = ref)} style={styles.siblingContent}>
```
paniclater: @cerinman https://facebook.github.io/react/docs/more-about-refs.html here's the react docs on that:

https://facebook.github.io/react/docs/more-about-refs.html

demo/app.js
paniclater: I just wanted an example callback for the demo without spending a lot of time implementing complicated behavior. I am making these props required in the drawer component to prevent the nav content rendering useless buttons.

src/components/Drawer.js
paniclater: Forgot to fix this.

paniclater: fixed now.

bsbeeks:
The border below the nav area seems a bit bold. Does it match the designs provided by Product?

src/components/Drawer.js
bsbeeks: @cerinman I actually agree with your first statement. I think this component should not control it's open/close state. Instead the component should animate in on mount and animate out on unmount. The parent component should control when it's added to the UI.

@paniclater thoughts?

src/components/Drawer.js
```diff
@@ -0,0 +1,146 @@
...
+    this._renderTransition(this.props.isOpen);
+  },
+
+  componentWillReceiveProps (newProps) {
+    if (newProps.isOpen !== this.props.isOpen) {
+      this._renderTransition(newProps.isOpen);
+    }
+  },
+
+  _renderNavContent () {
```
bsbeeks: I'd either eliminate this separate render method and do everything in the JSX, or pull the conditional used to render this content out of the JSX and into this method.

src/components/Drawer.js
```diff
@@ -0,0 +1,146 @@
...
+    const options = {
+      complete: this._doNavAnimation.bind(this, isOpen),
+      duration: this.props.duration,
+      easing: this.props.easing
+    };
+
+    Velocity(el, transition, options);
+  },
+
+  _doNavAnimation (isOpen) {
```
bsbeeks: The naming around the "nav" of this component is a bit confusing:
- What you are currently calling `nav` seems like it should be something like `topBar` or `header`...since it contains the exit/back icon and the forward/back nav. 
- the prop `navContent` implies that it's value would be something you'd render and not config/settings for the nav. I'd suggest renaming it to `navConfig`.
- A better name for this method would then be `_animatedTopBarContent`...or something similar.
- Better names for `_slideNavContent` and `_slideArrow` would be `_animateNav`, `_animateArrow`

src/components/Drawer.js
```diff
@@ -0,0 +1,146 @@
...
+      </div>
+    );
+  },
+
+  styles () {
+    return {
+      component: {
+        top: 0,
+        bottom: 0,
+        right: -800,
```
bsbeeks: What happens if the screen width is less than 800 pixels? We should make this width more dynamic by using percentages with maybe a max-width.

src/components/Drawer.js
```diff
@@ -0,0 +1,146 @@
...
+        overflow: 'hidden',
+        backgroundColor: StyleConstants.Colors.PORCELAIN
+      },
+      icon: {
+        color: StyleConstants.Colors.ASH
+      },
+      iconContainer: {
+        position: 'absolute',
+        left: -25,
+        top: 12
```
bsbeeks: magic numbers. If we are trying to center the icon, can use use `translate` instead?

src/components/Drawer.js
```diff
@@ -0,0 +1,146 @@
...
+      },
+      nav: {
+        backgroundColor: StyleConstants.Colors.PORCELAIN,
+        borderBottom: 'solid 1px ' + StyleConstants.Colors.ASH,
+        height: 15,
```
bsbeeks: why do we need to set the height?

src/components/Drawer.js
```diff
@@ -0,0 +1,146 @@
...
+    Velocity(el, transition, options);
+  },
+
+  render () {
+    const styles = this.styles();
+
+    return (
+      <div ref={(ref) => (this._component = ref)} style={styles.component}>
+        <nav style={styles.nav}>
+          <span ref={(ref) => (this._arrow = ref)} style={styles.iconContainer}>
```
bsbeeks: Is this span needed? Seems like we can put these props directly onto `Icon`

src/components/Drawer.js
```diff
@@ -0,0 +1,146 @@
...
+        right: -800,
+        position: 'absolute',
+        width: 800,
+        overflow: 'hidden',
+        backgroundColor: StyleConstants.Colors.PORCELAIN
+      },
+      icon: {
+        color: StyleConstants.Colors.ASH
+      },
+      iconContainer: {
```
bsbeeks: for consistency, I'd name these `arrow` and `arrowContainer`

src/components/Drawer.js
```diff
@@ -0,0 +1,146 @@
...
+
+    Velocity(el, transition, options);
+  },
+
+  render () {
+    const styles = this.styles();
+
+    return (
+      <div ref={(ref) => (this._component = ref)} style={styles.component}>
+        <nav style={styles.nav}>
```
bsbeeks: based on the naming suggestions above, this would become `styles.topBar` or `styles.header`

demo/app.js
```diff
@@ -502,10 +523,38 @@ const Demo = React.createClass({
...
+  _toggleDrawer () {
+    this.setState({
+      showDrawer: !this.state.showDrawer
+    });
+  },
+
   render () {
+    const navContent = {
+      duration: 200,
+      onNextClick: this._nextSibling,
```
tumentumurchudur: I would rename `_nextSibling` to something like `_clickNextSibling` to indicate that it is a function.

demo/app.js
```diff
@@ -502,10 +523,38 @@ const Demo = React.createClass({
...
+      label: _find(this.state.drawerSiblings, { selected: true }).id + ' of ' + this.state.drawerSiblings.length,
+      onPreviousClick: this._previousSibling
```
tumentumurchudur: Same here. `_clickPrevSibling`

bsbeeks: following our naming conventions it should be prefixed with `_handle`. `_handlePreviousSiblingClick`

src/components/Drawer.js
```diff
@@ -0,0 +1,146 @@
...
+      <div ref={(ref) => (this._navContent = ref)} style={styles.navContent}>
+        <Icon onClick={this.props.navContent.onPreviousClick} size={25} style={styles.icon} type='caret-left'/>
+        {this.props.navContent.label}
+        <Icon onClick={this.props.navContent.onNextClick} size={25} style={styles.icon} type='caret-right'/>
+      </div>
+    );
+  },
+
+  _renderTransition (isOpen) {
+    const el = this._component;
```
tumentumurchudur: Can you put `_component` on the state? Looks like it is stored on the component which seems anti-React.

src/components/Drawer.js
paniclater: @bsbeeks @cerinman the component is not controlling its open, close state -- it receives an isOpen flag as a prop on mount. To make this more clear, I'm removing the default prop of isOpen and setting isOpen to isRequired.

src/components/Drawer.js
paniclater: @tumentumurchudur I'm going to go a different direction with this -- but just to follow up this pattern is pulled directly from the react docs.
```
render: function() {
    return <TextInput ref={(c) => this._input = c} />;
  },
```
https://facebook.github.io/react/docs/more-about-refs.html

tumentumurchudur: @paniclater Looking at your use cases for this, they are all attached to either `div` or `span`. Basically, you could do 
```javascript
<div ref='_navContent' style={styles.navContent}>
```
instead of 
```javascript
<div ref={(ref) => (this._navContent = ref)} style={styles.navContent}>
```
because this is a simple `div` that has no methods or anything you can call using the ref.

Quoted from https://facebook.github.io/react/docs/more-about-refs.html
When attaching a ref to a DOM component like `<div />`, you get the DOM node back; when attaching a ref to a composite component like `<TextInput />`, you'll get the React class instance. In the latter case, you can call methods on that component if any are exposed in its class definition.

paniclater: I know I didn't quite catch you on the way out but I do want to bring to your attention also from the same page that using a string as a ref is possible but considered a legacy approach at this point:
>React also supports using a string (instead of a callback) as a ref prop on any component, although this approach is mostly legacy at this point.

src/components/Drawer.js
paniclater: @bsbeeks, @cerinman I've got all the transitions working succesfully on mount, but the unmount is posing a problem I've been working through unsuccesfully. I'm going to do some research on this and try to come up with a good solution. Right now the React Animations add on is looking the most hopeful. 

https://facebook.github.io/react/docs/animation.html

paniclater:
@bsbeeks, @cerinman, @tumentumurchudur:

I believe I've dealt with all the initial feedback on this PR.

Here's a breakdown of changes:

 - Rendering handled by parent component
 - Border below nav color corrected to match design
 - conditional statement for nav moved into render function
 - naming changes:
   - animation functions named to include `animate`
   - handler functions named to include `handle` + action
   - header component name changed
   - styles updated for simplicity
 - width changed to dynamic with percentages and animation values updated to work with dynamic values

I will provide inline comments on changes that I didn't make with some explanations.

src/components/Drawer.js
```diff
@@ -0,0 +1,149 @@
...
+    Velocity(el, transition, options);
+  },
+
+  render () {
+    const styles = this.styles();
+
+    return (
+      <div ref={(ref) => (this._component = ref)} style={styles.component}>
+        <header style={styles.header}>
+          <span ref={(ref) => (this._backArrow = ref)} style={styles.backArrow}>
```
paniclater: @tumentumurchudur: I didn't change the ref attributes to be strings because @bsbeeks said that string refs are no longer the recommended path to dealing with this problem & the react docs said the string refs work but are pretty much legacy at this point.

paniclater: @bsbeeks I didn't remove the span here because targeting the `Icon` element directly with a ref returned a Radium element that conflicted with Velocity's animation. I could have used the Radium element to get the dom node, but the span seemed a cleaner solution.

demo/app.js
```diff
@@ -502,10 +523,37 @@ const Demo = React.createClass({
     });
   },
 
+  _handleNextSiblingClick () {
+  },
```
paniclater: @cerinman I removed the console logs so we don't have eslint errors, but kept these in because they are required handlers in order to display the nav section with animation that I wanted to demonstrate.

cerinman:
My concerns have been addressed.  :+1: from me.  I'll not ship it until everyone else that commented has their say.

bsbeeks:
Looks good to me :shipit:

PR #207, opened 2016-03-11
Luke/select debuggin - fixes #203

### LINKED ISSUE #203: Select Component - Multiple options appear selected/in a Hover State.
In the Select component dropdown, if you select one option, close the dropdown, then go to select another option, sometimes when the dropdown reopens, two items will have the hover state. This is better understood by viewing the screenshots below:

screen shot 2016-03-08 at 1 59 24 pm
screen shot 2016-03-08 at 1 59 38 pm

Is this worth looking into?

lukeschunk:
This PR improves the User Experience of the `Select` open source component. 

### Before: 
select_old

### Description of Changes: 
The biggest change this PR makes is the `hover` style effect change is now handled by a `mouseOver` event. Before, the Hover backgroundColor was persisting after the component was closed, until a user hovered over it again, in which case it disappeared. This gave the illusion that multiple options were selected at a time. 

### After:
select_new


lukeschunk:
The PR that was just merged in @bsbeeks effected mine a little but I think everything is good - you can still pass down a color in props to change the active item background color. if you want it to be the same color no matter what on hover and active, I can fix that as well.

scotttball:
I think that the previous convention was to have active and hover the same color. I think that having them two different colors would be confusing for the user. @lukeschunk @bsbeeks

lukeschunk:
@scotttball @bsbeeks Updated the select to have a different style when selected. Still fixes the hover issue where multiple items were having the `:hover` state persist after the select was close. 

screen shot 2016-03-14 at 10 50 20 am

Also couldn't figure out a good way to better vertically align it better other than magic numbers - thoughts?

bsbeeks:
:shipit:

PR #208, opened 2016-03-11
Added ability to pass color prop to Select

r-walsh:
Effects `styles.option[':hover'].backgroundColor` and `styles.activeItem.backgroundColor`. Both default to `StyleConstants.Colors.PRIMARY` if no color is passed.

### DIFF
--- a/src/components/Select.js
+++ b/src/components/Select.js
@@ -45,6 +45,7 @@ const Select = React.createClass({
 
   componentDidMount () {
     styles.option[':hover'].backgroundColor = this.props.color || StyleConstants.Colors.PRIMARY;
+    styles.activeItem.backgroundColor = this.props.color || StyleConstants.Colors.PRIMARY;
   },
 
   _handleScrimClick () {
@@ -251,7 +252,6 @@ const styles = {
     position: 'relative'
   },
   activeItem: {
-    backgroundColor: StyleConstants.Colors.PRIMARY,
     color: StyleConstants.Colors.WHITE
   },
   invalid: {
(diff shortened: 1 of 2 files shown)


src/components/Select.js
```diff
@@ -38,6 +41,11 @@ const Select = React.createClass({
     };
   },
 
+  componentDidMount () {
+    styles.option[':hover'].backgroundColor = this.props.color || StyleConstants.Colors.PRIMARY;
+    styles.activeItem.backgroundColor = this.props.color || StyleConstants.Colors.PRIMARY;
```
bsbeeks: rather than setting the fallback color here, you can set the default prop to `StyleConstants.Colors.PRIMARY`.

bsbeeks: Actually, rather than do that. We've implemented a new pattern that hasn't made it's way to our open source components. Instead of defining our styles in a `const`, we now define them in a `styles` method. We then reference that method when needing styles. This gives us easier access to props and state.

```javascript
getDefaultProps () {
  return {
    color: StyleConstants.Colors.PRIMARY
  },
},

render () {
  const styles = this.styles();

  // other render code
},

styles () {
  return {
    // other styles
    activeItem: {
      backgroundColor: this.props.color
    },
    option: {
      // other styles
      ':hover': {
        backgroundColor: this.props.color
      }
    }
  };
}

**Comment by bsbeeks** [comment]
Actually, rather than do that. We've implemented a new pattern that hasn't made it's way to our open source components. Instead of defining our styles in a `const`, we now define them in a `styles` method. We then reference that method when needing styles. This gives us easier access to props and state.

```javascript
getDefaultProps () {
  return {
    color: StyleConstants.Colors.PRIMARY
  },
},

render () {
  const styles = this.styles();

  // other render code
},

styles () {
  return {
    // other styles
    activeItem: {
      backgroundColor: this.props.color
    },
    option: {
      // other styles
      ':hover': {
        backgroundColor: this.props.color
      }
    }
  };
}
```

r-walsh:
@bsbeeks refactored

bsbeeks:
:shipit:

PR #209, opened 2016-03-11
Adds Loading Button

### LINKED ISSUE #176: Add "Loading Button" Component
#### Overview
Another possible addition to standard button components. This button provides instant feedback to the user that clicking the button actually registered, even though the process may not be instant. Is could work as any button type (Primary, Secondary, Neutral, Base).

screen shot 2016-02-01 at 11 15 11 am

Some examples:
*Ladda*
http://lab.hakim.se/ladda/

*Mailchimp* http://ux.mailchimp.com/patterns/forms (about halfway down)
Screencast: https://jumpshare.com/v/YNcqsjgFRKp74BAKlKuu

*Marvel*
Screencast: https://jumpshare.com/v/NVNoHRej2wgkuYjJ9zjs

iheartrachie:
This updates the `Button` component to add a spinner and should close #176. It works on every button type but disabled.

## Example
```
<Button isActive={this.state.spinnerIsActive} onClick={this._handleSpinnerClick} type="neutral">Spinner</Button>
<Button actionText='Spinning...' isActive={this.state.spinnerWithTextIsActive} onClick={this._handleSpinnerWithTextClick}>Spinner with text</Button>
```
1iiyxefkkl

## Props
`isActive` _boolean_ **optional**
Default: `false`
A boolean that is toggled when the button is clicked and when the item is finished loading.
```
_handleSpinnerClick () {
  this.setState({
    spinnerIsActive: !this.state.spinnerIsActive
  });
},

<Button isActive={this.state.spinnerIsActive} onClick={this._handleSpinnerClick}>Spinner</Button>
```
<br>

`actionText` _string_ **optional**
The button text when `isActive` is `true`. If not defined, a spinner without text is shown.
`<Button isActive={this.state.spinnerIsActive} onClick={this._handleSpinnerClick}>Spinner</Button>`

## Challenges
* The spinner is based off the `Loader` component and uses `Spin` for the rotation. The rotation of the circle is slightly lopsided, which isn't too noticeable on the `Loader` but is when condensed down and in the context of a button. I've tried to adjust the pixels of the circle and `Spin` but wasn't able to resolve the issue. I'd like some feedback here as to how I can make this smoother.

## To Do
* Use Velocity to add a transition on the button when it changes width


src/components/Button.js
```diff
@@ -21,11 +25,40 @@ const Button = React.createClass({
...
+    };
+
+    if (this.props.isActive) {
+      return (
+        <div>
+          <Spin>
+            <div style={circleStyles} />
+          </Spin>
+          {this.props.actionText}
+        </div>
```
bsbeeks: This wrapping div isn't needed.

src/components/Button.js
```diff
@@ -137,7 +170,7 @@ const Button = React.createClass({
     return (
       <div {...this.props} style={[styles.component, styles[this.props.type], this.props.style]}>
         {this.props.icon ? <Icon size={20} style={styles.icon} type={this.props.icon} /> : null}
```
bsbeeks: Does this look/function ok if there is an icon?

src/components/Button.js
```diff
@@ -21,11 +25,40 @@ const Button = React.createClass({
...
       primaryColor: StyleConstants.Colors.PRIMARY,
       type: 'primary'
     };
   },
 
+  _renderButtonContent () {
+    const circleStyles = {
+      display: 'inline-block',
+      verticalAlign: 'middle',
+      borderRadius: '100%',
```
bsbeeks: try setting the border radius to the matching pixel height/width. It might fix the lopsidedness.

src/components/Button.js
iheartrachie: I can't get it to work with it. I'm guessing because of 54?

src/components/Button.js
iheartrachie: No. What should it do in that case? Hide the icon?

src/components/Button.js
iheartrachie: I'm not sure what you mean. The height & width are set to the same amount. On the `Loader` if you make the transparent boarders a color, you can see the lopsided spin. It's tough to capture via gif but I can show you on my computer.
i2qxhlf3dv

src/components/Button.js
bsbeeks: Hiding the icon seems most appropriate. Which brings up a thought...what if instead of using a circle with borders for the spin icon, we just use the `sync` icon from the Icon component? @derek-boman what say ye?

src/components/Button.js
bsbeeks: `borderRadius: this.props.actionText ? 9 : 11`

iheartrachie: Ah I see. Unfortunately that didn't make a difference.

bsbeeks: ...then let's defer to my other comment below about using the Icon component.

src/components/Button.js
bsbeeks: yep, my bad! I didn't see that.

derek-boman:
@bsbeeks I'm fine using the sync icon here.

bsbeeks:
@iheartrachie let's move forward with using the sync icon.

iheartrachie:
@bsbeeks, sorry about the delay on this. I was sick all last week. I just pushed up a commit that swaps out the circle boarder for the sync icon and adds support for icon buttons. It looks much better!
y86whnrahz

I'm not sure how I'd do Velocity animations on this since there isn't a start/end width property on the button. From what I understand it's a 'nice to have' bit anyway but if you have any insight as to how to accomplish this, I'll give it another shot.

derek-boman:
^ See the new issue above #227. Sync icon is okay for first cut but I think we should have a more appropriate icon for this and other "Loading" states.

loading_button

bsbeeks:
@iheartrachie I just added that new spinner icon, so if you rebase with master you'll have access to it.

iheartrachie:
The icon has been updated. The recorded gif is choppy but I promise it runs [super smooth](http://i.giphy.com/rhf5VlMiDQ5EI.gif)!
1i2guc4qgc

scotttball:
@iheartrachie Did we want it clockwise or counterclockwise?

bsbeeks:
we want it to spin clockwise.

iheartrachie:
Some stuff got mixed up when I rebased so I'm working with Beeks to redo this PR.

PR #210, opened 2016-03-13
Add Input Component

jimibue:
Adds a new `Input` component to be used as a input element in a form

# Example

          <Input
            label='Money:'
            onChange={ (e) => this._handleInputChange('input1', e) }
            placeholder='20'
            prefix='$'
            suffix='.00'
            type='number'
            value={this.state.input1}
            />
         
           <Input
            label='name:'
            onChange={ (e) => this._handleInputChange('input2', e) }
            type='text'
            value={this.state.input2}
            />

# Props
`label` *String* **Optional** 
The label for the input field, positioned on top of input field

`onChange` *Function* **Required** 
Method that is called when user provides input.  The name of the key in the input component parent is passed to this function so it can update state on that input.  This allows for multiple inputs to be added 
to parent component and each inputs value to be changed independently. 

` onChange={ (e) => this._handleInputChange('input1', e) }`

 Handler in parent component.. 
```
// In parent component (app.js)
_handleInputChange (str, e) {
    this.setState({
      [str]: e.target.value
    });
  }
```
`placeholder` *String* **Optional** 
Option for providing a input field with a placeholder 

`prefix` *String* **Optional** 
Option for providing text to go in front of input and is not
part of the input.

`suffix` *String* **Optional** 
Option for providing text to go at the end of input and is not
part of the input.

`type` *String* **Optional** 
Option for providing the type of field the input is i.e text, password, email, etc.
If no type is given input defaults to type text

`value` *String* **Required** 
The value that is in the input box.  This is setup so that the inputs containing parent state
keeps track of the value of the input
`value={this.state.input1}`

Default value of input can be set in the parents getInitialState method
```
  getInitialState () {
    return {
      input1: 'Default Value',
    };
  },
```

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -429,6 +429,7 @@ const Demo = React.createClass({
         value: 'accounts',
         displayValue: 'Accounts'
       },
+      input4: 'Default Value',
       lineChartData: [],
       selectedDatePickerDate: moment().unix(),
       showModal: false,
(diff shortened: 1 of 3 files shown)


bsbeeks:

Issue #150, opened 2016-01-27
Apprentice Project: Create a new `Input` component


bsbeeks:
# Purpose

The main purpose of this project is to gauge your technical and problem solving abilities for acceptance into MX's apprenticeship program. We are looking for clean, discoverable code. We will be evaluating your solution using the following criteria:
- Does the component address all of the desired functionality
- Is the code organized and easy to follow
- Did you follow the outlined process
- How (if at all) did you go above and beyond...this isn't required, but something we'll look at

Please feel free to comment on this Issue with any question or clarification.
# Process
- Fork the `mx-react-components` repo
- Build the component (Refer to CONTRIBUTING.md)
- Create a pull request against the `mx-react-components` repo master branch
- Once the PR is open we'll review it and get back to you
# Design Requirements

Input designs usually vary from project to project. Because of this the component should have a bare but configurable design. That being said, if you'd like to provide an option on the component to use a predefined design you are more than welcome to.
# Desired Functionality

The component should include **at least** the following criteria:
- Validation based on field type
- Prefix and suffix options (Example: the ability to add a static currency symbol)
- Hover and focus states
- Label option
- Responsive width. The component should fill the width of it's containing element.
- `onChange` handler. A way for parent elements to know when the value changes
- Placeholder option
- Default value option
### Other Requirements
- Name it `Input`
- Use `createClass` instead of ES6 classes
- Exposed it for use in`Index.js` 
- Add an example to `demo/app.js`
- Utilize ES6 syntax and features where appropriate
- Utilize `constants/Style` where appropriate
- Provide a detailed explanation of the component in your pull request. See https://github.com/mxenabled/mx-react-components/pull/17 for an example.

payhota:
Do you want the suffix to be fixed to the end of the input? Or would you rather it move toward the back as you type?

bsbeeks:
@payhota your call :smile:. I can see pros and cons to both approaches.

PR #211, opened 2016-03-14
Apprentice Project: Input component

payhota:
#150 

## Example:
```jsx
<Input
    id='intInput'
    label='Integer Input (with prefix and postfix)'
    onChange={this._handleInputChange}
    pattern='^[-]?\d+$'
    placeholderText='I will validate on blur. You gave me a regular expression to use.'
    postfix='.00'
    prefix='$'
    required={true}
    type='text'
    validate={true}
    validateMsg='Please enter a valid integer'
    validateOn='Blur'
/>
```

## Props:

`id` _String_ **recommended**
Default: _empty_

This becomes both the name and the id of the input field, which helps you identify it in a form and allows the field to be selected when the label is clicked.

`label` _String_ **recommended**
Default: _empty_

This is the label for the input field. You can click it to select the field if an `id` is given. It will turn red when the field is in an invalid state

`onChange` _Function_ **optional**

Handler for the parent element to know when the value changes. The keyboard event will be given to the function as an argument

`pattern` _String_ **optional**
Default: _empty_

If you require validation, you can enter a regular expression in this parameter that the component will use to validate the input. Be sure to make the validation parameter `true`. If the type is _email_ you do not have to add a pattern as there is already one built in to validate email addresses. 
**Note:** Do not add the leading or ending '/' as the string is converted to a regular expression

`placeholderText` _String_ **optional**
Default: _empty_

Text that will become the placeholder for the input field.

`postfix` _String_ **optional**
Default: _empty_

Adds a Bootstrap style postfix to the input. Useful for adding a currency character to the beginning. 

`prefix` _String_ **optional**
Default: _empty_

Adds a Bootstrap style prefix to the input. Useful for adding decimal places or an '@' symbol to the end.

`required` _Bool_ **optional**
Default: `false`

Makes the field required. The input will enter an invalid state when empty, only after it has been touched one time.

`type` _String_ **required**
Default: _text_

Corresponds to the type of input field you want to create. Any standard HTML input types should work.

`validate` _Bool_ **optional**
Default: `false`

Use this parameter to specify if you want to validate the input. If you add a pattern, this field is required.

`validateMsg` _String_ **optional**
Default: _'Invalid'_

If you want a specific message to be displayed when the field does not match your pattern, enter that message in this parameter.

`validateOn` _String_ **optional**
Default: _'Input'_

This is when the input box will be validated, currently the two options are 'Input' or 'Blur'. It can be annoying when typing an email for example if the field keeps changing color while you type, so 'Blur' may be more appropriate for certain situations.

`value` _String_ **optional**
Default: _empty_

Adds a default value to the input field.

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -16,7 +16,8 @@ const {
   SelectFullScreen,
   TimeBasedLineChart,
   ToggleSwitch,
-  TypeAhead
+  TypeAhead,
+  Input
 } = require('../src/Index');
 
 const styles = {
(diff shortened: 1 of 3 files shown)


bsbeeks:

PR #212, opened 2016-03-14
briancbarrow Input PR

briancbarrow:
I have made similar comments in my fork, but wanted to state them here for you. I am not 100% sure that this is what you were looking for, but this project pushed me which is a very good thing. I hadn't learned anything about React until this project. I was going to start learning it in the next couple of weeks but this gave me the perfect excuse to try it out. 

I just want to say thanks for the opportunity and I would love to hear any and all feedback you can provide.


bsbeeks:

PR #213, opened 2016-03-14
Input Component: Brendan Dayley

BrendanDayley:
This is my first experience with react, I gotta say it was a blast. I created a generic input field that could be manipulated with custom prefix suffix and input type based on the props the example I have uses it as an example for currency.

# example:

```
<Input 
    prefix="$" 
    suffix=".00" 
    placeholder="money" 
    defaultvalue="100" 
    label="My Cool"
    validate="number" 
/>
```
This input uses the **props:** prefix, suffix, placeholder, defaultvalue, label, and validate.

## prefix
is used by the component to define what comes before the input field

## suffix
defines what comes after

##placeholder
is the descriptive text in the input field

## defaultvalue
is the default value of the input field

## label
places a string before the input field

##validate
while the validation has room to grow for the program it allows preset values to be input to help determine what kind of input field needs to be validated as (in my examples case we have number because it is a currency based input, any non numeric character will throw the error).

the validate prop is what makes this input field cool I haven't finished adding all the different types of validation available but there is enough there to give you guys the Idea of what it could be.


bsbeeks:

PR #214, opened 2016-03-14
Experimental Test / Discussion PR around Animation and Unmounting

paniclater:
Test PR to Discuss A Pattern Component Animation
--
Given that we want:
 - Parent components to control mounting and unmounting of children components
 - Children components to control animation

I ran into a problem where animations would not be allowed to finish because the component would be removed from the DOM before the animation could finish.

This is my current preferred solution to this problem. I wanted to create a test / experimental PR before implementing it in the more complicated actual drawer component.

Pattern Details
--
 - Child component is rendered based upon a parent state variable or expression
 - When child component renders, its animation is executed in componentDidMount()
 - Child component is passed a `onClose` function prop that sets the state of the parent, causing the child component to be unmounted
 - The above onClose function is called on the `success` option of the Velocity animation.

Mount Flow:

1. Parent state is set to true 
2. Component Renders
3. Animation Executed on componentDidMount

Unmount Flow:

1. Close Click
2. Animation Executes
3. Function prop executed on animation success
4. Parent state set to false
5. Component unmounts.

@mxenabled/frontend-engineers Let me know what you think.

Note: Please don't judge UI concerns / naming. I was just trying to get this out the door for a discussion around the animation / unmounting pattern.

### DIFF
--- a/src/components/Drawer.js
+++ b/src/components/Drawer.js
@@ -1,0 +1,62 @@
+const React = require('react');
+const Velocity = require('velocity-animate');
+
+const Button = require('../components/Button');
+
+const Drawer = React.createClass({
+  propTypes: {
+    onClose: React.PropTypes.func.isRequired
+  },
+
+  componentDidMount () {
+    const el = this._container;
+    const transition = { right: 0 };
+    const options = {
+      duration: 500
+    };
+
+    Velocity(el, transition, options);
+  },
+
+  componentWillUnmount () {
+    console.log('component unmounted');
+  },
+
+  _handleCloseClick () {
+    const el = this._container;
+    const transition = { right: -800 };
+    const options = {
+      duration: 500,
+      complete: this.props.onClose
+    };
+
+    Velocity(el, transition, options);
+  },
+
+  render () {
+    const styles = this.styles();
+
+    return (
+    <div ref={(ref) => (this._container = ref)}style={styles.container}>
+      <Button onClick={this._handleCloseClick}>Close Drawer</Button>
+      <h1>Hello</h1>
+    </div>
+    );
+  },
+
+  styles () {
+    return {
+      container: {
+        backgroundColor: '#eaeaea',
+        color: '#88888',
+        width: 800,
+        height: 200,
+        position: 'absolute',
+        top: 0,
+        right: -800
+      }
+    };
+  }
+});
+
+module.exports = Drawer;
--- a/demo/app.js
+++ b/demo/app.js
@@ -6,6 +6,7 @@ const {
   Button,
   DatePicker,
   DatePickerFullScreen,
+  Drawer,
   DonutChart,
   FileUpload,
   Icon,
@@ -430,6 +431,7 @@ const Demo = React.createClass({
       },
       lineChartData: [],
       selectedDatePickerDate: moment().unix(),
+      showDrawer: false,
       showModal: false,
       showSmallModal: false,
       uploadedFile: null,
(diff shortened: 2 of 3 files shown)


cerinman:
To make sure I understand this, let me state it as I see it and you can tell me if I'm on the right track.

- Problem - Animation doesn't finish before component is removed on unmount.  In this scenario you are relying on the `componentWillUnmount` life cycle method.

- You're suggestion is to call the passed call back `onClose` on the success callback of the velocity animation that updates the prop used to determine if the drawer component is rendered in the parent component.

If I'm understanding all of that correctly then I have no issue in moving forward in this direction and I can't think of any drawbacks to doing it this way.  I'll reserve final judgement until after everyone else comments.

paniclater:
@cerinman really close. The callback onClose gets called, which updates the state of the parent variable. This way there is no prop on the drawer component that has any control over whether it is rendered or not.

cerinman:
@paniclater Yeah that is what I meant.  when I said, `updates the prop used to determine if the drawer component is rendered`.  I should have typed state variable rather than prop. :+1: from me.  I think the pattern is good.  Lets see what everyone else thinks.

paniclater:
@cerinman :+1: the semantics get tricky around this issue. I was pretty sure I understood what you were saying just wanted to clarify.

src/components/Drawer.js
```diff
@@ -0,0 +1,62 @@
...
+  componentWillUnmount () {
+    console.log('component unmounted');
+  },
+
+  _handleCloseClick () {
+    const el = this._container;
+    const transition = { right: -800 };
+    const options = {
+      duration: 500,
+      complete: this.props.onClose
```
paniclater: the `complete` option takes a function which is executed when the animation succesfully completes.

tumentumurchudur: This looks fine where the component allows the animation to finish and then calls  `this.props.onClose` causing the parent component to close the Drawer component. :+1:

paniclater: thanks @tumentumurchudur, I especially like this because it allows the parent component to add whatever it needs to handle to the onClose call, subsequent animations to surrounding components, event handler cleanup or whatever it needs to do without polluting the child component.

src/components/Drawer.js
```diff
@@ -0,0 +1,62 @@
...
+  },
+
+  componentDidMount () {
+    const el = this._container;
+    const transition = { right: 0 };
+    const options = {
+      duration: 500
+    };
+
+    Velocity(el, transition, options);
```
tumentumurchudur: This looks very similar to how I did the PieChart component where the chart animates on load. React is not aware of this at all. There is no way to animate `onload` transitions after the component is mounted.

paniclater: Yeah. I had issues with that as well. I messed around with an internal state variable in the child component but didn't come up with any solution that I was pleased with.

src/components/Drawer.js
tumentumurchudur: I wouldn't include clean-up code in `this.props.onClose` from the parent component because it knows nothing about the inner-workings of the child component. `componentWillUnmount` should handle cleanups.

paniclater: I meant clean up code that would be on the parent level -- something like an overall layout handler. The parent component will absolutely

bsbeeks:
I like this approach, it seems like a straight forward and scalable approach.

paniclater:
@mxenabled/frontend-engineers I'm going to incorporate this into the drawer component. Thanks for the feedback!

PR #215, opened 2016-03-14
Updates Select to compare objects  when determining active style. 

lukeschunk:
When trying to update the OS components in Gunny, I realized that if the list of options are objects, we would need to do some sort of object comparison. I believe a deep comparison is better.

### DIFF
--- a/src/components/Select.js
+++ b/src/components/Select.js
@@ -1,4 +1,4 @@
-const _ = require('lodash');
+const _isEqual = require('lodash/isEqual');
 const React = require('react');
 const ReactDOM = require('react-dom');
 const Radium = require('radium');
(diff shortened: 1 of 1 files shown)


paniclater:
:shipit:

src/components/Select.js
```diff
@@ -1,3 +1,4 @@
+const _ = require('lodash');
```
cerinman: Rather than pulling in the entire library for one method you can pull just `isEqual` by doing `const _isEqual = require('lodash/isEqual')` if we're on the latest version of lodash and `const _isEqual = require('lodash/lang/isEqual')` if we're not.

paniclater: Good catch @cerinman I completely missed that.

lukeschunk: @paniclater @cerinman I don't see either of those in the lodash node module folder - I actually looked for this yesterday because I noticed we did something similar in a different component. I just see the shallow comparison...

screen shot 2016-03-15 at 9 50 09 am

bsbeeks: did you try including it the way they suggested just to see if it would work?

cerinman: I chatted with him.  It's working.

lukeschunk:
@cerinman @paniclater This should be good now.

paniclater:
re-:shipit:

PR #220, opened 2016-03-16
added _isEqual to Icon conditional

lukeschunk:
Added lodash _isEqual method to the Icon conditional so the icon displays even if options in Select component are objects.


cerinman:
:shipit:

PR #221, opened 2016-03-16
DonutChart component animations

cerinman:
This PR adds D3 transition animations to the DonutChart component.

New Props added to DonutChart

- `animationDuration`: type(number) default(500) - used to determine the length of the load and data change animations of the chart in milliseconds.
- `animationTypeOnLoad`: type(string) default('roll') - determines the type of animation used for the load and data change events of the chart.  Can be one of `roll` or `pop`.
- `id`: type(string) default('donut-chart') - Used in cases where multiple charts are to be rendered at the same time.  Setting the id ensures that each chart's animations won't highjack the others.

@mxenabled/frontend-engineers


cerinman:
@mxenabled/engineers Ok I think this is ready to review.

demo/app.js
```diff
@@ -463,7 +465,21 @@ const Demo = React.createClass({
...
+          },
+          {
+            name: 'Data Point 2',
+            value: 80
+          },
+          {
+            name: 'Data Point 3',
+            value: 200
+          }
+        ]
```
paniclater: I know the setTImeout was already in here, but I'm assuming it is just mimicking an async api response for the demo?

cerinman: Yep, needed to testing delayed data to make sure the animations still work.

src/components/DonutChart.js
```diff
@@ -9,6 +10,8 @@ const DonutChart = React.createClass({
     activeIndex: React.PropTypes.number,
     activeOffset: React.PropTypes.number,
     animateOnHover: React.PropTypes.bool,
+    animationDuration: React.PropTypes.number,
```
paniclater: I think we should standardize the naming of these for all our animation components. I was just using `duration` in the drawer but I like the animation prefix better. I'll incorporate that into my next Drawer pr.

jmophoto:
This looks really good to me. Gotta love Maths!

:shipit:

tumentumurchudur:
@cerinman Can you set it up in the app where you click a slice on the outer chart and then it pops up the inner chart with animations? I also don't see the outer chart with the same type of animations as in the Spending widget.

cerinman:
@tumentumurchudur I already have it implemented in a branch internally but I didn't set it up that way here in the demo because I felt like the use case didn't need to be represented here.  I can post the gif of it working internally if you want.  I should also note that I reviewed the animations with Product and they signed off.

tumentumurchudur:
@cerinman You did a good job implementing this without giving much control to D3 of the DOM but I think it has so many moving parts. I would be interested in seeing how it works the way it functions in the Spending widget because essentially, that is what we are after in terms of animations.

src/components/DonutChart.js
```diff
@@ -61,101 +67,209 @@ const DonutChart = React.createClass({
...
+    return !_isEqual(this.props, nextProps) || !_isEqual(this.state, nextState);
+  },
+
+  _setupD3Functions (props) {
+    const dataSets = props.data.map(item => {
+      return item.value;
+    });
+    const valueTotal = dataSets.length ? dataSets.reduce((a, b) => {
+      return a + b;
+    }) : 0;
```
paniclater: I played around with reduce and you can accomplish the map and reduce in one step like so:
```
const valueTotal = props.data.reduce((previous, current) => {
  return { value: previous.value + current.value };
}).value;
```
Your way might be more readable? But I had fun poking around with reduce and wanted to share and see what you think.

paniclater: nevermind, I see you're using dataSets elsewhere, so skipping the map doesn't make sense.

cerinman: That is pretty cool and very good info.  I didn't know you could do that.  In this case though I need the results of the map and the results of the reduce to be separate.  I only use the reduce results to determine the end angle of the chart where the map results are used to determine the values of each arc in the chart.

paniclater: yeah just saw that. it was fun though!

paniclater:
Great job! I'm impressed! :shipit:

cerinman:
@tumentumurchudur here is a quick gif of this working internally.  You can also come by and I can demo it live for you at my desk.  I agree it has more moving parts but everything I've looked at in other React D3 libraries point to this being the way forward for now.

PR #222, opened 2016-03-16
updating eslint and webpack

paniclater:
I updated the github pages to use the new eslint-config-mx.

This then caused a few other dependency problems, which were solved by finding the right eslint and babel versions for the package.json

Finally I updated highlight.js & added a rule for webpack to not try and parse the autoit.js file it uses - this file contains 
>thousands of string concatenations and that is causing web pack to blow it's stack.
https://github.com/isagalaev/highlight.js/issues/895

I tested removing my node_modules, installing and running webpack and then locally viewing the files and everything is working in the demos -- including syntax highlighting which is what highlight.js is providing for us.

@mxenabled/frontend-engineers


bsbeeks:
:shipit:

PR #225, opened 2016-03-22
Add reset to errors on file select

jmophoto:
When a file is dropped onto the drop zone, the invalidMessage state is set to null. This adds that same behavior to the method that handles the file selection.

@mxenabled/frontend-engineers


tumentumurchudur:
:shipit:

PR #226, opened 2016-03-22
fixed icon values for apple, android, and windows

scotttball:
I changed the icon values to match the other lowercase values to stay consistent :+1: @mxenabled/frontend-engineers


bsbeeks:
:shipit:

PR #229, opened 2016-03-23
adjusted button hover and active states

lukeschunk:
Based on an offline conversation I had with @derek-boman we decided to adjust the hover and the active states so they're more consistent across the board. 

In general the pattern is `normal -> hover -> active` and for the color `color -> darker -> darkest `

Note this doesn't effect the base or disabled buttons.


bsbeeks:
:shipit:

PR #230, opened 2016-03-23
Luke/select bug

lukeschunk:
### First pass at fixing a bug where select component would start from top no matter what
* I don't believe that this is a perfect implementation - I know the callback in `setState` is an odd pattern. Looking for other ideas on how to make this work 
* Also adding an ID and using `getElementById` seems cumbersome but again, I'm still working out a better way. If you use the existing classname `mx-select-options` will not work in this case (2 elements with that class)

### BEFORE: 
* Notice how when an item is selected, then the component is reopened, it starts from the top

select_before

### AFTER:

select_after

### DIFF
--- a/src/components/Select.js
+++ b/src/components/Select.js
@@ -35,11 +35,13 @@ const Select = React.createClass({
   },
 
   getInitialState () {
+    console.log("getInitial Stateis running");
     return {
       highlightedValue: null,
       isOpen: false,
       selected: false,
-      hoverItem: null
+      hoverItem: null,
+      scrollLocation: 0
     };
   },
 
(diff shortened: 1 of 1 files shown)


src/components/Select.js
```diff
@@ -66,16 +67,25 @@ const Select = React.createClass({
     if (!isMobile) {
       this.setState({
         isOpen: !this.state.isOpen
+      }, () => {
+        if (document.getElementById('items')) {
+          const items = document.getElementById('items');
+
+          items.scrollTop = this.state.scrollLocation;
+        }
```
lukeschunk: I know this is weird to do a callback in `setState` but I haven't figured out a good way to get around it.  If i  do something like this:
```javascript
_handleClick () {
  if (!isMobile) {
    this.setState({
      isOpen: !this.state.isOpen
    });
  }

  this.setScrollPosition();
},

setScrollPosition () {
  if (document.getElementById('items')) {
    const items = document.getElementById('items');

    items.scrollTop = this.state.scrollLocation;
  }
},
```
this.state.scrollLocation` is not set to the proper value when the select component is closed, and thus it won't fix the problem

tumentumurchudur: I think what is happening is that when you call `setScrollPosition`, the DOM is already done updating. As soon as the `state` of your component changes, `render` is called to update the component.

lukeschunk: thanks for taking a look @tumentumurchudur I'll look into that

tumentumurchudur: With that being said, I would move it to the `render` as below:
```javascript
if (this.state.isOpen) {
   const items = document.getElementById('items');

   if (items) {
     items.scrollTop = this.state.scrollLocation;
   }
}
```

cerinman:
@lukeschunk This looks like it never went anywhere.  Can we close this in favor of your select refactor?

lukeschunk:
sure

PR #231, opened 2016-03-23
Add Loading Button

### LINKED ISSUE #176: Add "Loading Button" Component

iheartrachie:
This redoes PR #209 which updates the Button component to add a spinner and should close #176. It works on every button type but disabled.

## Example
```
<Button icon='add' isActive={this.state.spinnerIconOnlyIsActive} onClick={this._handleSpinnerIconOnlyClick} type='primaryOutline' />
<Button icon='delete' isActive={this.state.spinnerIsActive} onClick={this._handleSpinnerClick} type='secondary'>Button with text & without actionText</Button>
<Button actionText='Spinning...' isActive={this.state.spinnerWithTextIsActive} onClick={this._handleSpinnerWithTextClick}>Button with text and actionText</Button>
```
deqovj4ccu

## Props

`isActive` *boolean* **optional**
Default: false
A boolean that is toggled when the button is clicked and when the item is finished loading.

```
_handleSpinnerClick () {
  this.setState({
    spinnerIsActive: !this.state.spinnerIsActive
  });
},

<Button isActive={this.state.spinnerIsActive} onClick={this._handleSpinnerClick}>Spinner</Button>
```
<br/>

`actionText` *string* **optional**
The button text when isActive is true. If not defined, a spinner without text is shown.
```
<Button isActive={this.state.spinnerIsActive} onClick={this._handleSpinnerClick}>Spinner</Button>
```

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -38,9 +38,9 @@ const Button = React.createClass({
       return (
         <div>
         <Spin direction='counterclockwise'>
-            <Icon size='20' style={[styles.icon, styles.spinnerStyles]} type='spinner' />
+            <Icon size='20' style={[styles.icon, styles.spinner]} type='spinner' />
         </Spin>
-          {this.props.actionText ? <div style={styles.actionTextStyles}> {this.props.actionText} </div> : null }
+          {this.props.actionText ? <div style={styles.actionText}> {this.props.actionText} </div> : null }
         </div>
       );
     } else {
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -31,30 +31,20 @@ const Button = React.createClass({
     };
   },
 
-  _renderButtonContent () {
-    const styles = this.styles();
-
-    if (this.props.isActive) {
-      return (
-        <div>
-        <Spin direction='counterclockwise'>
-            <Icon size='20' style={[styles.icon, styles.spinner]} type='spinner' />
-        </Spin>
-          {this.props.actionText ? <div style={styles.actionText}> {this.props.actionText} </div> : null }
-        </div>
-      );
-    } else {
-      return this.props.children;
-    }
-  },
-
   render () {
     const styles = this.styles();
 
     return (
       <div {...this.props} style={[styles.component, styles[this.props.type], this.props.style]}>
         {this.props.icon && !this.props.isActive ? <Icon size={20} style={styles.icon} type={this.props.icon} /> : null}
-        {this._renderButtonContent()}
+        {this.props.isActive ? (
+          <div>
+            <Spin direction='counterclockwise'>
+                <Icon size='20' style={[styles.icon, styles.spinner]} type='spinner' />
+            </Spin>
+              {this.props.actionText ? <div style={styles.actionText}> {this.props.actionText} </div> : null }
+          </div>
+        ) : this.props.children}
       </div>
     );
   },
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -165,12 +165,12 @@ const Button = React.createClass({
         marginLeft: -5,
         marginRight: this.props.children ? 5 : -5
       },
-      spinnerStyles: {
+      spinner: {
         verticalAlign: 'initial',
         marginRight: -5,
         padding: !this.props.actionText && this.props.children ? 3 : 0
       },
-      actionTextStyles: {
+      actionText: {
         display: 'inline-block',
         paddingLeft: 10
       }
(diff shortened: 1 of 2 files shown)


src/components/Button.js
```diff
@@ -21,11 +25,41 @@ const Button = React.createClass({
...
+      marginBottom: -5,
+      marginLeft: -5,
+      marginRight: -5,
+      padding: !this.props.actionText && this.props.children ? 3 : 0
+    };
+
+    const actionTextStyles = {
+      display: 'inline-block',
+      paddingLeft: 10
+    };
```
cerinman: We've established a new pattern for component styles.  We've started adding a `styles` function at the bottom of the component below the main render function.  This new `styles` function just return an object that has all the styles.  Then in other functions if we need those styles we just make a call like so.

```javascript
const styles = this.styles()'

return <div style={styles.divStyle} />
```

Since your touching the styles in the render function and adding these here as well I would suggest going ahead and moving all the styles to a styles function to match the new pattern.  Come bug me if you have questions.

cerinman:
Looking good @iheartrachie! Just the one thing around styles that I can see.

src/components/Button.js
iheartrachie: Is declaring the styles variable on lines 35 & 52 okay or is there a better place to declare this once?

src/components/Button.js
```diff
@@ -21,13 +25,42 @@ const Button = React.createClass({
...
   getDefaultProps () {
     return {
+      isActive: false,
       primaryColor: StyleConstants.Colors.PRIMARY,
       type: 'primary'
     };
   },
 
+  _renderButtonContent () {
+    const styles = this.styles();
```
cerinman: Since your not using the styles object for the else portion of the if statement, I would move this into the the `this.props.isActive` portion.  No need to declare the variable if it won't be used.

src/components/Button.js
```diff
@@ -21,13 +25,42 @@ const Button = React.createClass({
...
+            <Icon size='20' style={[styles.icon, styles.spinner]} type='spinner' />
+        </Spin>
+          {this.props.actionText ? <div style={styles.actionText}> {this.props.actionText} </div> : null }
+        </div>
+      );
+    } else {
+      return this.props.children;
+    }
+  },
+
```
scotttball: can this be moved to the render?

cerinman: Yes, we're starting to lean toward not making extra render functions if the logic is simple enough to include in the main render function.

So in the main render function you can do something like.

```javascript
{this.props.isActive ? (
  <div>
    <Spin direction='counterclockwise'>
      <Icon size='20' style={[styles.icon, styles.spinner]} type='spinner' />
    </Spin>
    {this.props.actionText ? <div style={styles.actionText}> {this.props.actionText} </div> : null }
  </div>
) : this.props.children)
```

derek-boman:
is this complete then? If so, I have a place Jason O. could use it.

src/components/Button.js
```diff
@@ -21,13 +25,32 @@ const Button = React.createClass({
...
     };
   },
 
   render () {
-    const styles = {
+    const styles = this.styles();
+
+    return (
+      <div {...this.props} style={[styles.component, styles[this.props.type], this.props.style]}>
+        {this.props.icon && !this.props.isActive ? <Icon size={20} style={styles.icon} type={this.props.icon} /> : null}
```
lukeschunk: I believe if there are three or more props, they should go on individual lines: 
```<Icon
  size={20}
  style={styles.icon}
  type={this.props.icon}
/>
```

cerinman: I think the rule is actually if there are more than three.  I think it's fine as is.

https://github.com/mxenabled/eslint-config-mx/blob/master/index.js#L158

cerinman:
:shipit: @iheartrachie @derek-boman  I'll get this merged and released this afternoon after lunch.

src/components/Button.js
lukeschunk: :+1:

PR #232, opened 2016-03-24
Add footerStyle to modal and use Button component

jmophoto:
This adds a footerStyle prop to the modal component and switches the buttons to use the Button component.

This allows for more control of the style of the footer and the ability to add icons to the buttons.

@mxenabled/frontend-engineers 

After (passin in different padding that the default):
screen shot 2016-03-24 at 9 29 39 am

Before:
screen shot 2016-03-24 at 9 38 01 am

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -640,11 +640,13 @@ const Demo = React.createClass({
         <Modal
           buttons={[
             {
+              icon: 'close',
               label: 'Secondary',
               onClick: this._handleModalSecondaryClick,
               type: 'secondary'
             },
             {
+              icon: 'rocket',
               label: 'Primary',
               onClick: this._handleModalPrimaryClick,
               type: 'primary'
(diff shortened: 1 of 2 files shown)


cerinman:
:shipit:

demo/app.js
```diff
@@ -655,6 +657,7 @@ const Demo = React.createClass({
               Footer Content
             </div>
           )}
+          footerStyle={{ padding: '40px 20px' }}
```
scotttball: What's convention on handling inline css or putting it at the bottom and calling it?

jmophoto: Normally we put it all at the bottom and calling it. The app.ls is the demo we use for testing changes and is a little unwieldy.

cerinman: In the demo app file we just inline it.  Not sure why we went that way but I think it was more for ease of setting up demos.  Everywhere else though we always do the style object/function.

scotttball: :+1:

PR #234, opened 2016-03-28
Fix button actionText on modal to use props.

jmophoto:
In the last PR, the actionText for the buttons was hard coded to be 'Loading'. This changes that to pull from the actionText prop.

@mxenabled/frontend-engineers


cerinman:
:shipit:

PR #238, opened 2016-03-30
luke/moved handle modal open to parent component

lukeschunk:
@bsbeeks I might have misunderstood the reasoning for keeping the prop `isOpen` around. This seems to be working fine in the demo app, but perhaps there's something I'm not considering. 

What I did:
* Removed the conditional from the main render method in `Modal` - it now renders no matter what
* Added a ternary in the parent component `demo` that is dependent on `this.state.showModal`

### DIFF
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -160,9 +160,11 @@ const Modal = React.createClass({
   },
 
   render () {
+    /*eslint-disable */
     if (this.props.isOpen) {
       console.warn('WARNING: The prop "isOpen" is depracated in this version of the component. Please handle Modal opening from its parent.');
     }
+    /*eslint-enable */
 
     return (
       <div className='mx-modal' style={[styles.scrim, this.props.isRelative && styles.relative]}>
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -160,8 +160,8 @@ const Modal = React.createClass({
   },
 
   render () {
-    if(this.props.isOpen) {
-      console.warn('WARNING: The prop "isOpen" is depracated in this version of the component. Please handle Modal opening from its parent.')
+    if (this.props.isOpen) {
+      console.warn('WARNING: The prop "isOpen" is depracated in this version of the component. Please handle Modal opening from its parent.');
     }
 
     return (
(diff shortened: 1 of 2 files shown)


bsbeeks:
If by default `isOpen` is false (and not displaying), then by making this change the default changes to displaying. So a bunch of modals will start showing when they shouldn't. In additional any logic we have in other places that simply toggles the `isOpen` prop will not longer work.

bsbeeks:
I think the best approach is to remove the `isOpen` prop, like you have, and add a warning or error if `this.props.isOpen` is defined. That way uses will have some indication that they need to make changes. We'll also want to release this as a breaking change (so bump the second number in the version).

lukeschunk:
Okay that makes sense - I'll work on adding that warning.

lukeschunk:
@bsbeeks check out the conditional and warning I added in the render method - of course travis is failing since it's not in props validation so I'm going to look up `eslint ignore line` like you mentioned

src/components/Modal.js
```diff
@@ -162,34 +160,35 @@ const Modal = React.createClass({
...
-            {this._renderTitleBar()}
-            <div className='mx-modal-content' style={[styles.content, this.props.contentStyle]}>
-              {this.props.children}
-              {this._renderTooltip()}
-            </div>
-            {this._renderFooter()}
+    /*eslint-disable */
+    if (this.props.isOpen) {
+      console.warn('WARNING: The prop "isOpen" is depracated in this version of the component. Please handle Modal opening from its parent.');
+    }
```
bsbeeks: So this wouldn't only work if `this.props.isOpen` is set to true. We want to check to see if the prop is defined. So something like this would be better: `if (this.props.isOpen !== null) {`.

We should also move this into the `componentDidMount` or the `componentWillMount` method.

lukeschunk: ohhh you're right. should have thought of that.

bsbeeks:
I think this is good.

bsbeeks:
:shipit:

lukeschunk:
Realized I had to add the `this.props.isOpen !== null` rather than just checking for the true version

bsbeeks:
:shipit:

PR #240, opened 2016-03-30
Button Style Fixes

### LINKED ISSUE #239: Button Dimension Inconsistencies
If a button only has an icon, its height is shorter by 6px than a button with text.
screen shot 2016-03-29 at 6 35 05 pm
The dimension of buttons is set by padding and margins and thus why the height is different. I was able to work around this when creating the loading button but it's causing problems when we mix icon-only buttons and buttons with text in buttonGroups.
screen shot 2016-03-29 at 6 08 42 pm

If I try to work around this by setting the `minHeight: 15` on the button, the icon is offset. I can't override the negative margins used on the icon because it's set on the `Icon` itself.
screen shot 2016-03-29 at 6 32 24 pm

A better way to handle this at the button level rather than trying to make up for it in the respective pieces.

iheartrachie:
This updates the button styles to match the [style guide](https://app.frontify.com/d/KHQZnlj1XMIO/web-patterns#/components/buttons/button-types). It should also close #239. This is CSS only, no other changes.

## Before
wqhcorylos

## After
bdipsgzbse

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -132,14 +132,12 @@ const Button = React.createClass({
         color: this.props.primaryColor,
         fill: this.props.primaryColor,
         transition: 'all .2s ease-in',
-
+        border: '1px solid transparent',
         ':hover': {
           color: StyleConstants.adjustColor(this.props.primaryColor, -8),
           fill: StyleConstants.adjustColor(this.props.primaryColor, -8),
           transition: 'all .2s ease-in',
-          borderColor: StyleConstants.Colors.FOG,
-          borderRadius: 2,
-          borderWidth: 1
+          border: '1px solid' + StyleConstants.Colors.FOG
         },
         ':active': {
           color: StyleConstants.adjustColor(this.props.primaryColor, -16),
@@ -150,9 +148,7 @@ const Button = React.createClass({
       },
       neutral: {
         backgroundColor: 'transparent',
-        borderColor: StyleConstants.Colors.FOG,
-        borderRadius: 2,
-        borderWidth: 1,
+        border: '1px solid' + StyleConstants.Colors.FOG,
         color: this.props.primaryColor,
         ':hover': {
           backgroundColor: StyleConstants.Colors.PORCELAIN
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -63,7 +63,7 @@ const Button = React.createClass({
         fontFamily: StyleConstants.Fonts.SEMIBOLD,
         cursor: 'pointer',
         transition: 'all .2s ease-in',
-        minWidth: 50,
+        minWidth: 16,
         minHeight: 15
       },
       primary: {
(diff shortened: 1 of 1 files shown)


src/components/Button.js
```diff
@@ -135,30 +136,47 @@ const Button = React.createClass({
         ':hover': {
           color: StyleConstants.adjustColor(this.props.primaryColor, -8),
           fill: StyleConstants.adjustColor(this.props.primaryColor, -8),
-          transition: 'all .2s ease-in'
+          transition: 'all .2s ease-in',
+          borderColor: StyleConstants.Colors.FOG,
+          borderRadius: 2,
+          borderWidth: 1
```
scotttball: `borderWidth` can be combined with `borderColor` = `border: '1px solid' + StyleConstants.Colors.FOG`

src/components/Button.js
```diff
@@ -135,30 +136,47 @@ const Button = React.createClass({
...
-          transition: 'all .2s ease-in'
+          transition: 'all .2s ease-in',
+          backgroundColor: StyleConstants.Colors.PORCELAIN
+        }
+      },
+      neutral: {
+        backgroundColor: 'transparent',
+        borderColor: StyleConstants.Colors.FOG,
+        borderRadius: 2,
+        borderWidth: 1,
```
scotttball: same

src/components/Button.js
```diff
@@ -131,34 +132,47 @@ const Button = React.createClass({
...
-        marginRight: this.props.children ? 5 : -5
+        marginRight: this.props.children ? 5 : -5,
+        verticalAlign: 'initial'
       },
       spinner: {
-        verticalAlign: 'initial',
         marginRight: -5,
-        padding: !this.props.actionText && this.props.children ? 3 : 0
+        marginTop: -6,
+        padding: !this.props.actionText ? 3 : 0
```
bsbeeks: It's minor, but this would read better switched around: `this.props.actionText ? 0 : 3`

src/components/Button.js
jmophoto: One gotcha on using shorthand styles is that if it is something that you might be overwriting, you can run into issues. Radium throw a warning if you mix long and short hand. To if you use shorthand for the border, and then pass in a prop to change just the width, it complains.

bsbeeks: To clarify. I believe the error is only thrown if you mix shorthand/longhand related styles. For example, mixing `border` with `border-width` will throw an error. Mixing `border` with `width` would not throw an error.

jmophoto: Yes... I should have said borderWidth. :+1:

iheartrachie: Sure enough, I'm getting said error. I'll revert those changes to break them out.

src/components/Button.js
iheartrachie: Good catch. Cleaner logic! Thanks!

bsbeeks:
:shipit:

src/components/Button.js
scotttball: @bsbeeks explained it to me offline. mah bad.

PR #241, opened 2016-03-31
Add event to toggle click return

jmophoto:
Return the value and event on a toggle click.

This allows for use of `e.stopPropagation();` in the function passed to the component.

@mxenabled/frontend-engineers


bsbeeks:
:shipit:

Issue #184, opened 2016-02-25
New Button Style / Combo Button (button group with dropdown)


derek-boman:
#### Overview

Not sure the best way to implement this but wanted to open the discussion. Currently in Insight we have Button Groups like those shown below. This is a pattern I can see us using a lot in new widgets designs. (1) Does it make sense to standardize/componentize they way this is implemented? 

<img width="463" alt="screen shot 2016-02-01 at 10 17 26 am" src="https://git.moneydesktop.com/github-enterprise-assets/0000/0152/0000/8033/00e071ee-c8cd-11e5-9402-1a472fa6b6d0.png">

(2) One variation of how I see this being used is when one action catches 90% of use cases and other possible actions are used less often. In this case the primary action is always presented while other options are accessible via a dropdown (triggered from the second button in the group).

Example: Mailchimp
![screen shot 2016-02-01 at 10 24 00 am](https://git.moneydesktop.com/github-enterprise-assets/0000/0152/0000/8035/f1fdc5cc-c8cd-11e5-9c8f-15aa84fc3c72.png)

Example: Google Calendar
<img width="510" alt="screen shot 2016-02-01 at 10 24 56 am" src="https://git.moneydesktop.com/github-enterprise-assets/0000/0152/0000/8036/12b35c64-c8ce-11e5-8561-aeab06de91d2.png">

Mockup Example: Cash Flow 
![screen shot 2016-02-01 at 10 39 02 am](https://git.moneydesktop.com/github-enterprise-assets/0000/0152/0000/8039/10c9e042-c8d0-11e5-829c-7981f66b9bea.png)

I'll open a related issue for a simple dropdown menu component that I think we'll also use frequently.

@bsbeeks thoughts from you or your team?

iheartrachie:
I'm going to take a stab at the Button Group. @derek-boman, I'm assuming that all the buttons will need to be the same type?

derek-boman:
@iheartrachie: not sure what you mean by the_ same type_. If you're asking if you can have a button group with a secondary button on the left and a primary button on the right, the answer is no. They need to all be the same.

iheartrachie:
@derek-boman I got the `primaryOutline` type working, moving to `base` now. Should `neutral` buttons be outlined in a similar fashion on hover?

derek-boman:
[...]

PR #242, opened 2016-03-31
ButtonGroup Component

iheartrachie:
This adds the ButtonGroup component and addresses half of #184. ButtonGroup works with every Button `type` and icon-only buttons. Loading buttons are *not* supported.

## Example
```
<ButtonGroup
  buttons={[
    { text: 'Mar 2015 - Feb 2016' }
  ]}
  controlArrows={true}
  type='primaryOutline' />

<ButtonGroup
  buttons={[
    { icon: 'download' },
    { icon: 'search' },
    { icon: 'add' }
  ]}
  type='base' />
```
6pg324n8a1

## Props
`buttons` *array of objects* **required**
Default: an empty array
An array of objects that will populate the button values. Works with as little as one object. Objects take an `icon`, `text`, or both.

`icon` *string* **optional**
The name of the icon.
Default: undefined

`text` *string* **optional**
The text to be displayed in the button
Default: undefined
```
<ButtonGroup
  buttons={[
     { icon: 'download' },
     { icon: 'search' },
     { icon: 'add' }
  ]} />
```
<br/>
`controlArrows` *boolean* **optional**
Default: false
A boolean value to add arrows to the beginning and end of the button array.
```
<ButtonGroup
  buttons={[
    { text: 'Mar 2015 - Feb 2016' }
  ]}
  controlArrows={true} />
```
<br/>
`primaryColor` *string* **optional**
Default: PRIMARY
A string that changes the brand color if there is one.
```
<ButtonGroup
  buttons={[
     { text: 'Button 1' },
     { text: 'Button 2' }
  ]}
  primaryColor='#FFC0CB' />
```
<br/>
`type` *string* **optional**
Default: primaryOutline
The type of buttons for the entire group. The same types that are supported in the Button component work here.
```
<ButtonGroup
  buttons={[
     { text: 'Button 1' },
     { text: 'Button 2' }
  ]}
  type='base' />
```

### DIFF
--- a/src/components/ButtonGroup.js
+++ b/src/components/ButtonGroup.js
@@ -53,7 +53,7 @@ const ButtonGroup = React.createClass({
               ]}
               type={this.props.type}
             >
-                {button.text}
+              {button.text}
             </Button>);
         })}
       </div>
--- a/src/components/ButtonGroup.js
+++ b/src/components/ButtonGroup.js
@@ -11,6 +11,7 @@ const ButtonGroup = React.createClass({
       icon: React.PropTypes.string,
       text: React.PropTypes.string
     })),
+    controlAarows: React.PropTypes.bool,
     type: React.PropTypes.oneOf([
       'base',
       'disabled',
@@ -23,8 +24,14 @@ const ButtonGroup = React.createClass({
   getDefaultProps () {
     return {
       buttons: [],
+      controlAarows: false,
       type: 'primaryOutline'
     };
+  },
+
+  _getAarows () {
+    this.props.buttons.unshift({ icon: 'caret-left' });
+    this.props.buttons.push({ icon: 'caret-right' });
   },
 
   render () {
@@ -34,15 +41,17 @@ const ButtonGroup = React.createClass({
 
     return (
       <div {...this.props}>
+        {this.props.controlAarows ? this._getAarows : null}
         {this.props.buttons.map(function (button, index, arr) {
           const isFirstChild = index === 0;
           const isLastChild = index === arr.length - 1;
 
           return (
             <Button
+              icon={button.icon}
               key={index}
               style={[styles.component,
-                hasHoverStyles && styles.hoverStyles,
+                hasHoverStyles && styles.buttonHover,
                 isFirstChild && styles.firstChild,
                 isLastChild && styles.lastChild]}
               type={buttonType}>
@@ -57,7 +66,8 @@ const ButtonGroup = React.createClass({
     return {
       component: {
         borderRadius: 0,
-        borderRightWidth: 0
+        borderRightWidth: 0,
+        verticalAlign: 'middle'
       },
       firstChild: {
         borderRadius: '2px 0 0 2px'
@@ -66,7 +76,7 @@ const ButtonGroup = React.createClass({
         borderRadius: '0 2px 2px 0',
         borderRightWidth: 1
       },
-      hoverStyles: {
+      buttonHover: {
         borderRightWidth: 1,
         ':hover': {
           borderColor: StyleConstants.Colors.FOG,
--- a/src/components/ButtonGroup.js
+++ b/src/components/ButtonGroup.js
@@ -1,0 +1,67 @@
+const React = require('react');
+const Radium = require('radium');
+
+const Button = require('./Button');
+
+const ButtonGroup = React.createClass({
+  propTypes: {
+    buttons: React.PropTypes.arrayOf(React.PropTypes.shape({
+      icon: React.PropTypes.string,
+      text: React.PropTypes.string
+    })),
+    type: React.PropTypes.oneOf([
+      'base',
+      'disabled',
+      'neutral',
+      'primary',
+      'primaryOutline'
+    ])
+  },
+
+  getDefaultProps () {
+    return {
+      buttons: [],
+      type: 'primary'
+    };
+  },
+
+  render () {
+    const styles = this.styles();
+    const buttonType = this.props.type;
+
+    return (
+      <div {...this.props}>
+        {this.props.buttons.map(function (button, index, arr) {
+          const isFirstChild = index === 0;
+          const isLastChild = index === arr.length - 1;
+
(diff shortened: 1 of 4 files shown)


iheartrachie:
I'm setting the button's key as the array index. Will that be enough to handle onClick (especially the controlArrows)? Is there a better way to do that?

src/components/ButtonGroup.js
```diff
@@ -0,0 +1,101 @@
...
+
+const StyleConstants = require('../constants/Style');
+
+const ButtonGroup = React.createClass({
+  propTypes: {
+    buttons: React.PropTypes.arrayOf(React.PropTypes.shape({
+      icon: React.PropTypes.string,
+      text: React.PropTypes.string
+    }).isRequired),
+    controlArrows: React.PropTypes.bool,
```
bsbeeks: I'm thinking we don't provide a prop or any type of specific support for these arrows. If a consumer of this component needs arrow buttons, they can just pass them in through the `buttons` prop.

src/components/ButtonGroup.js
```diff
@@ -0,0 +1,101 @@
...
+
+  render () {
+    const styles = this.styles();
+    const buttonType = this.props.type;
+    const primaryColor = this.props.primaryColor;
+
+    return (
+      <div {...this.props}>
+        {this.props.controlArrows ? this._getArrows() : null}
+        {this.props.buttons.map(function (button, index, arr) {
```
bsbeeks: I'd refactor this method to:
```javascript
{this.props.buttons.map((button, i) => {
  const isFirstChild = i === 0;
  const isLastChild = i === this.props.buttons.length - 1;
  const isOnlyChild = isFirstChild && isLastChild;

  // rest of method
})}
```

src/components/ButtonGroup.js
```diff
@@ -0,0 +1,101 @@
...
+              key={index}
+              primaryColor={primaryColor}
+              style={[
+                styles.component,
+                styles[buttonType],
+                isFirstChild && styles.firstChild,
+                isLastChild && styles.lastChild,
+                isOnlyChild && styles.onlyChild
+              ]}
+              type={buttonType}>
```
bsbeeks: End caret goes on it's own line

src/components/ButtonGroup.js
iheartrachie: Since I get undefined on props inside this method (I'm assuming this is because `this` is being scoped to the method?) I created a const and went about it that way.

src/components/ButtonGroup.js
iheartrachie: Is this rule the same on self-closing tags like when I make a ButtonGroup?

src/components/ButtonGroup.js
```diff
@@ -0,0 +1,101 @@
...
+
+  styles () {
+    return {
+      component: {
+        borderRadius: 0,
+        borderWidth: 1,
+        borderRightWidth: 0
+      },
+      firstChild: {
+        borderRadius: '2px 0 0 2px'
```
scotttball: I believe you can do shorthand `'2px 0'`

src/components/ButtonGroup.js
```diff
@@ -0,0 +1,101 @@
...
+      },
+      lastChild: {
+        borderRadius: '0 2px 2px 0',
```
scotttball: `'0 2px'` It's nitpicky, but @bsbeeks has instilled that in us :smile:

src/components/ButtonGroup.js
```diff
@@ -0,0 +1,101 @@
...
+
+  _getArrows () {
+    this.props.buttons.unshift({ icon: 'caret-left' });
+    this.props.buttons.push({ icon: 'caret-right' });
+  },
+
+  render () {
+    const styles = this.styles();
+    const buttonType = this.props.type;
+    const primaryColor = this.props.primaryColor;
```
scotttball: I probably wouldn't pull `buttonType` or `primaryColor` in a const because you're only using them once here. There's no duplication.

src/components/ButtonGroup.js
lukeschunk: Since this is top and left only I think this is OK. @scotttball  yours I believe would set top and bottom to 2 and right and left to 0

scotttball: too true. I misread.

src/components/ButtonGroup.js
```diff
@@ -0,0 +1,95 @@
...
+          const isOnlyChild = isFirstChild && isLastChild;
+
+          return (
+            <Button
+              icon={button.icon}
+              key={i}
+              primaryColor={primaryColor}
+              style={[
+                styles.component,
+                styles[buttonType],
```
jmophoto: I may be missing something, but I am not seeing these styles defined. Should there be a styles.primaryOutline there or is that handled in the Button component?

bsbeeks: This maps to `styles.base`

src/components/ButtonGroup.js
```diff
@@ -31,32 +31,29 @@ const ButtonGroup = React.createClass({
...
                 isFirstChild && styles.firstChild,
                 isLastChild && styles.lastChild,
                 isOnlyChild && styles.onlyChild
               ]}
-              type={buttonType}
+              type={this.props.type}
             >
-                {button.text}
+              {button.text}
             </Button>);
```
scotttball:
```
  </Button>
);
```

src/components/ButtonGroup.js
```diff
@@ -31,32 +31,29 @@ const ButtonGroup = React.createClass({
 
   render () {
     const styles = this.styles();
-    const buttonType = this.props.type;
-    const primaryColor = this.props.primaryColor;
-    const buttons = this.props.buttons;
 
     return (
       <div {...this.props}>
```
scotttball: is this needed? @bsbeeks

bsbeeks: Yes

PR #246, opened 2016-04-05
Added navigation modal

jmophoto:
This adds a NavigationModal component. This is a simple caret (or other specified icon) that when clicked shows a modal of items.

### Props ###
- componentStyle: object with styles to be applied to the overall component
- iconSize: number to determine the size of the icon. Defaults to 30
- iconStyle: object with styles to be applied to the icon
- iconType: string to determine the icon. Defaults to `caret-down`
- items: array with each object containing content node and onClick function
- modalItemStyle: object with styles to be applied to the modal items
- modalStyle: object with styles to be applied to the modal

The naming of the component is up for suggestions.

screen shot 2016-04-05 at 4 13 35 pm

@mxenabled/frontend-engineers

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -14,6 +14,7 @@ const {
   Icon,
   Loader,
   Modal,
+  NavigationModal,
   RangeSelector,
   Select,
   SelectFullScreen,
@@ -619,6 +620,17 @@ const Demo = React.createClass({
       onPreviousClick: this._handlePreviousSiblingClick
     };
 
+    const navItems = [
+      {
+        content: 'View Details',
+        onClick: this._handleNavModalClick1
+      },
+      {
+        content: 'Edit Settings',
+        onClick: this._handleNavModalClick2
+      }
+    ];
+
     return (
       <div>
         <br/><br/>
@@ -638,6 +650,10 @@ const Demo = React.createClass({
             onFileRemove={this._handleFileChange}
             uploadedFile={this.state.uploadedFile}
           />
+        </div>
+
+        <div style={{ textAlign: 'center', width: '80%', margin: 'auto' }}>
+          <NavigationModal items={navItems}/>
         </div>
 
         <br/><br/>
--- a/demo/app.js
+++ b/demo/app.js
@@ -623,11 +623,11 @@ const Demo = React.createClass({
     const navItems = [
       {
         content: 'View Details',
-        onClick: this._handleNavModalClick1
+        onClick: this._handleNavModalClick
       },
       {
         content: 'Edit Settings',
-        onClick: this._handleNavModalClick2
+        onClick: this._handleNavModalClick
       }
     ];
 
--- a/src/components/NavigationModal.js
+++ b/src/components/NavigationModal.js
@@ -1,0 +1,83 @@
+const React = require('react');
+const _merge = require('lodash/merge');
+
+const Icon = require('./Icon');
+
+const StyleConstants = require('../constants/Style');
+
+const NavigationModal = React.createClass({
+  propTypes: {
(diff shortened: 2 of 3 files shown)


src/components/NavigationModal.js
```diff
@@ -0,0 +1,83 @@
...
+const Icon = require('./Icon');
+
+const StyleConstants = require('../constants/Style');
+
+const NavigationModal = React.createClass({
+  propTypes: {
+    componentStyle: React.PropTypes.object,
+    iconSize: React.PropTypes.number,
+    iconStyle: React.PropTypes.object,
+    iconType: React.PropTypes.string,
```
cerinman: would it make sense to combine these icon props into a single icon object prop?

src/components/NavigationModal.js
```diff
@@ -0,0 +1,83 @@
...
+  render () {
+    const styles = this.styles();
+
+    return (
+      <div onClick={this._handleToggle} style={styles.component}>
+        <Icon size={this.props.iconSize || 30} style={styles.icon} type={this.props.iconType || 'caret-down'} />
+        {this.state.showModal ? (
+          <div style={styles.modal}>
+            {this.props.items.map((item, index) => {
+              return (<div key={index} onClick={item.onClick} style={styles.modalItem}>{item.content}</div>);
```
cerinman: I would separate out things onto their own lines rather than keeping it all on a single line.

src/components/NavigationModal.js
```diff
@@ -0,0 +1,83 @@
...
+    );
+  },
+
+  styles () {
+    return {
+      component: _merge({
+        fontFamily: StyleConstants.Fonts.REGULAR,
+        position: 'relative',
+        cursor: 'pointer',
+        display: 'flex',
```
cerinman: :+1: for using flex!  :cookie:

src/components/NavigationModal.js
```diff
@@ -0,0 +1,83 @@
...
+      }, this.props.iconStyle),
+      modal: _merge({
+        borderRadius: 3,
+        boxShadow: '0 0 30px rgba(0,0,0,0.15)',
+        width: 200,
+        backgroundColor: StyleConstants.Colors.WHITE,
+        position: 'absolute',
+        top: '50%',
+        left: '50%'
+      }, this.props.modalStyle),
```
cerinman: Why did we not use Radium here?  I know we use merge in the mobile components but here there seems no need.

demo/app.js
```diff
@@ -619,6 +620,17 @@ const Demo = React.createClass({
       onPreviousClick: this._handlePreviousSiblingClick
     };
 
+    const navItems = [
+      {
+        content: 'View Details',
+        onClick: this._handleNavModalClick
```
cerinman: Not seeing where you added this function to the demo app.

src/components/NavigationModal.js
jmophoto: I am really liking Flex!

jmophoto:
Holding off on this in favor of combining it with the select menu.

src/components/NavigationModal.js
cerinman: Also these are not required and have no default values and your example in demo doesn't supply them.  I'm also not seeing how you protect against using them below in the render function for the Icon component.

cerinman: scratch the deleted comment.  I see how you're protecting against them.  Would it not be easier to supply default values for these rather than doing conditional for the Icon component?

cerinman:
:+1:

cerinman:
Can we go ahead and close this @jmophoto ?

jmophoto:
👍

PR #247, opened 2016-04-05
Luke/remove select prop

lukeschunk:
Removed some props that are unavailable on the component in the demo app. Searched the component and looked at the validation and couldn't find anything.

If this is too minor of a change I can expand this PR and go through all the components to see which props we're passing that aren't available. 

Prop Validation on select comp:
```javascript
  propTypes: {
    color: React.PropTypes.string,
    dropdownStyle: React.PropTypes.oneOfType([React.PropTypes.object, React.PropTypes.array]),
    onChange: React.PropTypes.func,
    options: React.PropTypes.array,
    optionsStyle: React.PropTypes.oneOfType([React.PropTypes.object, React.PropTypes.array]),
    optionStyle: React.PropTypes.oneOfType([React.PropTypes.object, React.PropTypes.array]),
    placeholderText: React.PropTypes.string,
    scrimStyle: React.PropTypes.oneOfType([React.PropTypes.object, React.PropTypes.array]),
    selected: React.PropTypes.object,
    selectedStyle: React.PropTypes.oneOfType([React.PropTypes.object, React.PropTypes.array]),
    valid: React.PropTypes.bool
  },
```

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -607,7 +607,7 @@ const Demo = React.createClass({
 
   _handleSpinnerWithTextClick () {
     this.setState({
-      spinnernpWithTextIsActive: !this.state.spinnerWithTextIsActive
+      spinnerWithTextIsActive: !this.state.spinnerWithTextIsActive
     });
   },
 
--- a/demo/app.js
+++ b/demo/app.js
@@ -607,7 +607,7 @@ const Demo = React.createClass({
 
   _handleSpinnerWithTextClick () {
     this.setState({
-      spinnerWithTextIsActive: !this.state.spinnerWithTextIsActive
+      spinnernpWithTextIsActive: !this.state.spinnerWithTextIsActive
     });
   },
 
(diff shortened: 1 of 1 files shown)


cerinman:
@lukeschunk if we're going to clean up the styles within the demo app then I would suggest the following.

- Do a PR for each component to keep review easy.
- Go ahead and do a styles function with a styles object like we do else where.
- kill props that are no longer in the component as you have done here.  

Might as well clean up the demo app to match our standards.

lukeschunk:
I agree it's important to clean up the demo app - when I was referencing how to use the `Select` component I looked in the demo app and saw props in use that weren't necessary. Obviously it's always good to check the component src too...anyway, I will do what you suggested

cerinman:
@lukeschunk I'm going to close this for now.  We can clean up the demo app later when we have more time.  Feel free to reopen if you disagree/find some time to work on it.

PR #248, opened 2016-04-06
Remove validation message in favor of validation callback

jmophoto:
The validation messages and placeholder text were not displaying correctly if the dropzone was smaller.

This changes the placeholder text to be a child passed into the component, allowing the user to customize the message.

This also adds an onValidation prop to be used as a callback. This allows the parent component to handle validation errors, passing messages back through the children props if desired. 

This also changes the hidden input used for the file selection to `display: none` instead of `visibility: hidden`. Before, there was an issue with a container scrolling because of the input.

@mxenabled/frontend-engineers


cerinman:
:shipit:

Issue #194, opened 2016-03-04
Icon Component: Passing an array of styles to the style prop not working


cerinman:
The `propTypes` for the `style` prop in the Icon component show it as accepting either an array or an object.  If you pass it an array, the styles in the array are not applied.

iheartrachie:
FYI, this is working in the button component where we're using the spinner icon when the button isActive. You can see the styles definition in my PR [here](https://github.com/mxenabled/mx-react-components/pull/231/files#diff-734f11edfaa69e2a96b8cb8f7f06fb6bR41).

cerinman:
Chatted with @iheartrachie on this offline.  I wondering now if our issue internally is a radium array that ends up looking like this `[object, array]` rather than `[object, object]`.  I'll take a closer look and update this issue accordingly.  Thanks for  looking into this Rachel.

PR #250, opened 2016-04-06
Remove Radium from Icon, update to use styles method

bsbeeks:
No UI changes. This should fix any issues we've been seeing related to https://github.com/mxenabled/mx-react-components/issues/194. It will be a breaking change.

// @mxenabled/frontend-engineers


lukeschunk:
:shipit: pending travis. @bsbeeks maybe we can take this breaking change up along with the modal change I recently worked on

bsbeeks:
Yeah, I'm going to do a minor (breaking) release that will include a bunch of changes.

Issue #244, opened 2016-04-04
New component / Simple Inputs


derek-boman:
We have a solid input component in place in Insight / Target that would be useful to have in the open-source. In doing so can we make sure that the component syncs up with styles for Simple Inputs from Form Elements section of Aggro style guide: https://app.frontify.com/d/KHQZnlj1XMIO/web-patterns

_From Insight / Target:_
![image](https://cloud.githubusercontent.com/assets/17180771/14257170/bb8cce76-fa59-11e5-86bd-433546dd2e6c.png)

scotttball:
Can this be closed?

derek-boman:
@scotttball @tumentumurchudur I'd like to see an example of this in use before we close it. Is this being used anywhere already?

bsbeeks:
@derek-boman you can view the docs for it which include a demo: http://mxenabled.github.io/mx-react-components/#/components/simple-input

derek-boman:
Cool. I'll check it out. Nice to see you still contributing @bsbeeks 😏

PR #252, opened 2016-04-07
SimpleInput: Adds SimpleInput component

tumentumurchudur:
https://github.com/mxenabled/mx-react-components/issues/244

This PR adds a SimpleInput component.

Props
- name: specifies the name of the component
- placeholder: puts a placeholder text when Input field is empty
- type: specifies the type of Input field
- valid: specifies whether the value of Input field is valid. 

screen shot 2016-04-07 at 11 01 19 am

@mxenabled/frontend-engineers

### DIFF
--- a/src/components/SimpleInput.js
+++ b/src/components/SimpleInput.js
@@ -6,6 +6,10 @@ const StyleConstants = require('../constants/Style');
 const Input = React.createClass({
   propTypes: {
     placeholder: React.PropTypes.string,
+    style: React.PropTypes.oneOfType([
+      React.PropTypes.array,
+      React.PropTypes.object
+    ]),
     type: React.PropTypes.string,
     valid: React.PropTypes.bool
   },
--- a/src/components/SimpleInput.js
+++ b/src/components/SimpleInput.js
@@ -7,10 +7,6 @@ const Input = React.createClass({
   propTypes: {
     name: React.PropTypes.string,
     placeholder: React.PropTypes.string,
-    style: React.PropTypes.oneOfType([
-      React.PropTypes.array,
-      React.PropTypes.object
-    ]),
     type: React.PropTypes.string,
     valid: React.PropTypes.bool
   },
--- a/src/components/SimpleInput.js
+++ b/src/components/SimpleInput.js
@@ -1,0 +1,62 @@
+const React = require('react');
+const Radium = require('radium');
+
+const StyleConstants = require('../constants/Style');
+
+const Input = React.createClass({
+  propTypes: {
+    name: React.PropTypes.string,
+    placeholder: React.PropTypes.string,
+    style: React.PropTypes.oneOfType([
+      React.PropTypes.array,
+      React.PropTypes.object
+    ]),
+    type: React.PropTypes.string,
+    valid: React.PropTypes.bool
+  },
+
+  getDefaultProps () {
+    return {
+      type: 'text',
+      valid: true
+    };
+  },
+
+  render () {
+    const styles = this.styles();
+
+    return (
+      <input
+        {...this.props}
+        name={this.props.name}
+        style={[styles.wrapper, !this.props.valid && styles.error, this.props.style]}
+        type={this.props.type}
+      />
+    );
+  },
+
+  styles () {
+    return {
+      wrapper: {
+        backgroundColor: StyleConstants.Colors.WHITE,
+        border: '1px solid ' + StyleConstants.Colors.FOG,
+        borderRadius: 3,
+        padding: 10,
+        width: '100%',
+        WebkitAppearance: 'none',
+
+        ':focus': {
+          backgroundColor: StyleConstants.Colors.WHITE,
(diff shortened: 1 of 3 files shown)


src/components/SimpleInput.js
```diff
@@ -0,0 +1,58 @@
...
+  },
+
+  render () {
+    const styles = this.styles();
+
+    return (
+      <input
+        {...this.props}
+        name={this.props.name}
+        style={[styles.wrapper, !this.props.valid && styles.error, this.props.style]}
```
cerinman: If we don't need Radium for the `:focus` to work on the wrapper, can we move to using the new `Object.assign` or `_merge` pattern that we discussed internally?

cerinman:
Just the one thing that I'm seeing.  Looks good.

src/components/SimpleInput.js
tumentumurchudur: @cerinman Used `Object.assign` to concatenate the style objects but we still use `Radium` for the `:focus`.

cerinman: I'm not seeing `Object.assign` used anywhere.  Did you mean that wanted to leave it as is because of the `:focus`/Radium dependency or did you change it and just not push up the change?

src/components/SimpleInput.js
```diff
@@ -0,0 +1,55 @@
+const React = require('react');
+const Radium = require('radium');
+
+const StyleConstants = require('../constants/Style');
+
+const Input = React.createClass({
+  propTypes: {
+    name: React.PropTypes.string,
+    placeholder: React.PropTypes.string,
+    type: React.PropTypes.string,
```
jmophoto: My first thought was if we wanted to restrict the type prop to only valid input types, but then I saw the list.

It makes me curious though about how this would handle a type of `checkbox` or even `file`

jmophoto:
Seems like this would need a default value, value and onChange prop, too.

src/components/SimpleInput.js
```diff
@@ -0,0 +1,55 @@
...
+
+  render () {
+    const styles = this.styles();
+
+    return (
+      <input
+        {...this.props}
+        name={this.props.name}
+        style={Object.assign({}, styles.wrapper, this.props.style)}
+        type={this.props.type}
```
jmophoto: Curious about the spread operator here, which could also address my earlier question about value and defaultValue...

If those are passed in as props, would they need to be pulled out or would they be included with the `...this.props`?

If so, do you still need the name and type?

src/components/SimpleInput.js
tumentumurchudur: @jmophoto That is a good point. We should keep the `type` to `text` for now. It definitely wouldn't work right with `type` being `checkbox`, `file`, `radio`, etc.

src/components/SimpleInput.js
tumentumurchudur: They would be included with the `...this.props` such as `onChange`, `onKeyUp`, `placeholder` except for `style`. We may need to pull `style` out if passed and set it to `text`. 
I don't see a need for `name`. Any thoughts on this? @mxenabled/frontend-engineers

jmophoto: :cookie: Very cool to know. I hadn't thought of that. Much more concise. I still think they should be included in propTypes, just to have a reference to what is expected/passed.

cerinman: I think the idea here is to let normal input properties to pass through with out prop type validation and then to overwrite any property that we don't want to come through via the spread operator.  

I agree that we can probably remove name but keep style and type.

tumentumurchudur: @cerinman we should remove `style` because this component won't handle different types such as `checkbox`, `radio`, etc passed through `this.props.style`. 

@jmophoto are you suggesting we add `onChange` to the `props`?

jmophoto: That was my initial thought, just for a reference, but I don't know that it is needed. 

I would leave the style prop in, though. Even for just a text input, we may want to style it. You could set the width, border, etc.

tumentumurchudur: Sorry, I meant `type` prop should be removed.

cerinman: Yes I would drop name but keep style and type.  Type could then just be set to `text` rather than using a prop to set it.  This would ensure that the input stays as type text even if the user of the component passes in a type prop as our specified type overwrites the type coming in with the `...this.props` spread.  

Unless I'm miss understanding what we're trying to achieve with enforcing type as text.

cerinman: example

```javascript
{...this.props}
style={Object.assign({}, styles.wrapper, this.props.style)}
type='text'
```

tumentumurchudur: Sounds good. Let's keep `type`.

tumentumurchudur:
Ready for review.
@mxenabled/frontend-engineers

jmophoto:
Looks good to me! :shipit:

PR #253, opened 2016-04-07
Fixed deprecation warning in modal

jmophoto:
In implementing these changes, I found that not passing isOpen into the Modal component still triggered the error. 

Since `this.props.isOpen` would be `undefined` `this.props.isOpen !== null` would be true.

This adds the lodash _has method to check.

@mxenabled/frontend-engineers

### DIFF
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -1,6 +1,5 @@
 const React = require('react');
 const Radium = require('radium');
-const _has = require('lodash/has');
 
 const Button = require('./Button');
 const Icon = require('./Icon');
@@ -61,7 +60,7 @@ const Modal = React.createClass({
 
   componentDidMount () {
   /*eslint-disable */
-    if (_has(this.props, 'isOpen')) {
+    if (this.props.hasOwnProperty('isOpen')) {
       console.warn('WARNING: The prop "isOpen" is deprecated in this version of the component. Please handle Modal opening from its parent.');
     }
   /*eslint-enable */
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -1,5 +1,6 @@
 const React = require('react');
 const Radium = require('radium');
+const _has = require('lodash/has');
 
 const Button = require('./Button');
 const Icon = require('./Icon');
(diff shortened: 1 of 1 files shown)


src/components/Modal.js
```diff
@@ -60,8 +61,8 @@ const Modal = React.createClass({
 
   componentDidMount () {
   /*eslint-disable */
-    if (this.props.isOpen !== null) {
-      console.warn('WARNING: The prop "isOpen" is depracated in this version of the component. Please handle Modal opening from its parent.');
+    if (_has(this.props, 'isOpen')) {
+      console.warn('WARNING: The prop "isOpen" is deprecated in this version of the component. Please handle Modal opening from its parent.');
```
cerinman: rather than using lodash `has`, does `this.props.hasOwnProperty('isOpen')` work?

src/components/Modal.js
```diff
@@ -60,8 +61,8 @@ const Modal = React.createClass({
 
   componentDidMount () {
   /*eslint-disable */
-    if (this.props.isOpen !== null) {
-      console.warn('WARNING: The prop "isOpen" is depracated in this version of the component. Please handle Modal opening from its parent.');
+    if (_has(this.props, 'isOpen')) {
```
tumentumurchudur: How about just `if (this.props.isOpen)`? No dependency on `lodash`

lukeschunk: I'm pretty sure `if(this.props.isOpen)` will only check to see if the prop is false, not if it exists

cerinman: In that case, wouldn't the warning NOT be shown when the prop is present but set as false?  I would think we would want to show the warning all the time if the prop is present.

tumentumurchudur: @lukeschunk it should work because  if `isOpen` does not exist, `this.props.isOpen` would be `undefined` which is a falsey value.

tumentumurchudur: Check this out https://jsfiddle.net/na7saszd/

src/components/Modal.js
tumentumurchudur: I like @cerinman's suggestion. :+1:  Here is another one.
```javascript
if (this.props.isOpen || this.props.isOpen === false)
```

src/components/Modal.js
jmophoto: I had the same thought as @tumentumurchudur, and yes, it doesn't trigger. 

As for the hasOwnProperty, I read that is isn't completely reliable on IE. Not sure we are that concerned about that, though. Thoughts?

cerinman: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty

looks to be ok with IE.

jmophoto: Excellent. I will change that.

jmophoto:
@cerinman: pushed up the hasOwnProperty change and ready for another review.

bsbeeks:
:shipit:

PR #254, opened 2016-04-08
Fix styles for Icon component in DatePicker components

cerinman:
We were passing in an array of styles to the Icon component inside the DatePicker and DatePickerFullScreen components.  This was generating an `Invalid prop` warning since we update the Icon component.  This PR fixes the warning by using `Object.assign` rather than a Radium array for the offending Icons.

@mxenabled/frontend-engineers 

After change screen shots to show styling still working as intended.

screen shot 2016-04-08 at 10 50 54 am
screen shot 2016-04-08 at 10 51 03 am


jmophoto:
:shipit:

PR #257, opened 2016-04-11
Button styles fix

### LINKED ISSUE #256: Icons are expanding loading button height
After some changes in e5a2182, icons used in buttons are throwing off the padding, height, and alignment (in button groups).
screen shot 2016-04-11 at 9 21 25 am

iheartrachie:
Fixes #256 and reworks Button and ButtonGroup components to use the `Object.assign()` pattern on styles. Radium is still required on the Button component so I didn't remove it there. Note: button height goes from 31px to 29px which is more in line with the style guide anyway.

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -169,10 +169,10 @@ const Button = React.createClass({
         fill: StyleConstants.Colors.FOG
       },
       icon: {
-        marginBottom: this.props.children ? -5 : -3,
+        marginBottom: -3,
         marginLeft: -5,
         marginRight: this.props.children ? 5 : -5,
-        marginTop: this.props.children ? -5 : -4,
+        marginTop: -4,
         verticalAlign: 'middle'
       },
       spinner: {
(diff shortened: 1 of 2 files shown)


cerinman:
:shipit:

src/components/ButtonGroup.js
```diff
@@ -90,4 +89,4 @@ const ButtonGroup = React.createClass({
   }
 });
 
-module.exports = Radium(ButtonGroup);
```
scotttball: We still need Radium for the button `:hover`

scotttball: and for line 46

src/components/Button.js
```diff
@@ -40,7 +40,7 @@ const Button = React.createClass({
         {this.props.isActive ? (
           <div>
             <Spin direction='counterclockwise'>
-                <Icon size='20' style={[styles.icon, styles.spinner]} type='spinner' />
+                <Icon size='20' style={styles.spinner} type='spinner' />
```
scotttball: `<Icon size={20} style={styles.spinner} type='spinner' />`

iheartrachie: This should be overwritten when you push your React 15 changes up.

scotttball: also, if Luke hasn't changed it already, the indent is off

PR #261, opened 2016-04-12
DisplayInput Component

### LINKED ISSUE #245: New Component / Display Input
This is a special type of input and should be considered a unique component. While the input in issue #244 is primarily for forms and creation flows with multiple inputs. This component is more for to display a list of details but allows for inline editing. In this case the user may only need to edit one (if any) of the details and it would not make sense to display info as a form. 

Some MX application examples of this pattern are: account details, transaction details, goal details, etc.

The spec can be found in Aggro style guide > Form Elements > Display Input: https://app.frontify.com/d/KHQZnlj1XMIO/web-patterns

Example: 
image

iheartrachie:
Creates the `DisplayInput` component and closes #245. The component is used for text input and is fully responsive. Its functionality should be similar to the `SimpleInput` component.

## Example
```
_handleInputShowHint () {
  this.setState({
    showHint: true
  });
},

 _handleInputHideHint () {
  this.setState({
    showHint: false
  });
},

_handleInputFocus () {
  this.setState({
    statusMessage: null,
    showHint: false
  });
},

_handleInputStatusMessage () {
  this.setState({
    statusMessage: {
      type: 'success',
      message: 'Saved!'
    },
    valid: true
  });
},

<DisplayInput
  hint='Click to Edit'
  label='Display Input'
  onBlur={this._handleInputStatusMessage}
  onFocus={this._handleInputFocus}
  onMouseOut={this._handleInputHideHint}
  onMouseOver={this._handleInputShowHint}
  placeholder='Type something'
  showHint={this.state.showHint}
  status={this.state.statusMessage}
  valid={true}
/>
```
**Successful input**
7ypatz535p

**Error on input**
lxv4tyrkz8

## Props
`hint` _string_ **optional**
Hint text to display to user on input hover

`label` _string_ **optional**
Input label shown to the left of the input field

`placeholder` _string_ **optional**
Placeholder text shown inside the input field

`primaryColor` _string_ **optional**
Default: PRIMARY
A string that changes the brand color if there is one. Used on input focus and for the hint color

`showHint` _boolean_ **optional**
A boolean that controls when the hint text is displayed.

`status` _Object_ **optional**
An Object that contains a status type (error/success) and a status message that is displayed to the right of the input field.

`valid` _boolean_ **optional**
Default: true
A boolean that marks the field as valid or not.


iheartrachie:
I'm a little unsure about how I'm doing the event handlers. They feel cumbersome to me. Any suggestions for alternatives would be greatly appreciated! :heart:

lukeschunk:
I could be wrong about this - but don't we want the mouse/click/etc handlers to be at the component level and then bubbled up to the top level via props? @cerinman ?

demo/app.js
```diff
@@ -628,6 +629,35 @@ const Demo = React.createClass({
...
+    this.setState({
+      showHint: true
+    });
+  },
+
+  _handleInputHideHint () {
+    this.setState({
+      showHint: false
+    });
+  },
```
cerinman: You can combine these two functions.

```
_toggleShowInputHint () {
  this.setstate({
    showHint: !this.state.showHint
  });
}
```

paniclater:
@lukeschunk by passing in these functions to the component we get better reusability and the benefit of having the input be able to interact with other portions of the app without having any knowledge of them.

demo/app.js
```diff
@@ -961,6 +991,18 @@ const Demo = React.createClass({
         />
         <br/><br/>
         <SearchInput />
+        <DisplayInput
+          hint='Click to Edit'
+          label='Display Input'
+          onBlur={this._handleInputStatusMessage}
+          onFocus={this._handleInputFocus}
+          onMouseOut={this._handleInputHideHint}
+          onMouseOver={this._handleInputShowHint}
```
cerinman: Ah, hadn't seen that you had `onMouseOut` and `onMouseOver` listeners added.  My previous comment about combing the two functions might not work depending on how you have all that setup.

lukeschunk:
@paniclater my idea wouldn't change the reusability - it would be like how tumen handled his recent SearchInput component. Put the click/blur handlers as props, and then put the functions on the top level. It just seems to be the pattern we usually follow - I'm not opposed to this though, just haven't seen it before.

src/components/DisplayInput.js
```diff
@@ -0,0 +1,145 @@
...
+      primaryColor: StyleConstants.Colors.PRIMARY,
+      valid: true
+    };
+  },
+
+  render () {
+    const styles = this.styles();
+
+    return (
+      <div style={Object.assign({}, styles.wrapper)}>
```
cerinman: No need for the `Object.assign` if your not merging two or more style objects together.

demo/app.js
iheartrachie: This is the best way I could find to handle a hover state. Is there a better way?

paniclater:
@lukeschunk I see what you mean now. I misunderstood what you were trying to say. I think the benefit of passing in the handlers as props the way Tumen did is that you can specify what element of the component is being targeted. If the entire component is going to respond to blur or click then I think passing them in as props is overkill.

lukeschunk:
@paniclater sounds good - that's kind of what I was wondering - the only benefit I can see is if they're props the person using the component will have an explicit reminder of 'hey if i want anything to happen on hover I have to pass a prop' - this way it might be easier to forget, but I agree it is more code and could be overkill.

src/components/DisplayInput.js
```diff
@@ -0,0 +1,145 @@
...
+  },
+
+  styles () {
+    return {
+      wrapper: Object.assign({
+        borderBottom: this.props.valid ? '1px solid ' + StyleConstants.Colors.ASH : '1px solid ' + StyleConstants.Colors.STRAWBERRY,
+        borderLeft: '1px solid transparent',
+        borderRadius: 3,
+        borderRight: '1px solid transparent',
+        borderTop: '1px solid transparent',
```
cerinman: I'm not an css expert so some one correct me if I'm wrong here, but specifying the borders for left, right, and top is not required here right?  By omitting them they should end up as none I would think.

src/components/DisplayInput.js
```diff
@@ -0,0 +1,145 @@
...
+        transition: 'all .2s ease-in',
+        WebkitAppearance: 'none',
+        whiteSpace: 'nowrap',
+        width: '100%',
+
+        ':focus': {
+          borderBottom: this.props.valid ? '1px solid ' + this.props.primaryColor : '1px solid ' + StyleConstants.Colors.STRAWBERRY,
+          borderLeft: '1px solid transparent',
+          borderRight: '1px solid transparent',
+          borderTop: '1px solid transparent',
```
cerinman: same here

demo/app.js
cerinman: @tumentumurchudur do you want to take a look at this PR when you have a sec?  You always see more than I do and maybe you might have an answer around using `onMouseOut` and `onMouseOver` for showing/hiding the hint.

src/components/DisplayInput.js
iheartrachie: Yeah you're right. I'll remove them.

src/components/DisplayInput.js
```diff
@@ -0,0 +1,145 @@
...
+      valid: true
+    };
+  },
+
+  render () {
+    const styles = this.styles();
+
+    return (
+      <div style={Object.assign({}, styles.wrapper)}>
+        {this.props.label ?
```
tumentumurchudur: I would put the `divs` in `()`
```javascript
{this.props.label ? (
  <div key='label' style={styles.label}>
    <div style={styles.labelText}>{this.props.label}</div>
  </div>) : null}
```

tumentumurchudur:
@iheartrachie  Can you post some screenshots?

tumentumurchudur:
@iheartrachie never mind. Just saw the gifs.

src/components/DisplayInput.js
```diff
@@ -0,0 +1,139 @@
...
+          {...this.props}
+          key='input'
+          label={this.props.label}
+          style={styles.input}
+          type='text'
+        />
+        <div style={styles.hint}>
+          <div style={styles.hintText}>
+            {this.props.showHint && !this.props.status ? <div>{this.props.hint}</div> : null}
+            {this.props.status ? <div style={styles[this.props.status.type]}>{this.props.status.message}</div> : null}
```
tumentumurchudur: Same here. Put `div` in `()`

src/components/DisplayInput.js
iheartrachie: Is this because I have JSX inside of JS?

src/components/DisplayInput.js
```diff
@@ -0,0 +1,138 @@
...
+    };
+  },
+
+  render () {
+    const styles = this.styles();
+
+    return (
+      <div style={styles.wrapper}>
+        {this.props.label ?
+          <div key='label' style={styles.label}>
```
tumentumurchudur: I don't see a reason to have `key='label'` here. You can remove that.

iheartrachie: Without it I get a `Radium requires each element with interactive styles to have a unique key` error.

tumentumurchudur: Yeah, Radium needs it. 👍

demo/app.js
tumentumurchudur: I think it is fine the way @iheartrachie has it. 🍪

src/components/DisplayInput.js
```diff
@@ -0,0 +1,138 @@
...
+  render () {
+    const styles = this.styles();
+
+    return (
+      <div style={styles.wrapper}>
+        {this.props.label ? (
+          <div style={styles.label}>
+            <div style={styles.labelText}>{this.props.label}</div>
+          </div>) : null
+        }
```
tumentumurchudur: Sorry for being nitpicky. Let's move `}` up next to `null}`

tumentumurchudur:
This looks awesome. :shipit:

cerinman:
thanks @tumentumurchudur!  🍪  Your reviewing skills are legendary..

PR #262, opened 2016-04-13
Luke/React to v15.0.1

lukeschunk:
When I was trying to update Gunny to React `v15.0.1` I got an error that involved two different versions of React Installed - `v0.14.x` from the open source, and `v15.0.1` from gunny. 

This PR updates React in the Open Source. It also addresses all the errors in the demo app that were caused by the update. 
* React 15 doesn't like strings of numbers without px: ie `size='20'` is bad - we decided on `size={20}`
* The DatePickerFullScreen was mad that the value was null in some cases, I emulated the normal DatePicker and added getInitialState to the full screen version. This fixed the error. 

I noticed some other things that need to be cleaned up in the components - ie potential unused state in the datePicker, but that's for another PR. 

### Important:
Because this updates our OSS components to `react 15.0.1` i'm assuming this would be a major version bump? Any thoughts on this?

### DIFF
--- a/src/components/DatePickerFullScreen.js
+++ b/src/components/DatePickerFullScreen.js
@@ -40,6 +40,16 @@ const DatePickerFullScreen = React.createClass({
     };
   },
 
+  getInitialState () {
+    return {
+      currentDate: null,
+      inputValue: this._getInputValueByDate(this.props.defaultDate),
+      isValid: true,
+      selectedDate: this.props.defaultDate,
+      showCalendar: false
+    };
+  },
+
   componentDidMount () {
     window.onkeyup = e => {
       if (e.keyCode === 27) {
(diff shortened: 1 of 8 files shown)


jmophoto:
Thinking this might have the reverse effect for the duplicate React versions for the projects that aren't updated. I would say yes to bumping the major version.

src/components/DatePickerFullScreen.js
```diff
@@ -40,6 +40,16 @@ const DatePickerFullScreen = React.createClass({
     };
   },
 
+  getInitialState () {
```
iheartrachie: Should this be `getDefaultProps` instead?

jmophoto: There are `defaultProps`, too. These are states that only the component handles.

cerinman:
Needs master merged and conflicts resolved.  I agree that this should be a major version bump.

lukeschunk:
This has been merged with master

cerinman:
:shipit: Let make sure that we we publish this we get the change log updated right away about the React 15 upgrade and how it might affect users that use our components with an app that has an early version of React.  Also make sure we give everyone internally a heads up so that they know about needing to update React internally before using this as well.

PR #265, opened 2016-04-13
Fix Eslint Errors

scotttball:
Added Travis checks to fail on linting errors (it wasn't before). I also updated the devDependencies and fixed the eslint errors.


bsbeeks:
Any reason you can't just use Eslint and Babel directly?

scotttball:
@bsbeeks I didn't think it was exiting with an error of 1 on eslint failures from how I was testing it. I just checked it now and it seems fine...

scotttball:
@bsbeeks Yeah, I guess I was wrong. I do like your little function for exiting with errors in webpack. 👍

PR #266, opened 2016-04-14
Changes styles on button from an array to an object

iheartrachie:
@cerinman It looks like the only change needed to be made the this. Everything else was the same.


cerinman:
:shipit:

PR #267, opened 2016-04-18
fixes eslint errors after updating to latest linter

iheartrachie:
I noticed after a series of recent updates that we were getting a few eslint errors. I addressed them here. No code change.


scotttball:
:shipit:

PR #268, opened 2016-04-19
Let the ToggleSwitch children have access to the state.

phantomxc:
This will let me create changing icons inside of the toggle switch.


cerinman:
:shipit:

Issue #269, opened 2016-04-20
Margin Issues on Buttons with Icons


iheartrachie:
There are two button types this is affecting:
1. **Loading Button**
The spinner styles aren't being applied when the button `isActive`, causing the button height to be expanded.
![ajmbwfnwlc](https://cloud.githubusercontent.com/assets/1916697/14682816/ac1c96f0-06e4-11e6-8de3-29d305fe1026.gif)
2. **Button Groups**
The icon's negative margins `-4px -5px -3px` are okay on the demo page but docs would require `-6px -5px -3px`. How can we make sure this is consistent?
![screen shot 2016-04-20 at 10 43 34 am](https://cloud.githubusercontent.com/assets/1916697/14682838/c400258e-06e4-11e6-89dd-fb283d11b1c0.png)

Both components look fine on the demo page but not in our documentation. This may not be an issue elsewhere, we still need to investigate.
@jmophoto

scotttball:
1. <b>Loading Button</b>
The button height is being modified because the `<Icon>` used in the animation has a fixed height on it which modifies the button height. 
2. <b>Button Groups</b>
I'll have to dig into it to figure out what the deal is with it.

scotttball:
@iheartrachie and I spoke offline and we think that the buttons in the `gh-pages` are being affected by the parent styling and overriding some of the styling for the buttons and causing this.

iheartrachie:
I verified this works without issue in an internal project so I'm going to agree with Scott. I'll fiddle around to see if I can find the cause. If not, I'll close this issue.

iheartrachie:
The button group issue will be closed by #278. The loading button issue was a Radium and React 15 issue that Scott has since resolved. I'm going to go ahead and close this issue.

PR #278, opened 2016-04-26
Button group styles fix

iheartrachie:
This is a partial fix for #269 to address the alignment issues with button groups. We were also seeing a TypeOf error in Firefox. This should resolve that as well.

@scotttball @jmophoto, I found that if I removed  `line-height` from the demo app, it messed up the buttons so I specifically declared it on the buttons too, just in case. I may not fix the issue on gh-pages though. If not I'll remove `line-height` there.


scotttball:
@iheartrachie and I spoke about this offline and setting a height for this component will keep it from expanding out of its container if someone changes the font size or line height of the component. :shipit:

PR #284, opened 2016-04-29
Remove style arrays from button, button group and modal

jmophoto:
We are getting a TypeError in Firefox with some components. I tracked this down to an issue where the button component was getting passed an array of styles from the modal and Firefox was not handling that properly. 

This adds a prop type to button to throw a warning if sending a style array and also removes radium in Modal in favor of Object.assign.

The button and button group components still require radium since they have a hover state.

This should be similar to converting the Icon over to non-array styles. It wouldn't 'break' the component, but register a React warning.


paniclater:
:shipit:

Issue #275, opened 2016-04-25
Drawer navigation revisions


derek-boman:
I've updated the Construction section of the drawer component here: https://app.frontify.com/d/KHQZnlj1XMIO/web-patterns#/components/drawer/construction

This issue is revise the drawer navigation sub-component with new stylings and button styles recently launched.

![navigation bar 1 example](https://cloud.githubusercontent.com/assets/17180771/14802096/d64142fc-0b0b-11e6-8ef9-0c7f739bc9a5.png)
![navigation bar 1](https://cloud.githubusercontent.com/assets/17180771/14802106/e4537932-0b0b-11e6-8ebc-c2ccae3be687.png)
![navigation bar 1 copy](https://cloud.githubusercontent.com/assets/17180771/14802108/e454b28e-0b0b-11e6-9ede-a75ea11ae121.png)
![navigation bar small](https://cloud.githubusercontent.com/assets/17180771/14802107/e4539c14-0b0b-11e6-859f-62fce35285c3.png)

PR #289, opened 2016-05-02
Drawer Component Update

cerinman:
### Issue
Fixes https://github.com/mxenabled/mx-react-components/issues/275

### Details
This brings the Drawer component up to date with product designs as well as making the entire component absolute positioned to fix an issue internally where the drawer was be contained within a div.

Things changed

- Cleans up App usage of Drawer
- Removes animations for items in the Drawer header bar
- Corrects style positioning/padding/margins on header bar items
- Implements flex box into header bar to make things line up better
- Replaces header bar icons with buttons of type base
- Adds Radium and media queries to updating styling depending on resolution.
- Adds new function for determining velocity animation distance for drawer to match new media queries.  NOTE: This uses the `window.innerWidth` prop.  See https://developer.mozilla.org/en-US/docs/Web/API/Window/innerWidth for details but all browsers/browser versions that we care about support it and we should be ok with using it.
- Adds title(type: string, default: '') prop to Drawer to be used within header bar.

#### Above 1200px resolution
above1200

#### Between 750px and 1200px resolutions
between750and1200

#### Below 750px resolutions
below750

#### Without title prop provided
screen shot 2016-05-02 at 4 00 54 pm

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -717,15 +717,30 @@ const Demo = React.createClass({
     return (
       <div>
         <br /><br />
-        <Button onClick={this._handleShowDrawerClick} style={{ position: 'absolute', left: 15, top: 15 }} type={this.state.showDrawerButtonType}>Toggle Drawer</Button>
-        {this.state.showDrawer ?
-          <div style={{ textAlign: 'center', width: '80%', margin: 'auto', position: 'relative', height: '200', overflow: 'hidden' }}>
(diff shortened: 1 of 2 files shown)


iheartrachie:
Do we need a prop for `primaryColor`?

iheartrachie:
Nevermind, I guess right now we only have buttons and those already have that prop.

:shipit:

PR #292, opened 2016-05-04
Remove global box-sizing from demo app and fix Drawer/Button components

cerinman:
- Kills the global box-sizing being added via the reset.css in the demo app.
- Removes un-needed heights from button and drawer components now that the demo app is working as expected.
- Fixes a position issue in the drawer component. 
- Aligns title text for drawer component.
- Adds place holder for nav element in header of the drawer component so items retain their correct position if a navConfig is not provided.

Demo app after change:
screen shot 2016-05-04 at 10 23 53 am


demo/reset.css
```diff
@@ -56,8 +56,6 @@
 /* Eric Meyer's Reset CSS v2.0 - http://cssreset.com */
 html,body,div,span,applet,object,iframe,h1,h2,h3,h4,h5,h6,p,blockquote,pre,a,abbr,acronym,address,big,cite,code,del,dfn,em,img,ins,kbd,q,s,samp,small,strike,strong,sub,sup,tt,var,b,u,i,center,dl,dt,dd,ol,ul,li,fieldset,form,label,legend,table,caption,tbody,tfoot,thead,tr,th,td,article,aside,canvas,details,embed,figure,figcaption,footer,header,hgroup,menu,nav,output,ruby,section,summary,time,mark,audio,video{border:0;font-size:100%;font:inherit;vertical-align:baseline;margin:0;padding:0}article,aside,details,figcaption,figure,footer,header,hgroup,menu,nav,section{display:block}body{line-height:1}ol,ul{list-style:none}blockquote,q{quotes:none}blockquote:before,blockquote:after,q:before,q:after{content:none}table{border-collapse:collapse;border-spacing:0}
 
-* { box-sizing: border-box; }
```
bsbeeks: Curious why you're doing this? Shouldn't the components work whether this is set or not?

cerinman: Setting it globally was causing an issue where the expected results outside the demo app was different than in the demo app.  It was causing issues and confusion when implementing internally.  We decided that it was best to apply is selectively to the component code rather than as a global property in reset.css to cut down on the confusion.

bsbeeks: 👍

bsbeeks:
:shipit:

PR #294, opened 2016-05-04
Drawer: Passes style prop to Drawer component

tumentumurchudur:
This PR allows users of the `Drawer` component to pass a style object as a prop from the parent component.

- Adds `this.props.style`


paniclater:
:shipit:

paniclater:
Glad to see this getting some use!

PR #296, opened 2016-05-05
ButtonGroup Update

### LINKED ISSUE #295: ButtonGroup doesn't have props for onClick
We need to be able to pass in onClick as part of the `buttons` array in the ButtonGroup component so that when the items are actually clickable.

iheartrachie:
This PR adds onClick prop for event handling in button groups as well as fixes a border issue. Closes #295 @jmophoto 

## Before
screen shot 2016-05-05 at 12 09 02 pm

## After
screen shot 2016-05-05 at 12 08 32 pm

### DIFF
--- a/src/components/ButtonGroup.js
+++ b/src/components/ButtonGroup.js
@@ -9,6 +9,7 @@ const ButtonGroup = React.createClass({
   propTypes: {
     buttons: React.PropTypes.arrayOf(React.PropTypes.shape({
       icon: React.PropTypes.string,
+      onClick: React.PropTypes.func,
       style: React.PropTypes.object,
       text: React.PropTypes.string
     }).isRequired),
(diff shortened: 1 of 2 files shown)


lukeschunk:
:shipit:

src/components/ButtonGroup.js
```diff
@@ -44,6 +45,7 @@ const ButtonGroup = React.createClass({
             <Button
               icon={button.icon}
               key={i}
+              onClick={button.onClick}
```
lukeschunk: sorry one thing - where is button.onClick defined?

iheartrachie: `button` is an object in the `buttons` array. I'm mapping through the array on line 39.

lukeschunk:
:shipit:

PR #297, opened 2016-05-05
Debounce the RangeSelector callbacks

phantomxc:
I noticed the callback functions provided to the props `onLowerDragStop` and `onUpperDragStop` were executed multiple times while it was being dragged. Ideally this would truly only be called once after the drag was stopped and the mouse up ( or equivalent ) event was fired but that would be harder to implement. For the time being I implemented a solution with debounce.

A perfect example of this can be found in the Insight & Target Segment Balance filter...It will send off multiple execute calls to the backend while your dragging.

### DIFF
--- a/src/components/RangeSelector.js
+++ b/src/components/RangeSelector.js
@@ -2,6 +2,7 @@ const React = require('react');
 const ReactDOM = require('react-dom');
 const Radium = require('radium');
 const _throttle = require('lodash/throttle');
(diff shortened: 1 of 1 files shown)


lukeschunk:
i like this

lukeschunk:
:shipit:

cerinman:
Yep, looks good.  Feel free to merge this when ever you're ready @phantomxc.  We can get it moving up in the next release.

PR #299, opened 2016-05-06
Button style updates

iheartrachie:
Hopefully this will be the last update to buttons for a while. There are some inconsistencies in the reset.css we're using in app.js, the docs, and our production apps. These updates will allow us to manage those inconsistencies and make the buttons more universal.

1. Icon only buttons to use transform for placement rather than relying on negative margins.
2. Allow individual buttons in button groups to be styled. (Like for overriding global box-border styles).

## Before
screen shot 2016-05-06 at 11 50 23 am

## After
screen shot 2016-05-06 at 11 50 35 am

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -37,7 +37,8 @@ const Button = React.createClass({
 
     return (
       <div {...this.props} style={Object.assign({}, styles.component, styles[this.props.type], this.props.style)}>
-        {this.props.icon && !this.props.isActive ? <Icon size={20} style={styles.icon} type={this.props.icon} /> : null}
+        {(this.props.icon && !this.props.isActive && !this.props.children) ? <Icon size={20} style={styles.iconOnly} type={this.props.icon} /> : null}
+        {this.props.icon && !this.props.isActive && this.props.children ? <Icon size={20} style={styles.icon} type={this.props.icon} /> : null}
         {this.props.isActive ? (
           <div>
             <Spin direction='counterclockwise'>
@@ -65,7 +66,8 @@ const Button = React.createClass({
         cursor: 'pointer',
         transition: 'all .2s ease-in',
         minWidth: 16,
-        height: 13
+        height: 13,
+        position: 'relative'
       }, this.props.style),
       primary: {
         backgroundColor: this.props.primaryColor,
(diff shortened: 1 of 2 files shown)


src/components/ButtonGroup.js
```diff
@@ -48,6 +48,7 @@ const ButtonGroup = React.createClass({
               onClick={button.onClick}
               primaryColor={this.props.primaryColor}
               style={Object.assign({},
+                button.style,
```
jmophoto: I would move `button.style` to the last item in the Object.assign or it will be overridden by `styles.component`, etc.

src/components/Button.js
```diff
@@ -37,7 +37,8 @@ const Button = React.createClass({
 
     return (
       <div {...this.props} style={Object.assign({}, styles.component, styles[this.props.type], this.props.style)}>
-        {this.props.icon && !this.props.isActive ? <Icon size={20} style={styles.icon} type={this.props.icon} /> : null}
+        {(this.props.icon && !this.props.isActive && !this.props.children) ? <Icon size={20} style={styles.iconOnly} type={this.props.icon} /> : null}
+        {this.props.icon && !this.props.isActive && this.props.children ? <Icon size={20} style={styles.icon} type={this.props.icon} /> : null}
```
jmophoto: Duplicate lines here, one with the needed `()` and one without.

src/components/ButtonGroup.js
iheartrachie: Good catch!

src/components/Button.js
iheartrachie: The third Item in the conditional (`this.props.children`) is where the difference is. ESLint throws an error when I try to nest ternary expressions. I added the parens around the second conditional for consistency though.

src/components/Button.js
```diff
@@ -38,7 +38,7 @@ const Button = React.createClass({
     return (
       <div {...this.props} style={Object.assign({}, styles.component, styles[this.props.type], this.props.style)}>
         {(this.props.icon && !this.props.isActive && !this.props.children) ? <Icon size={20} style={styles.iconOnly} type={this.props.icon} /> : null}
-        {this.props.icon && !this.props.isActive && this.props.children ? <Icon size={20} style={styles.icon} type={this.props.icon} /> : null}
+        {(this.props.icon && !this.props.isActive && this.props.children) ? <Icon size={20} style={styles.icon} type={this.props.icon} /> : null}
```
jmophoto: Instead of two conditionals, add one to the style. 

`style={this.props.children ? styles.icon : styles.iconOnly}`

jmophoto:
Looks good! :shipit:

PR #300, opened 2016-05-09
Barchart: New component for barcharts

tumentumurchudur:
This PR adds a new component `Barchart` that allows users to render barcharts from a data array.

## Props

- animateOnHover: bool enables animation for hover event on bars in the chart
- color: string applies color to the chart
- data: array contains the data to be drawn in the chart. Example: `[{label: 'Feb', value: 123}, {label: 'Mar', value: 345.34}]`
- height number determines the height of the chart
- hoverColor: string applies color to each bar in the chart on hover event
- onClick: func callback function for on click event for each bar
- onHover: func callback function for on hover event for each bar
- width: number determines the width of the barchart

ezgif com-video-to-gif

### DIFF
--- a/src/components/BarChart.js
+++ b/src/components/BarChart.js
@@ -56,11 +56,13 @@ const Rect = React.createClass({
   },
 
   _handleMouseOver (label, value, x, y) {
+    const animateDuration = 500;
+
     this.props.onHover(label, value, x, y);
 
     this.setState({
       hovering: true,
-      milliseconds: this.props.animateOnHover ? 500 : this.state.milliseconds
+      milliseconds: this.props.animateOnHover ? animateDuration : this.state.milliseconds
     });
   },
 
@@ -75,7 +77,7 @@ const Rect = React.createClass({
   },
 
   render () {
-    const animateHeight = d3.ease('back-out', .5);
+    const animateHeight = d3.ease('back-out', 0.5);
     const height = this.props.height * animateHeight(Math.min(1, this.state.milliseconds / 1000));
     const y = this.props.height - height + this.props.y;
     const style = {
--- a/src/components/BarChart.js
+++ b/src/components/BarChart.js
@@ -17,6 +17,7 @@ const SetIntervalMixin = {
 
 const Rect = React.createClass({
   propTypes: {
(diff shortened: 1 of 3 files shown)


src/components/BarChart.js
```diff
@@ -0,0 +1,219 @@
...
+        </span>
+      );
+    });
+  },
+
+  render () {
+    const styles = this.styles();
+    const data = this.props.data.map(d => {
+      return d.value;
+    });
```
cerinman: It looks like we're mapping over the data twice here in the render function.  Here and below on line 168.  Is it possible to just expect the data prop to be formatted correctly and let the formatting of the data be a responsibility of the the user of the BarChart component?  Probably not to big of a deal for small data sets but maybe we should consider it none the less.

cerinman:
Just the one questions Tumen.  Looks good.

src/components/BarChart.js
tumentumurchudur: Yes, the formatting of the data is left up to the parent component to handle and pass down to the `BarChart` component. The reason for `map` is because `.domain()` function on line 160 takes an array of values to calculate the yScale. `this.props.data` is an array of objects and each object has `label` and `value` properties.

src/components/BarChart.js
```diff
@@ -0,0 +1,219 @@
...
+
+    const yScale = d3.scale.linear()
+      .domain([0, d3.max(data)])
```
cerinman: In that case can we use d3.values so that we can drop the map above?

```javascript
d3.max(d3.values(this.props.data))
```

https://github.com/mbostock/d3/wiki/Arrays#d3_values

tumentumurchudur: This won't work because `d3.values(this.props.data)` returns an array of objects `[{label: 'Jun', value: 154}, {...}]`. I need to pass an array of values `[122, 154, 89, 155, ...]`. I can probably make it work by using `d3.values` but it would further complicate the code, especially starting on line 170. I think it is best to leave it as is.

cerinman: 👍

cerinman:
:shipit:

PR #301, opened 2016-05-09
Modal: Add primaryColor prop to Buttons prop array

cerinman:
Internally, the hover states and border colors are off for Modals because we updated the Buttons component to take in a `primaryColor` prop for setting that stuff but then never updated the Modal to use it.  This PR does just that.


iheartrachie:
:shipit:

PR #303, opened 2016-05-09
Fix left margin on button icon

jmophoto:
Found that the left margin on a button with an icon was too wide. This adds a negative margin on the icon only if there is props.children

Before:
screen shot 2016-05-09 at 2 39 36 pm

After:
screen shot 2016-05-09 at 2 39 26 pm


iheartrachie:
Jason showed me and this works with the spinner too.

iheartrachie:
:shipit:

PR #304, opened 2016-05-10
Fix for width issue on DisplayInput

iheartrachie:
The display input's width was set to 100% which after we removed box-sizing, messed things up. Removing the rule from the component fixes the issue.

## Before
screen shot 2016-05-10 at 4 10 29 pm

## After
screen shot 2016-05-10 at 4 10 51 pm
screen shot 2016-05-10 at 4 11 11 pm


jmophoto:
:shipit:

PR #305, opened 2016-05-12
Drawer: Adds style props for header and content

tumentumurchudur:
This PR adds the ability to style header and content sections of the `Drawer` component. It also aligns the label in the navigation with the next and previous carets.

Before
screen shot 2016-05-12 at 10 50 56 am

After
screen shot 2016-05-12 at 11 04 50 am


paniclater:
:shipit:

PR #307, opened 2016-05-13
RangeSelector updateOnDrag

jmophoto:
This is an alternative to @phantomxc's `_debounce` changes.

Instead of adding a _debounce to limit the returns on drag, this adds an `updateOnDrag` prop that is defaulted to `false`.

- When true, the range selector will return a value on each interval as it did before.
- When false, the range selector will only return a value on drag stop.

This also fixes an issue where the track click was returning a value that was two clicks ago, ie:
1. Upper is 30
2. Click to 45
3. Position is changed to 45 but value of 30 is sent
4. Click to 50
5. Position is changed to 50 but value of 45 is sent

Also moved styles from a const to a method. Thus the big changes in the `render`.

### DIFF
--- a/src/components/RangeSelector.js
+++ b/src/components/RangeSelector.js
@@ -17,6 +17,7 @@ const RangeSelector = React.createClass({
     onUpperDragStop: React.PropTypes.func,
     presets: React.PropTypes.array,
     selectedColor: React.PropTypes.string,
+    updateOnDrag: React.PropTypes.bool,
     upperBound: React.PropTypes.number
   },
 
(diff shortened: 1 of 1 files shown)


src/components/RangeSelector.js
```diff
@@ -150,7 +144,7 @@ const RangeSelector = React.createClass({
       updatedState.dragging = 'Upper';
     }
 
-    this.setState(updatedState);
+    this.setState(updatedState, this._handleDragging(e));
```
jmophoto: Since the `_handleDragging` access `state`, calling it as a chain will not guaranty it gets called with the updatedState. React provides a callback on the setState that will be triggered after the state is updated.

redperk:
:shipit:

PR #309, opened 2016-05-14
DateRangePicker Component

iheartrachie:
The DateRangePicker component is based off of the DatePicker component but instead has a start and end date.

## Example:
```
<DateRangePicker
  onDateSelect={this._handleDateRangeSelect}
  selectedEndDate={this.state.selectedEndDate}
  selectedStartDate={this.state.selectedStartDate}
/>
```
datepicker

## Props

`format` _String_
Default: 'MMM D, YYYY'
The format used to display the date.

`locale` _String_
Default: 'en'
The locale of the user, used to format the timestamp.

`minimumDate` _Number_
Dates before this are not selectable.

`onDateSelect` _Function_
Function to handle date selection.

`placeholderText` _String_
Default: 'Select A Date Range'
The text show in the input field when no date is selected.

`primaryColor` _String_
Default: PRIMARY
Color used on date select and highlight.

`selectedEndDate` _Number_
The end date, to be passed in by `onDateSelect`

`selectedStartDate` _Number_
The start date, to be passed in by `onDateSelect`

`style` _Object_
Addition styles that can be set on the component.

### DIFF
--- a/src/components/DateRangePicker.js
+++ b/src/components/DateRangePicker.js
@@ -54,7 +54,6 @@ const DatePicker = React.createClass({
     updatedDates = updatedDates || this._noEndDate(date, selectedStartDate, selectedEndDate);
     updatedDates = updatedDates || this._noStartDate(date, selectedStartDate, selectedEndDate);
     updatedDates = updatedDates || this._noStartOrEndDate(date, selectedStartDate, selectedEndDate);
-    updatedDates = updatedDates || this._deselectDate(date, selectedStartDate, selectedEndDate);
 
     this.props.onDateSelect(updatedDates[0], updatedDates[1]);
   },
--- a/src/components/DateRangePicker.js
+++ b/src/components/DateRangePicker.js
@@ -45,7 +45,6 @@ const DatePicker = React.createClass({
   },
 
   _handleDateSelect (date) {
-    // clears date when clicking on it again
     if (this.state.selectedStartDate === date) {
       this.state.selectedStartDate = null;
       return;
@@ -64,6 +63,11 @@ const DatePicker = React.createClass({
     console.log(this.state.selectedStartDate, this.state.selectedEndDate);
 
     this.props.onDateSelect(this.state.selectedStartDate, this.state.selectedEndDate);
+  },
+
+  _handleDateHover (date) {
+    this.state.activeSelectDate = date;
+    console.log(moment.unix(date).format(this.props.format));
   },
 
   _handlePreviousClick () {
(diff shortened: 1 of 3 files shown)


src/components/DateRangePicker.js
```diff
@@ -0,0 +1,353 @@
...
+      const isSelectedStartDay = startDate.isSame(moment.unix(this.state.selectedStartDate), 'day');
+      const isSelectedEndDay = startDate.isSame(moment.unix(this.state.selectedEndDate), 'day');
+      const isActiveRange = startDate.isBetween(moment.unix(this.state.selectedStartDate), moment.unix(this.state.activeSelectDate));
+      const isSelectedRange = startDate.isBetween(moment.unix(this.state.selectedStartDate), moment.unix(this.state.selectedEndDate));
+
+      const day = (
+        <div
+          key={startDate}
+          onClick={disabledDay ? null : this._handleDateSelect.bind(null, startDate.unix())}
+          onMouseOver={(disabledDay || !this.state.selectedStartDate || this.state.selectedEndDate) ? null : this._handleDateHover.bind(null, startDate.unix())}
```
iheartrachie: `onClick` and `onMouseOver` are being called regardless of these conditionals.

src/components/DateRangePicker.js
```diff
@@ -0,0 +1,353 @@
...
+          style={Object.assign({},
+            styles.calendarDay,
+            isCurrentMonth && styles.currentMonth,
+            disabledDay && styles.calendarDayDisabled,
+            (isToday && (!isActiveRange || !isSelectedRange)) && styles.today,
+            isSelectedStartDay && Object.assign({}, styles.selectedDay, styles.selectedStartDay),
+            isSelectedEndDay && Object.assign({}, styles.selectedDay, styles.selectedEndDay),
+            (isSelectedRange || isActiveRange) && styles.betweenDay
```
iheartrachie: isActiveRange is being declared before it's set on line 117 which is why the event is one behind the user. However= even if I check to see if the dates are between here, I see the same behavior.

src/components/DateRangePicker.js
```diff
@@ -0,0 +1,353 @@
...
+
+  _handleDateSelect (date) {
+    if (this.state.selectedStartDate === date) {
+      this.state.selectedStartDate = null;
+      return;
+    }
+    if (this.state.selectedEndDate === date) {
+      this.state.selectedEndDate = null;
+      return;
+    }
```
iheartrachie: This deselects the date after it's been selected. Is there a cleaner way to do this?

cerinman:
I think this is a good first pass.  Lets roll with it and we'll adjust as needed.  :shipit:

jmophoto:
Nice work! 🍪

src/components/DateRangePicker.js
```diff
@@ -0,0 +1,462 @@
...
+    const selectedStartDate = this.props.selectedStartDate;
+    const selectedEndDate = this.props.selectedEndDate;
+    let updatedDates = [null, null];
+
+    updatedDates = this._deselectDate(date, selectedStartDate, selectedEndDate);
+    updatedDates = updatedDates || this._startAndEndDate(date, selectedStartDate, selectedEndDate);
+    updatedDates = updatedDates || this._noEndDate(date, selectedStartDate, selectedEndDate);
+    updatedDates = updatedDates || this._noStartDate(date, selectedStartDate, selectedEndDate);
+    updatedDates = updatedDates || this._noStartOrEndDate(date, selectedStartDate, selectedEndDate);
+    updatedDates = updatedDates || this._deselectDate(date, selectedStartDate, selectedEndDate);
```
paniclater: I'm confused why _deselectDate gets called on line 52 and then again on line 57 (if updatedDates is still falsey).

src/components/DateRangePicker.js
```diff
@@ -0,0 +1,462 @@
...
+    }
+  },
+
+  _handleDateSelect (date) {
+    const selectedStartDate = this.props.selectedStartDate;
+    const selectedEndDate = this.props.selectedEndDate;
+    let updatedDates = [null, null];
+
+    updatedDates = this._deselectDate(date, selectedStartDate, selectedEndDate);
+    updatedDates = updatedDates || this._startAndEndDate(date, selectedStartDate, selectedEndDate);
```
paniclater: On first read through I assumed `_startAndEndDate` was called if there were start and end dates because of the name. I see now it is called if `updatedDates` is falsey and then handles the situation where there are start and end dates.

src/components/DateRangePicker.js
iheartrachie: Thanks for catching this. I had rearranged some things and forgot to remove it. It's all better now :)

paniclater:
:shipit:

paniclater:
:shipit:

PR #310, opened 2016-05-17
Refactor of ToggleSwitch for animations, default icons and more style control

phantomxc:
- New styles property that allows you to just override any key in the style const in the render functions.
- Changed the concept of 'left' and 'right' to just a boolean being stored as this is a glorified checkbox. 
- Allow custom Icons to be passed in via props instead of children.
- Some default animations and style changes to match the renderings from product.

Before: 
ezgif com-video-to-gif 4

After:
ezgif com-video-to-gif 5

### DIFF
--- a/src/components/ToggleSwitch.js
+++ b/src/components/ToggleSwitch.js
@@ -1,89 +1,59 @@
 const React = require('react');
 const Radium = require('radium');
 const StyleConstants = require('../constants/Style');
+const Icon = require('./Icon');
 
 const ToggleSwitch = React.createClass({
   propTypes: {
-    activeColor: React.PropTypes.string,
-    children: React.PropTypes.node,
-    componentStyle: React.PropTypes.oneOfType([
-      React.PropTypes.array,
-      React.PropTypes.object
-    ]),
-    defaultPosition: React.PropTypes.oneOf(['left', 'right']),
-    inactiveColor: React.PropTypes.string,
-    labelStyle: React.PropTypes.oneOfType([
-      React.PropTypes.array,
-      React.PropTypes.object
-    ]),
+    checked: React.PropTypes.bool,
+    checkedByDefault: React.PropTypes.bool,
+    falseIcon: React.PropTypes.element,
     leftLabel: React.PropTypes.string,
     onToggle: React.PropTypes.func,
     rightLabel: React.PropTypes.string,
+    showIcons: React.PropTypes.bool,
     showLabels: React.PropTypes.bool,
-    toggleStyle: React.PropTypes.oneOfType([
-      React.PropTypes.array,
-      React.PropTypes.object
-    ]),
-    trackStyle: React.PropTypes.oneOfType([
-      React.PropTypes.array,
-      React.PropTypes.object
-    ])
+    styles: React.PropTypes.object,
+    trueIcon: React.PropTypes.element
   },
 
   getDefaultProps () {
     return {
-      activeColor: StyleConstants.Colors.PRIMARY,
-      defaultPosition: 'left',
-      inactiveColor: StyleConstants.Colors.FOG,
-      leftLabel: 'On',
+      checkedByDefault: false,
+      leftLabel: 'Off',
       onToggle () {},
-      rightLabel: 'Off',
-      showLabels: true,
-      toggleStyle: {},
-      trackStyle: {}
+      rightLabel: 'On',
+      showLabels: false,
+      showIcons: true,
+      styles: {}
     };
   },
 
   getInitialState () {
+    let checked = false;
+
+    if ('checked' in this.props) {
+      checked = this.props.checked;
+    } else if ('checkedByDefault' in this.props) {
+      checked = this.props.checkedByDefault;
+    }
     return {
-      activePosition: this.props.defaultPosition
+      checked: !!checked
     };
   },
 
-  _handleLeftLabelClick () {
-    const activePosition = 'left';
-
-    this.setState({
-      activePosition
-    });
-
-    this.props.onToggle(activePosition);
-  },
-
-  _handleRightLabelClick () {
-    const activePosition = 'right';
-
-    this.setState({
-      activePosition
-    });
-
-    this.props.onToggle(activePosition);
-  },
-
   _handleToggle (event) {
-    const activePosition = this.state.activePosition === 'left' ? 'right' : 'left';
-
     this.setState({
-      activePosition
+      checked: !this.state.checked
+    }, () => {
+      this.props.onToggle(this.state.checked, event);
     });
-
-    this.props.onToggle(activePosition, event);
   },
 
   _renderLeftLabel (styles) {
     if (this.props.showLabels) {
       return (
-        <span className='left-label' onClick={this._handleLeftLabelClick} style={[styles.label, this.props.labelStyle, this.state.activePosition === 'left' && styles.activeLabel || styles.inactiveLabel]}>{this.props.leftLabel}</span>
+        <span className='left-label' onClick={this._handleToggle} style={[styles.label, !this.state.checked && styles.activeLabel || styles.inactiveLabel]}>{this.props.leftLabel}</span>
       );
     } else {
       return null;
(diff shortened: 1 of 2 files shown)


iheartrachie:
This is 🔥 ! I like it!

src/components/ToggleSwitch.js
```diff
@@ -93,73 +63,103 @@ const ToggleSwitch = React.createClass({
...
-    // Let all the children of the toggle switch be aware of the activePosition
-    return React.Children.map(this.props.children, child => {
-      return React.cloneElement(child, { activePosition: this.state.activePosition });
-    });
+  _renderIcons (styles) {
+    if (!this.props.showIcons) {
+      return null;
+    }
+    let trueIcon = this.props.trueIcon ? this.props.trueIcon : <Icon className='true-icon' style={Object.assign({}, styles.icon, styles.trueIcon)} type='check-skinny' />;
+    let falseIcon = this.props.falseIcon ? this.props.falseIcon : <Icon className='false-icon' style={Object.assign({}, styles.icon, styles.falseIcon)} type='close-skinny' />;
```
iheartrachie: Should this be const?

src/components/ToggleSwitch.js
```diff
@@ -1,89 +1,59 @@
...
-  },
-
   _handleToggle (event) {
-    const activePosition = this.state.activePosition === 'left' ? 'right' : 'left';
-
     this.setState({
-      activePosition
+      checked: !this.state.checked
+    }, () => {
+      this.props.onToggle(this.state.checked, event);
```
iheartrachie: Must better! 🍪

src/components/ToggleSwitch.js
```diff
@@ -93,73 +63,103 @@ const ToggleSwitch = React.createClass({
...
         fontFamily: StyleConstants.FontFamily,
         fontSize: '12px',
         position: 'relative'
       },
-      inactiveLabel: {
-        color: this.props.inactiveColor
+      icon: {
+        fill: StyleConstants.Colors.WHITE,
+        position: 'absolute',
+        top: '0px',
```
iheartrachie: Don't forget to remove the px on single values like this so that it become `top: 0` instead.

src/components/ToggleSwitch.js
```diff
@@ -93,73 +63,103 @@ const ToggleSwitch = React.createClass({
...
         {this._renderLeftLabel(styles)}
-        <div className='toggle-switch-track' onClick={this._handleToggle} style={[styles.track, this.props.trackStyle]} >
-          <div className='toggle-switch-toggle' style={[styles.toggle, styles[this.state.activePosition], this.props.toggleStyle]}></div>
-          {this._getChildren()}
+        <div
+          className='toggle-switch-track'
+          onClick={this._handleToggle}
+          style={Object.assign({}, styles.track, styles[this.state.checked + 'Track'])}
+        >
+          <div className='toggle-switch-toggle' style={Object.assign({}, styles.toggle, styles[this.state.checked + 'Toggle'])}></div>
```
iheartrachie: Do we need these class names here? I know @tumentumurchudur commented about them on another PR. Not sure what use they're providing.

src/components/ToggleSwitch.js
phantomxc: ya, leftover from my own refactor.

src/components/ToggleSwitch.js
phantomxc: This just allows someone to style the elements in a normal css sheet instead of the style prop. Some further work needs to be done to add animation support with dynamic class names. It would easily be accomplished with a library such as https://github.com/JedWatson/classnames

src/components/ToggleSwitch.js
```diff
@@ -1,89 +1,59 @@
...
     showLabels: React.PropTypes.bool,
-    toggleStyle: React.PropTypes.oneOfType([
-      React.PropTypes.array,
-      React.PropTypes.object
-    ]),
-    trackStyle: React.PropTypes.oneOfType([
-      React.PropTypes.array,
-      React.PropTypes.object
-    ])
+    styles: React.PropTypes.object,
```
tumentumurchudur: I think this should be in singular. We pass it as `this.props.style`

src/components/ToggleSwitch.js
```diff
@@ -1,89 +1,59 @@
...
-      React.PropTypes.object
-    ]),
-    defaultPosition: React.PropTypes.oneOf(['left', 'right']),
-    inactiveColor: React.PropTypes.string,
-    labelStyle: React.PropTypes.oneOfType([
-      React.PropTypes.array,
-      React.PropTypes.object
-    ]),
+    checked: React.PropTypes.bool,
+    checkedByDefault: React.PropTypes.bool,
```
tumentumurchudur: What is the difference between `checked` and `checkedByDefault`? Can we have just `checked` and then have its default value be the default value in `getInitialState`?

src/components/ToggleSwitch.js
```diff
@@ -1,89 +1,59 @@
...
+    falseIcon: React.PropTypes.element,
```
tumentumurchudur: Can this be `leftIcon` or `rightIcon` to make it consistent with `leftLabel` and `rightLabel`?

src/components/ToggleSwitch.js
tumentumurchudur: I would do
```javascript
 this.props.onToggle(!this.state.checked, event);
 this.setState({
   checked: !this.state.checked
 });
```

src/components/ToggleSwitch.js
phantomxc: I actually started with that but changed with the thought that it's more obvious which state that icon represents. I wanted to do it with the labels too but it didn't work out quite the same as you can't really position the labels. The icons can be passed in as their own element and be positioned independently.

src/components/ToggleSwitch.js
phantomxc: really it's just to make it obvious that it's a representing a view preference instead of your data state.

tumentumurchudur: yeah, it seems redundant to have `checkedByDefault` to determine the default value of `checked`. I would just use `checked` and set its default value in `getInitialState` function like I mentioned above.

src/components/ToggleSwitch.js
```diff
@@ -93,73 +63,103 @@ const ToggleSwitch = React.createClass({
...
+  _renderIcons (styles) {
+    if (!this.props.showIcons) {
+      return null;
+    }
+    const trueIcon = this.props.trueIcon ? this.props.trueIcon : <Icon className='true-icon' style={Object.assign({}, styles.icon, styles.trueIcon)} type='check-skinny' />;
+    const falseIcon = this.props.falseIcon ? this.props.falseIcon : <Icon className='false-icon' style={Object.assign({}, styles.icon, styles.falseIcon)} type='close-skinny' />;
+
+    return (
+      <span>
+        {trueIcon} {falseIcon}
```
tumentumurchudur: How do you feel about changing it like this?
```javascript
{this.props.trueIcon ? (
   <Icon 
     style={Object.assign({}, styles.icon, styles.trueIcon)} 
     type='check-skinny' 
  />) : null}

{this.props.falseIcon ? (
   <Icon 
     style={Object.assign({}, styles.icon, styles.falseIcon)} 
     type='close-skinny' 
  />) : null}

```

src/components/ToggleSwitch.js
tumentumurchudur: 👍

src/components/ToggleSwitch.js
phantomxc: This doesn't accomplish the same thing. `this.props.trueIcon` is actually a react element so we render the icon if they provide it otherwise we default to our own. If you don't want icons, you pass in false to the showIcons prop.

tumentumurchudur: Yes, you are right. 👍  How about this?
```javascript
const trueIcon = this.props.trueIcon || <Icon className='true-icon' style={Object.assign({}, styles.icon, styles.trueIcon)} type='check-skinny' />
const falseIcon = this.props.falseIcon || <Icon className='false-icon' style={Object.assign({}, styles.icon, styles.falseIcon)} type='close-skinny' />;
```

src/components/ToggleSwitch.js
phantomxc: Any specific reason why?

tumentumurchudur: Yes, the way you had it works but we don't include functions inside `setState` in our internal codebase. I think it is simpler and easier to follow. @mxenabled/frontend-engineers any thoughts?

phantomxc: Being that onToggle is a callback provided and setState 

> does not immediately mutate this.state but creates a pending state transition.

It felt fitting to use the callback functionality provided by setState. https://facebook.github.io/react/docs/component-api.html#setstate

tumentumurchudur: Sounds good. 👍

src/components/ToggleSwitch.js
```diff
@@ -93,73 +60,103 @@ const ToggleSwitch = React.createClass({
...
+      </span>
+    );
   },
 
   render () {
-    const styles = {
-      activeLabel: {
-        color: this.props.activeColor
-      },
+    const styles = Object.assign({}, {
```
cerinman: We should move the styles object to it's own styles function like we do internally so we don't have to pass it around as an argument.

src/components/ToggleSwitch.js
cerinman: I'm cool with using the callback on setState.  I think it helps clarify that setState is an async operation.  It may not always be required given that you can pass values via argument to another function but I like the idea of setting state and then letting other functions pull from state.  It just seems to be a better pattern.

tumentumurchudur: I'm starting to like that pattern also. It is definitely worth using in certain cases.

src/components/ToggleSwitch.js
tumentumurchudur: I meant the name of the prop. Instead of naming it `falseIcon`, we could name it `leftIcon` to make it more consistent with the `leftLabel` and `rightLabel` props.

src/components/ToggleSwitch.js
```diff
@@ -1,89 +1,59 @@
...
+      showLabels: false,
+      showIcons: true,
+      styles: {}
     };
   },
 
   getInitialState () {
+    let checked = false;
+
+    if ('checked' in this.props) {
```
tumentumurchudur: This should go away once you set `checked` to `false` in `getDefaultProps`.

src/components/ToggleSwitch.js
tumentumurchudur: Agreed.

phantomxc:
This should be good for another go

cerinman:
:shipit:

cerinman:
@phantomxc this has been shipped for a while.  Has something changed or can we merge this in and release?

phantomxc:
@cerinman I just don't check this email very often...

PR #313, opened 2016-05-19
Prevent onClick for disabled buttons

### LINKED ISSUE #312: Make disabled button truly disabled
When setting the button type to disabled it currently only changes the styles. Let's disable onClick events for it as well.

iheartrachie:
Closes #312.

This PR prevents onClick events from firing on buttons if the type is set to `disabled`. I don't know if it's kosher to pass `onClick` like this but it worked and seems to be intuitive when using the button component.


cerinman:
:shipit:

PR #315, opened 2016-05-23
DonutChart: Return Event on click and name transitions

jmophoto:
This adds the click event to the on click so it can be passed back to the parent component. This allows for an `e.stopPropogation` to be called.

Also names the different D3 transitions so they are not interrupted when more than one is called.

Before:
donut_before

After:
donut_after

### DIFF
--- a/src/components/DonutChart.js
+++ b/src/components/DonutChart.js
@@ -41,7 +41,7 @@ const DonutChart = React.createClass({
       activeIndex: -1,
       activeOffset: 0,
       animateOnHover: false,
-      animationDuration: 5000,
+      animationDuration: 500,
       arcWidth: 10,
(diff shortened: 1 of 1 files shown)


cerinman:
:shipit:

src/components/DonutChart.js
```diff
@@ -41,7 +41,7 @@ const DonutChart = React.createClass({
       activeIndex: -1,
       activeOffset: 0,
       animateOnHover: false,
-      animationDuration: 500,
+      animationDuration: 5000,
```
cerinman: Did you change this for testing and forget to put it back?

jmophoto: Yes, I did. Pushing up a fix to revert that back.

cerinman:
:shipit: again.  yay for github latency issues.

PR #316, opened 2016-05-23
Fix fontFamily for Drawer header

cerinman:
When I updated the Drawer component last, I accidentally used a bad key name for getting the regular font for the Drawer component header. I used `NORMAL` but it should have been `REGULAR`.  

See `Style` constants file for reference here. https://github.com/mxenabled/mx-react-components/blob/master/src/constants/Style.js#L41


paniclater:
:shipit:

PR #317, opened 2016-05-24
Drawer header color change

derek-boman:
Change default drawer header from Porcelain to White to prevent need for local overrides.


cerinman:
:shipit:

PR #318, opened 2016-05-24
Modal Height

iheartrachie:
We are seeing issues on internal projects where the height of the modal is exceeding the viewport and wasn't scrollable. I added some flex-box stuff the handle the modal height sizing.

## Before
9zhun7jwtn

## After
pthqeyoist

### DIFF
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -251,7 +251,6 @@ const styles = {
     position: 'relative',
     maxHeight: 'calc(100% - 140px)',
     overflow: 'auto'
-    //flex: '1 2 100%'
   },
   footer: {
     backgroundColor: StyleConstants.Colors.PORCELAIN,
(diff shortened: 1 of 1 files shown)


jmophoto:
This looks good, but since it touches so many places, I would suggest testing locally before we move ahead.

cerinman:
👍  from me as well but would be nice to have a confirmation around testing internally as Jason suggested to be safe.  We use the modal a lot.

iheartrachie:
It's not work on internal apps and I suspect because of conflicts with flexbox. I'll keep poking around and will let ya'll know how this will be resolved.

iheartrachie:
This doesn't fix the issue we're seeing internally but it doesn't break anything either. I'm completely stumped on what's going on internally because even the same modal content as demo (just images) isn't scrollable. There's obviously some conflicting CSS but I can't pinpoint it. If anyone else would like to take a stab at this, feel free. In the meantime, I'm going to close this PR.

PR #319, opened 2016-05-25
DisplayInput: Adds style prop for label

tumentumurchudur:
This PR allows users of the `DisplayInput` component to pass a `style` object for the label. It also moves the status text to the center of the input field.

image


redperk:
:shipit:

PR #320, opened 2016-05-27
Barchart: Adds the ability to set color of clicked bar 

tumentumurchudur:
This PR allows users of the `BarChart` component to be able to set the color of the clicked bar by passing a prop `colorWhenClicked`.

ezgif com-video-to-gif 1


src/components/BarChart.js
```diff
@@ -175,12 +207,14 @@ const BarChart = React.createClass({
       return (
         <Rect
           animateOnHover={this.props.animateOnHover}
+          clickedRect={this.state.clickedRect}
           color={this.props.color}
+          colorWhenClicked={this.props.colorWhenClicked}
           height={height}
           hoverColor={this.props.hoverColor}
           key={index}
```
redperk: Is there something unique we can change this to?

tumentumurchudur: Yeah, let me address that.

tumentumurchudur:
@redperk Your concern with `index` being used for key is addressed. Ready for another review.

redperk:
:shipit:

PR #322, opened 2016-05-31
Add icon option to Select

jmophoto:
This adds functionality to display an icon on a select item if it is passed in as part of the option object.

Also takes out style arrays and moves the Object.assign to the style method.

screen shot 2016-05-31 at 9 23 45 am


iheartrachie:
:shipit:

jmophoto:
Made some changes and cleaned up, so this will need another review, please @iheartrachie.

iheartrachie:
Double :shipit:

PR #323, opened 2016-05-31
Add Radio Button Component

jmophoto:
This adds a radio button component.

Use:
```<RadioButton checked={true} onClick={this._handleClick}>Label</RadioButton>```

Props:
`activeStyle`: Object to override default styling of the active disc in the button
`checked`: Boolean to hide/show active disc in button
`onClick`: Callback when component is clicked
`style`: Object to override default styling of button

The component also accepts children and will display them in a div centered on the button. This can be used for labels, etc.

Default styling:
screen shot 2016-05-31 at 3 12 49 pm

Style overriding to make it big and ugly:
screen shot 2016-05-31 at 3 17 10 pm

### DIFF
--- a/src/components/RadioButton.js
+++ b/src/components/RadioButton.js
@@ -7,12 +7,14 @@ const RadioButton = React.createClass({
     activeButtonStyle: React.PropTypes.object,
     buttonStyle: React.PropTypes.object,
     checked: React.PropTypes.bool,
+    color: React.PropTypes.string,
     onClick: React.PropTypes.func,
     style: React.PropTypes.object
   },
 
   getDefaultProps () {
     return {
+      color: StyleConstants.Colors.PRIMARY,
       onClick: () => {}
     };
   },
@@ -52,7 +54,7 @@ const RadioButton = React.createClass({
         width: '60%',
         height: '60%',
         borderRadius: '100%',
-        backgroundColor: StyleConstants.Colors.PRIMARY
+        backgroundColor: this.props.color
       }, this.props.activeButtonStyle)
     };
   }
--- a/src/components/RadioButton.js
+++ b/src/components/RadioButton.js
@@ -1,0 +1,61 @@
+const React = require('react');
+
+const StyleConstants = require('../constants/Style');
+
+const RadioButton = React.createClass({
+  propTypes: {
+    activeStyle: React.PropTypes.object,
+    checked: React.PropTypes.bool,
+    onClick: React.PropTypes.func,
+    style: React.PropTypes.object
+  },
+
+  getDefaultProps () {
+    return {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -1,6 +1,6 @@
 {
   "name": "mx-react-components",
-  "version": "3.1.33",
+  "version": "3.1.34",
```
cerinman: We usually don't include version bumps in the PR.  Was there a reason you added it here?  This would cause problems if someone deployed before you and they bumped the version.

jmophoto: This was actually supposed to go up with the last release and didn't. I will remove it.

src/components/RadioButton.js
```diff
@@ -0,0 +1,61 @@
...
+const StyleConstants = require('../constants/Style');
+
+const RadioButton = React.createClass({
+  propTypes: {
+    activeButtonStyle: React.PropTypes.object,
+    buttonStyle: React.PropTypes.object,
+    checked: React.PropTypes.bool,
+    onClick: React.PropTypes.func,
+    style: React.PropTypes.object
+  },
```
iheartrachie: Can we add `brandColor` here and set `PRIMARY` as the default?

derek-boman:
Looks good. Thanks for adding this.

src/components/RadioButton.js
iheartrachie: I just noticed `activeButtonStyle` that will override the active color but we're using `brandColor` elsewhere. Not sure what's preferred.

derek-boman: ☝️ probably a good idea. Widgets will probably need this at some point.

jmophoto: Excellent. Added a color prop. Basically a shortcut to the activeStyle.color which would be the piece that is overridden most often. Thanks @iheartrachie.

src/components/RadioButton.js
```diff
@@ -7,12 +7,14 @@ const RadioButton = React.createClass({
     activeButtonStyle: React.PropTypes.object,
     buttonStyle: React.PropTypes.object,
     checked: React.PropTypes.bool,
+    color: React.PropTypes.string,
```
iheartrachie: We're using `brandColor` as the prop name elsewhere. Let's do that here too for the sake of consistency.

jmophoto: We use `brandColor` internally, but since this is a public component, I would leave it as color for greater clarity.

iheartrachie: Oh you're right it looks like we're using `color` or `primaryColor`. I'll keep an eye on that in the future and switch `primaryColor` to `color` when I see it.

iheartrachie:
:shipit:

PR #324, opened 2016-06-01
SimpleSelect Component

### LINKED ISSUE #185: MX components / Select menus
image

We have this standard Select menu, which is good, i believe we'll need another type of Simple dropdown like the settings icon in top right of Insight/Target. The trigger for this menu would vary, sometimes as a Base Button, sometimes as an icon, etc. The menu itself would follow an established pattern (although there are probably use cases for both right and left aligned menus).

LMK any thoughts here. I have a few places to implement this right away.

screen shot 2016-02-01 at 10 53 17 am

Thoughts from @bsbeeks: I think we should adjust the existing Select element to not have a pre formatted trigger. We should let the user decide what the trigger is and how it looks. If we have a specific style that we reuse in a project, then we can create a project specific component that wraps the Select and defines a shared trigger style.

iheartrachie:
Closes #185

This adds the SimpleSelect component. This is just the dropdown and scrim and not the icon/button/whatever that triggers the menu.

## Example
```
<div style={{ margin: '0 auto', width: 177 }}>
  <Button
    icon='gear'
    onClick={this._handleSimpleSelectClick}
    type='base'
  >
    This is a menu Button
  </Button>
  <SimpleSelect
    items={[
      { text: 'Menu Item 1' },
      { text: 'Menu Item 2' },
      { text: 'Menu Item 3' }
    ]}
    onScrimClick={this._handleSimpleSelectClick}
    showItems={this.state.showSimpleSelectItems}
  />
</div>
```
screen shot 2016-06-01 at 11 11 45 am

## Props
`iconSize` _number_ **optional**
Default: 20
The icon size used for icons in menu items if used

`iconStyles` _object_ **optional**
Styles for the icons in menu items if used

`items` _array_ **required**
An array of objects that specify `icon`, `text`, and/or `onClick` of the item.

`itemStyles` _object_ **optional**
Styles of the item in the list

`menuStyles` _object_ **optional**
Styles for the menu

`onScrimClick` _function_ **optional**
`onClick` handler for the menu scrim

`showMenu` _boolean_ **optional**
Default: false
Boolean value that dictates whether or not the menu is shown

`style` _object_ **optional**
Styles for the container around the component

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -866,7 +866,7 @@ const Demo = React.createClass({
                 { text: 'Menu Item 3' }
               ]}
               onScrimClick={this._handleSimpleSelectClick}
-              showItems={this.state.showSimpleSelectItems}
+              showMenu={this.state.showSimpleSelectItems}
             />
           </div>
         </div>
--- a/demo/app.js
+++ b/demo/app.js
@@ -856,7 +856,9 @@ const Demo = React.createClass({
               icon='gear'
               onClick={this._handleSimpleSelectClick}
               type='base'
-            >This is a menu Button</Button>
+            >
+              This is a menu Button
+            </Button>
             <SimpleSelect
               items={[
                 { text: 'Menu Item 1' },
--- a/demo/app.js
+++ b/demo/app.js
@@ -859,10 +859,14 @@ const Demo = React.createClass({
           <SimpleSelect
             align='right'
             items={[
-              { text: 'Menu Item 1' },
+              {
+                icon: 'add',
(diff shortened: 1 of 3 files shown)


jmophoto:
I like this with one question. What are your thoughts about leaving the open/closed up to the parent. This is what we do now with modals and the loader.

iheartrachie:
If that's how we're doing it on the modal, we should do the same here. I adjusted it to match so it's ready for another look-see.

jmophoto:
🍪  Looks good! :shipit:

PR #325, opened 2016-06-01
Add Icons constant

jmophoto:
This moves the Icons constant out of the app.js and into a constant. I am working on changes to the gh-pages IconDocs and realized that that page is out of sync with what we actually have in the component. This would centralize so the docs would stay current with the code.


iheartrachie:
❤️  👯  🍪  🍆  :shipit:

PR #326, opened 2016-06-02
Tooltip Component

jmophoto:
This adds a tooltip component with the following props:

- `children` (node): Text or JSX to be displayed inside the tooltip
- `icon` (string): Icon to be displayed to indicate a tooltip. Defaults to `info`
- `iconSize` (number): Size of the icon to be displayed: Defaults to `20`
- `placement` (string): Used to quickly position tooltip. Must be `top`, `right`, `bottom` or `left`
- `style` (object): Used to override component styles. Can pass `fill` in to set the icon color
- `tooltipStyle` (object): Used to override the tooltip styling. Can be used for custom positioning.

Top:
screen shot 2016-06-02 at 5 12 10 pm

Right:
screen shot 2016-06-02 at 5 11 36 pm

Bottom:
screen shot 2016-06-02 at 5 13 08 pm

Left:
screen shot 2016-06-02 at 5 11 54 pm

@derek-boman

### DIFF
--- a/src/components/Tooltip.js
+++ b/src/components/Tooltip.js
@@ -1,0 +1,129 @@
+const React = require('react');
+
+const Icon = require('./Icon');
+
+const StyleConstants = require('../constants/Style');
+
+const Tooltip = React.createClass({
+  propTypes: {
+    icon: React.PropTypes.string,
+    iconSize: React.PropTypes.number,
+    placement: React.PropTypes.oneOf(['left', 'top', 'right', 'bottom']),
+    style: React.PropTypes.object,
+    tooltipStyle: React.PropTypes.object
+  },
+
+  getDefaultProps () {
+    return {
+      icon: 'info',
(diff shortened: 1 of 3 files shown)


src/components/Tooltip.js
```diff
@@ -0,0 +1,130 @@
...
+      case 'bottom':
+        return {
+          top: offSet,
+          left: '50%',
+          marginLeft: -(this.props.iconSize / 2 + width / 2)
+        };
+      default:
+        return null;
+    }
+  },
```
iheartrachie: This makes me happy. 🍪

iheartrachie:
I like this! Is there a way to make it work with text instead of just an icon?

jmophoto:
I hadn't thought of that use-case, but it would be added later. For now, most of what we need is icon-based.

iheartrachie:
:shipit:

PR #327, opened 2016-06-02
Defualt cursor for disabled Buttons

iheartrachie:
This would set the cursor to be default if the button is disabled. Since we're disabling `onClick` events, we don't want to confuse the user by having a pointer of hover for disabled buttons.


jmophoto:
🍪  :shipit:

PR #328, opened 2016-06-03
Change fileupload icon styles from color to fill

jmophoto:
This updates the icons in the FileUpload to display correctly.


iheartrachie:
:shipit:

PR #329, opened 2016-06-03
Moved Icons to AppConstants

jmophoto:
My idea for the Icons constant that could be shared with the gh-pages wasn't quite working. This adds an AppConstants component and renames the Icons.js to App.js in the Constants directory.


iheartrachie:
:shipit:

PR #331, opened 2016-06-07
Add Row and Column components for responsive grid

jmophoto:
This is a rough cut for a responsive grid system.

- Add Row component to wrap columns
- Add Column component to wrap content

In the column there are the following props
- breakLg
- breakMd
- breakSm
- colLg
- colMd
- colSm
- columnCount
- offLg
- offMd
- offLg
- ordSm
- ordMd
- ordLr
- style

`break*` sets the breakpoints for the screen width. Can be different for each column. 

`col*` sets the columns for the designated screen width. This is based on a 12-column layout, so `col-sm={12}` would be full width at a small screen width and `col-lg={6}` would be 50% at large screen width.

`columnCount` set the number of columns to base calculations. defaults to 12

`off*` sets the offset for the designated screen width. Same column layout as above. 

`ord*` sets the order of the display. Use negative numbers to float to the top.

`style` overrides/adds to styles for each column. Can be used to change gutter width (left/right padding)

~~Right now, the breakpoints are just temp numbers I threw in there, but they could also be passed.~~
Set to 750 for medium and 1200 for large

I have tested and this is working in Chrome, Firefox, Safari and Edge. 

The top row has an offset of 3 and width of 9 for large, offset of 4 and width of 6 for medium and none for small.

The bottom row is a nested set with the parent being 6, 9 and 12 for large, medium and small and the children being 6 for all.

Default is 12 so if the param is not included, it will display full width.

ezgif com-video-to-gif 1


jmophoto:
Thought to add based on Bootstrap
- [ ] Column ordering
- [ ] Column count variable
- [ ] Gutter width

jmophoto:
CamelCased props and added an order.

Here, divs 6 and 7 are moved to the top, with 7 being first and 6 being second on medium.

screen shot 2016-06-08 at 11 23 18 am
screen shot 2016-06-08 at 11 23 27 am

src/components/Column.js
```diff
@@ -0,0 +1,94 @@
...
+  styles () {
+    return {
+      component: Object.assign({}, {
+        boxSizing: 'border-box',
+        flexBasis: (this.props['col' + this.state.screenWidth] / this.props.columnCount * 100) + '%',
+        flexGrow: 0,
+        flexShrink: 0,
+        marginLeft: (this.props['off' + this.state.screenWidth] / this.props.columnCount * 100) + '%',
+        order: this.props['ord' + this.state.screenWidth],
+        padding: '0 10px'
```
iheartrachie: why the padding?

jmophoto: That creates the gutters between the columns.

src/components/Column.js
```diff
@@ -0,0 +1,94 @@
...
+    colLg: React.PropTypes.number,
+    colMd: React.PropTypes.number,
+    colSm: React.PropTypes.number,
+    columnCount: React.PropTypes.number,
+    offLg: React.PropTypes.number,
+    offMd: React.PropTypes.number,
+    offSm: React.PropTypes.number,
+    ordLg: React.PropTypes.number,
+    ordMd: React.PropTypes.number,
+    ordSm: React.PropTypes.number,
```
cerinman: Was there a reason for abbreviating all the props?  As some one new to this component, this might be hard to follow.

jmophoto: Not a good one, no. My first inclination was to have full prop names ie:`columnSmall`. Wondered if that would be a lot for the parent component, but I think I agree about more discoverable over less typing.

I will change that are re-push.

src/components/Column.js
```diff
@@ -0,0 +1,94 @@
...
+    let screenWidth = 'Sm';
+
+    if (width >= this.props.breakLg) {
+      screenWidth = 'Lg';
+    } else if (width >= this.props.breakMd) {
+      screenWidth = 'Md';
+    }
+
+    return screenWidth;
+  },
```
cerinman: This is nit picky but can we move this `_getScreenWidth` function up above `_handleWindowResize`.  Not only does it alphabetize it but it also puts it closer to it's other use in `getInitialState`

src/components/Row.js
```diff
@@ -0,0 +1,31 @@
...
+        boxSizing: 'border-box',
+        display: 'flex',
+        flexWrap: 'wrap',
+        margin: '0 -10px'
+      }, this.props.style)
+    };
+  }
+});
+
+
```
cerinman: Looks like you have an extra line break here

src/components/Column.js
```diff
@@ -0,0 +1,94 @@
...
+        flexShrink: 0,
+        marginLeft: (this.props['off' + this.state.screenWidth] / this.props.columnCount * 100) + '%',
+        order: this.props['ord' + this.state.screenWidth],
+        padding: '0 10px'
+      }, this.props.style)
+    };
+  }
+});
+
+
```
cerinman: Extra line break

cerinman:
Should we put the row and column components in a sub grid folder within the components folder?  Similar to what we did with the `D3` sub folder.  If the only component that will ever user them is the `grid` component then I would argue that they should be in a sub folder, but if we plan to expose and let people use the `Column` and `Row` components stand alone then I'm fine with leaving them as is.

jmophoto:
Agree about the `grid` directory. I think they would usually be used in tandem, so that would make sense. I have changed that and pushed.

cerinman:
:shipit:

PR #333, opened 2016-06-09
Add animateLeftDistance prop to Drawer component

cerinman:
### Issue

Nesting Drawer components has an issue where children Drawer components don't animate to the correct left position because of their parent container.

### Solution

This PR adds a new prop `animateLeftDistance` - (type: number) - (default: null) - Read as a percentage, this prop is the distance from the left of the containing div that the drawer will animate to when opened.

If supplied, the component will used the prop value over calculating a left distance on the fly.

I messed around with positioning trying to avoid adding this prop but this seems to be the only universal fix that prevents weird drawer positioning.

#### Before
screen shot 2016-06-09 at 2 57 54 pm

#### After
screen shot 2016-06-09 at 2 58 32 pm

### DIFF
--- a/src/components/Drawer.js
+++ b/src/components/Drawer.js
@@ -1,3 +1,4 @@
+const _isNumber = require('lodash/isNumber');
 const Radium = require('radium');
 const React = require('react');
 const Velocity = require('velocity-animate');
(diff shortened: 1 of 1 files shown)


jmophoto:
:shipit:

PR #334, opened 2016-06-10
Add ability to hide scrim for drawer via showScrim prop

cerinman:
### Issue

Now that we can nest Drawer components, we don't want the nested Drawers to show their scrim.

### Solution

This adds another new prop to the Drawer component

`showScrim` - Type(`Bool`) - Default(`true`) - If set to false, sets the color of the scrim to transparent so that you can still see the content below.

#### Before
screen shot 2016-06-10 at 9 53 32 am

#### After
screen shot 2016-06-10 at 9 53 53 am


jmophoto:
:shipit:

PR #335, opened 2016-06-13
Modal update

cerinman:
This PR does the following:

- Moves the styles object into a styles function with in the component as we've done else where.
- Adds a new prop `showScrim` - (type: `bool`) - (default: `true`) - When set to false, makes the background color of the scrim transaparent.
- Adds a new prop `showCloseIcon` - (type: `bool`) - (default: `true`) - When set to false, hides the close icon in the top right corner of the modal. Clicking on the scrim still closes the Modal even if the scrim background color is `transparent`.
- Fixes the the props validation for `footerStyle`.  It was set to `oneOfType` array or object but we're not using Radium here so an array would cause issues.  It's now just validated as an object.

These changes were made so that I can match some new designs from product internally but also caused the need for the minor cleanup that was involved.

#### With scrim and close icon
screen shot 2016-06-13 at 2 31 44 pm

#### With OUT scrim and close icon
screen shot 2016-06-13 at 2 32 09 pm


paniclater:
:shipit:

redperk:
Looks good to me, but @paniclater beat me to it. 👍

PR #336, opened 2016-06-13
New Calendar component

cerinman:
This PR extracts the calendar logic from the `DatePicker` component to it's own `Calendar` component.  I've done this because we now have duplicate code for the calendar in 5 locations between the open source and internally.  I don't want to maintain all that so here we go.

This PR does the following:

- Extracts the calendar code to it's own component as already noted above.
- Refactors the `DatePicker` component to use the new `Calendar` component.
- Sets up the Demo app to use the new `Calendar` component.

I'll update the `DatePickerFullScreen` and new `DateRangePicker` components in separate PRs after this one gets merged.  I didn't do that here because the `DatePickerFullScreen` component is way behind the curve for being up to date and the new `DateRangePicker` needs to be able to select ranges so I'll have to read through it's code a bit more before I implement the `Calendar` component there.

#### Screen

screen shot 2016-06-13 at 5 10 35 pm
screen shot 2016-06-13 at 5 11 04 pm

### DIFF
--- a/src/components/Calendar.js
+++ b/src/components/Calendar.js
@@ -6,15 +6,11 @@ const Icon = require('./Icon');
 
 const StyleConstants = require('../constants/Style');
 
-const DatePicker = React.createClass({
+const Calendar = React.createClass({
   propTypes: {
-    closeOnDateSelect: React.PropTypes.bool,
-    defaultDate: React.PropTypes.number,
-    format: React.PropTypes.string,
     locale: React.PropTypes.string,
     minimumDate: React.PropTypes.number,
     onDateSelect: React.PropTypes.func,
-    placeholderText: React.PropTypes.string,
     primaryColor: React.PropTypes.string,
     selectedDate: React.PropTypes.number,
     style: React.PropTypes.object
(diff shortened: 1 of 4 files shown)


jmophoto:
🍪  for this. Will use it in the date time picker! :shipit:

src/components/Calendar.js
```diff
@@ -0,0 +1,225 @@
...
+  },
+
+  getInitialState () {
+    return {
+      currentDate: this.props.selectedDate || moment().unix()
+    };
+  },
+
+  componentWillReceiveProps (newProps) {
+    if (newProps.selectedDate && newProps.selectedDate !== this.props.selectedDate) {
```
redperk: I'm confused on why the same thing is written twice. Is this correct or is it supposed to be `if (newProps && newProps.selectedDate) .....`?

iheartrachie:
I'm all for DRYing up code as much as possible. This is already merged and this should work without issue for `DatePickerFullScreen` but I'm not sure I see how it can be cohesive with `DateRangePicker`. It has slightly different props but vastly different CSS and logic in the component.

src/components/Calendar.js
redperk: or wait, is it checking first to make sure it's not undefined and if it's not then it's checking to see if the newProps.selectedDate is not equal to this.props.selectedDate?

cerinman: Copied the code over from the DatePicker component so I can't say for sure but it appears that whoever wrote it was trying to protect against undefined errors.

cerinman:
Not sure either @iheartrachie but I'm going to try and make it work somehow so we don't have to maintain multiple calendars.  I'll keep you posted.

redperk:
:shipit:

PR #338, opened 2016-06-14
Add DateTimePicker

jmophoto:
This adds a DateTimePicker to the components.

- Accepts and returns a unix timestamp
- Is daylight savings time aware based on the selected date
- Uses the calendar component (thx @cerinman!)
- Adds `event` return to calendar so the parent can use that for `e.stopPropagation()`

### Props ###
`calendarStyle`: object. Style object to be passed to the calendar component
`closeOnDateSelect`: boolean, default true. Closes the calendar on select
`dateFormat`: string, default 'MMM D, YYYY'. Sets the display format for the date
`dateIcon`: string, default 'calendar'. Sets the icon for the date select box
`datePlaceholder`: string, default 'Select a Date'. Placeholder for the date select box
`locale`: string, default 'en'. Locale to display the date in
`minimumDate`: number. unix timestamp to limit the selectable calendar days
`onDateSelect`: function. Callback when date or time is selected
`primaryColor`: string, default PRIMARY. Color of the calendar days, active selects, icons, etc.
`selectedDate`: number, Unix timestamp for the selected day
`showIcons`: boolean, default true. Used to hide/show icons
`styles`: object. Style object to override component styles
`timeFormat`: string, default 'LT'. Sets the display format for the time
`timeIcon`: string, default 'clock'. Sets the icon for the time select box
`timePlaceholder`: string, default 'Select a Time'. Placeholder for the time select box
`timezone`: string, defaults to browser timezone. Sets the timezone for displaying time
`timzoneFormat`: 'abbr' or 'name'. Sets the display format of the timezone or if blank, hides it
`timezoneNames`: object, defaults to US time zone names. Used to display the names of the timezone

Initial with timezone name
screen shot 2016-06-14 at 4 56 44 pm

Initial with timezone abbreviation
screen shot 2016-06-14 at 4 57 55 pm

Selected day in Daylight Savings Time
screen shot 2016-06-14 at 4 57 04 pm

Selected day in Standard Time
screen shot 2016-06-14 at 4 57 16 pm

Without Icons
screen shot 2016-06-14 at 4 58 34 pm

With custom date icon
screen shot 2016-06-14 at 4 59 04 pm


src/components/DateTimePicker.js
```diff
@@ -0,0 +1,292 @@
...
+  },
+
+  _handleTimeFocus () {
+    this.setState({
+      editTime: true
+    });
+  },
+
+  _handleTimeSelect (e) {
+    const selectedDate = this.props.selectedDate ? moment.unix(this.props.selectedDate) : moment();
```
iheartrachie: So if I were to focus on the field, then blur, would the selectedDate become today?

jmophoto: Had to test that... and no. Focus and then blur without clicking on a date does not set the date to today.

src/components/DateTimePicker.js
```diff
@@ -0,0 +1,292 @@
...
+    const value = e.target.value.toLowerCase().replace('.', '');
+
+    const time = value.split(':');
+    const pm = value.includes('pm');
+    const hour = pm ? Number(time[0]) + 12 : Number(time[0]);
```
iheartrachie: Does a user manually type in 'pm'?

jmophoto: This is a work-around for Firefox since it doesn't honor the `time` type for an input. Instead, it treats it as a regular text input, so this is a catch to convert it to a time.

src/components/DateTimePicker.js
```diff
@@ -0,0 +1,292 @@
...
+    const styles = this.styles();
+
+    return (
+      <div style={styles.component}>
+        <div
+          onBlur={this._handleDateBlur}
+          onClick={this._handleDateClick}
+          onFocus={this._handleDateFocus}
+          ref='dateSelect'
+          style={Object.assign({}, styles.selectWrapper, this.state.showCalendar ? styles.activeSelectWrapper : null)}
```
iheartrachie: Since we're only changing the color can we check this in the styles?
```
this.state.showCalendar ? this.props.primaryColor : StyleConstants.Colors.FOG
```
Just my personal opinion but it feels cleaner than having an additional style.

src/components/DateTimePicker.js
iheartrachie: However if I focus/blue on time, it populates with a default time and thusly date. Not sure if that's intended behavior.

src/components/DateTimePicker.js
iheartrachie: It looks like the only options are am/pm. I don't know how critical it is right now but we'll probably want to allow for military time in the future.

src/components/DateTimePicker.js
jmophoto: The issue with that is the selectWrapper style is used for the date and the time, so I need a conditional in each one. 

On line 177
`style={Object.assign({}, styles.selectWrapper, this.state.editTime ? styles.activeSelectWrapper : null)}`

src/components/DateTimePicker.js
jmophoto: Yeah... I need a date to set the time, so setting a date by default had to set the time.

src/components/DateTimePicker.js
```diff
@@ -0,0 +1,292 @@
...
+          onBlur={this._handleTimeBlur}
+          onFocus={this._handleTimeFocus}
+          style={Object.assign({}, styles.selectWrapper, this.state.editTime ? styles.activeSelectWrapper : null)}
+          tabIndex={0}
+        >
+          {this.props.showIcons ? (
+            <Icon
+              size={20}
+              style={styles.selectedIcon}
+              type='clock'
```
iheartrachie: Should be `this.props.timeIcon`

jmophoto: Yes. Yes it should. 👍

src/components/DateTimePicker.js
jmophoto: If they enter military time when using Firefox, then it handles it correctly. This issue was entering 1:00 PM. The time split would make it 01 instead of 13. If they enter 13:00 then the split works as expected.

src/components/DateTimePicker.js
iheartrachie: Ohhhh yes I see. Funny because I just put up a PR yesterday doing the exact same thing you did here and I didn't even catch it! Don't mind me!

iheartrachie:
This looks great! I pulled down the branch to play around with it and found it a gem to use. :cookie::cookie::cookie:!!!!!

Any chance I could convince you to bring https://github.com/mxenabled/mx-react-components/pull/339 with you?

:shipit:

PR #339, opened 2016-06-15
Fixes typo to allow for custom color prop

iheartrachie:
Setting a custom primaryColor prop wasn't working when hovering over dates. This fixes the issue. @jmophoto


jmophoto:
:shipit:

PR #340, opened 2016-06-15
Jmo/toggle switch

jmophoto:
I found that putting this into a project with box-sizing set to border box made the track too small. In fixing that, I had some ideas for refactoring.

- Changed the icon props from an element to a string and set defaults
- Changed the style prop to styles
- Took out state since the prop is toggled in the parent, so true/false can be checked with the prop.checked
- Moved _renders into main body with ternary
- Removed 'px' from styles
- Fixed fontFamily and fontSize to use style constants
- Combined falseToggle and trueToggle into toggle style with conditional

Thoughts on this @phantomxc?

### DIFF
--- a/src/components/ToggleSwitch.js
+++ b/src/components/ToggleSwitch.js
@@ -47,11 +47,11 @@ const ToggleSwitch = React.createClass({
           onClick={this._handleToggle}
           style={Object.assign({}, styles.track, styles[this.props.checked + 'Track'])}
         >
-          {this.props.showIcons && this.props.checked ? (
+          {this.props.showIcons ? (
             <Icon className='false-icon' style={Object.assign({}, styles.icon, styles.trueIcon)} type={this.props.trueIcon} />
           ) : null}
           <div className='toggle-switch-toggle' style={styles.toggle}></div>
-          {this.props.showIcons && !this.props.checked ? (
+          {this.props.showIcons ? (
             <Icon className='false-icon' style={Object.assign({}, styles.icon, styles.falseIcon)} type={this.props.falseIcon} />
           ) : null}
         </div>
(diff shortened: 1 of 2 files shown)


src/components/ToggleSwitch.js
```diff
@@ -6,101 +6,69 @@ const Icon = require('./Icon');
...
         <div
           className='toggle-switch-track'
           onClick={this._handleToggle}
-          style={Object.assign({}, styles.track, styles[this.state.checked + 'Track'])}
+          style={Object.assign({}, styles.track, styles[this.props.checked + 'Track'])}
         >
-          <div className='toggle-switch-toggle' style={Object.assign({}, styles.toggle, styles[this.state.checked + 'Toggle'])}></div>
-          {this._renderIcons(styles)}
+          {this.props.showIcons ? (
+            <Icon className='false-icon' style={Object.assign({}, styles.icon, styles.trueIcon)} type={this.props.trueIcon} />
```
phantomxc: The className should be true-icon. I'm also not sure they both need to be wrapped in showIcons conditional as the conditional is the same and there is no element between them.

src/components/ToggleSwitch.js
```diff
@@ -6,101 +6,69 @@ const Icon = require('./Icon');
...
       showLabels: false,
       showIcons: true,
-      style: {}
-    };
-  },
-
-  getInitialState () {
-    return {
-      checked: this.props.checked
+      style: {},
```
phantomxc: styles

jmophoto: I removed the default style prop altogether.

src/components/ToggleSwitch.js
jmophoto: Yep. Them being separated was when I thought I would them as relative instead of absolute. Changed that and the class name.

phantomxc:
:shipit:

PR #341, opened 2016-06-16
Tommy/line chart dots

cerinman:
Adds a new prop to the TimeBasedLineChart for limiting the circles on the line to the beginning, middle, and end of the line.

- `limitLineCircles` - (type: `bool`) - (default: `false`) - If set to true will limit the number of circles drawn on the line to just the beginning, middle, and end.

This was done to match a new design internally.

#### this.props.limitLineCircles === false
screen shot 2016-06-16 at 12 30 01 pm

#### this.props.limitLineCircles === true
screen shot 2016-06-16 at 12 30 22 pm

### DIFF
--- a/src/components/TimeBasedLineChart.js
+++ b/src/components/TimeBasedLineChart.js
@@ -293,19 +293,9 @@ const TimeBasedLineChart = React.createClass({
   // Helper Functions
   _getDataForLineCircles () {
     if (this.props.limitLineCircles) {
-      const data = [];
-
-      this.props.data.forEach((dataPoint, index) => {
-        const shouldIncludeDataPoint = index === 0 ||
-                                       index === this.props.data.length - 1 ||
-                                       dataPoint.x === this.props.breakPointDate;
-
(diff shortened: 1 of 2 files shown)


src/components/TimeBasedLineChart.js
```diff
@@ -289,6 +291,26 @@ const TimeBasedLineChart = React.createClass({
   },
 
   // Helper Functions
+  _getDataForLineCircles () {
+    if (this.props.limitLineCircles) {
+      const data = [];
+
+      this.props.data.forEach((dataPoint, index) => {
```
bsbeeks: You could use `.filter` instead:
```javascript
return this.props.data.filter((dataPoint, index) => {
  return index === 0 || index === this.props.data.length - 1 || dataPoint.x === this.props.breakPointDate;
});
```

cerinman: 👍

cerinman:
ready for another review

bsbeeks:
Nice! :shipit:

PR #342, opened 2016-06-16
Experimental PR To Demonstrate Differences in React Component Testing Options

paniclater:
**This PR:**
 - adds karma configuration for component testing with / without enzyme
 - adds chrome karma launcher for debugging
 - alters the loader component to allow for the demonstration of event testing
 - demonstrates the differences between shallow and full render tests with vanilla react testing tools
 - demonstrates the differences between shallow and full render tests with enzyme

--

**Takeaways:**

--

**Enzyme:**
I think enzyme is a useful tool for dom traversal and it makes dealing with click situation much less verbose. 
In general I think it will make our tests less brittle:
`loaderComponentOutput.props.children.props.children[1].props.onClick`
 
versus:

`wrapper.find('.mx-loader-text').props().onClick`

The first example (especially with any kind of hefty tree of elements will break everytime stuff is reordered. 

With jquery like selectors of enzyme in the second example the test will succeed even if the elements are completely reordered.

--

**Shallow vs. Full Render:**

 - shallow render is faster, prevents children components from altering test results in parent components
 - full render allows you to test state directly. 
 - we should use both and prioritize doing expensive mounting in `beforeAll` blocks where possible.


paniclater:
Not sure how to tell it that `React` is actually being used for the travis test.

jtanner:
👍  for using `enzyme`.  I like how the `'Loader Enzyme'` block of tests is less verbose and easier to read.

paniclater:
@jtanner thanks for bringing it up! I had completely assumed it was a jasmine replacement or that it didn't have jasmine support.

jtanner:
@paniclater can this be closed?

cerinman:
Can this be closed @paniclater?

PR #343, opened 2016-06-20
Columns: add 0 option to column to hide

jmophoto:
I realized we needed an easy way to add/hide columns. This adds a check for a `0` value and if found, will not render the column at that size.

`columnLarge={0}`


redperk:
:shipit:

PR #345, opened 2016-06-20
Add Progress Bar

jmophoto:
Add progress bar component.

#### Props ####
`baseColor`: string, default FOG, color for the full width bar
`height`: number, default 10, height of the bar
`percentage`: number, percentage of width for the progress bar. Shown in percentage value, so `50` for `50%`
`progressColor`: string, default PRIMARY, color of the progress bad
`styles`: object, styles to override either the component or the progress divs

The `height`, `baseColor` and `progressColor` props can also be overridden within the styles prop, but since they would be used most often, I have exposed them as helper props.

Also, I switched from Object.assign on the styles to lodash's `_merge`. This allows us to override/provide a new style without wiping out all of the others.

Default and custom colors and styles
```
<ProgressBar percentage={50} />
<ProgressBar baseColor={Styles.Colors.ASH} color={Styles.Colors.ORANGE} percentage={25} styles={{ component: { width: '75%' } }}/>
```
screen shot 2016-06-20 at 1 10 41 pm

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -22,6 +22,7 @@ const {
   Loader,
   Modal,
   PageIndicator,
+  ProgressBar,
   RadioButton,
   RangeSelector,
   Row,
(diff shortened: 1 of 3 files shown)


cerinman:
Nice and simple!  :shipit:

PR #346, opened 2016-06-20
Column: change to window.innerWidth for column

jmophoto:
Internally, we are using this in an iFrame, which made getting the width by calling `document.documentElement.clientWidth` unhelpful. I change it to `window.innerWidth`. Testing this, it returns the correct value stand-alone and also from the iFrame.


redperk:
:shipit:

PR #347, opened 2016-06-21
Adds constants for category colors

derek-boman:
To be used mostly in Spending widget but also in Transaction Details drawer and probably other places.


paniclater:
There's a eslint problem Travis is angry about.

cerinman:
:shipit:

PR #348, opened 2016-06-21
Add default range options to DateRangePicker

cerinman:
Designs internally call for allowing a set of default ranges to pick from when using the DateRangePicker.  This adds two new props to the `DateRangePicker` to make that happen.

- `defaultRanges` - (type:`array of range objects`) - (default: `see PR details`)
- `showDefaultRanges - (type:`bool`) - (default: `false`) - when set to true will show a list of predefined ranges to select from.  Can be over written by supplying your own array of ranges.

screen shot 2016-06-21 at 10 44 07 am
screen shot 2016-06-21 at 10 44 17 am
screen shot 2016-06-21 at 10 44 38 am
screen shot 2016-06-21 at 10 44 48 am


src/components/DateRangePicker.js
```diff
@@ -256,35 +336,38 @@ const DatePicker = React.createClass({
...
-            <Icon
-              onClick={this._handlePreviousClick}
-              size={20}
-              style={styles.calendayHeaderNav}
-              type='caret-left'
-            />
-            <div>
-              {moment(this.state.currentDate, 'X').format('MMMM YYYY')}
+        <div style={styles.optionsWrapper}>
+          {this.props.showDefaultRanges && this._renderDefaultRanges()}
```
phantomxc: This reads a little weird to me but I _think_ I like it better than throwing in the nulls in the ternary...

src/components/DateRangePicker.js
```diff
@@ -256,35 +336,38 @@ const DatePicker = React.createClass({
...
+                onClick={this._handleNextClick}
+                size={20}
+                style={styles.calendayHeaderNav}
+                type='caret-right'
+              />
+            </div>
+            <div style={styles.calendarWeek}>
+              {['S', 'M', 'T', 'W', 'T', 'F', 'S'].map((day, i) => {
+                return (
+                  <div key={i} style={styles.calendarWeekDay}>
```
phantomxc: wouldn't it be better to use `day` for the key?

phantomxc:
:shipit: I think this is really nifty.

PR #350, opened 2016-06-21
DateTimePicker: validate hours and minutes

jmophoto:
Browsers that don't support the `time` input default to a standard text field. This PR adds validation to check that the numbers for hour and minutes are valid. 

Valid
- 12:01
- 13:59
- 12:01 AM

Invalid
- 24:61
- 13:01 PM


cerinman:
:shipit:

PR #352, opened 2016-06-22
Date range picker tweaks

cerinman:
### Issue

After using the `DateRangePicker` internally, we decided that it's current functionality didn't match intended behavior exactly.

The picker was working fine but in order to deselect a range you had the select the start and end date from the calendar.  This caused confusion and made it difficult to work with if a range crossed multiple months.

### Solution

The picker now clears the current range if the user clicks a date in the calendar and a complete range was already in place.  This does away with the confusion and prevents us from having to had the ability to clear the range via external means such as a button.

I also fixed a duplicate key warning that was caused by my last set of changes where we set the key for the days label header to use the day rather than an index.  Tuesday and Thursday are both represented by a `T` so we were getting a duplicate key warning.  We now use the index and day for the key.

#### New Behavior
daterangepicker

### DIFF
--- a/src/components/DateRangePicker.js
+++ b/src/components/DateRangePicker.js
@@ -83,76 +83,31 @@ const DateRangePicker = React.createClass({
     }
   },
 
+  _endDateIsBeforeStartDate (startDate, endDate) {
+    return moment.unix(endDate).isBefore(moment.unix(startDate));
+  },
+
   _handleDateSelect (date) {
-    const selectedStartDate = this.props.selectedStartDate;
-    const selectedEndDate = this.props.selectedEndDate;
-    let updatedDates = [null, null];
+    let endDate;
+    let startDate;
+    const existingRangeComplete = this.props.selectedStartDate && this.props.selectedEndDate;
+    const existingRangeEmpty = !this.props.selectedStartDate && !this.props.selectedEndDate;
 
-    updatedDates = this._deselectDate(date, selectedStartDate, selectedEndDate);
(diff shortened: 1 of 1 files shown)


jmophoto:
This looks solid to me. 👍

src/components/DateRangePicker.js
```diff
@@ -83,76 +83,31 @@ const DateRangePicker = React.createClass({
...
+    const existingRangeComplete = this.props.selectedStartDate && this.props.selectedEndDate;
+    const existingRangeEmpty = !this.props.selectedStartDate && !this.props.selectedEndDate;
+
+    if (existingRangeComplete || existingRangeEmpty) {
+      startDate = date;
+      endDate = null;
+    } else {
+      startDate = this.props.selectedStartDate;
+      endDate = date;
     }
```
iheartrachie: This doesn't allow a user to expand the date range either. Are we cutting that out too?

cerinman: Yes

iheartrachie:
This is small and I haven't dug into where but I noticed when selecting a range there are rounded corners on the start date, but not end date. The styling on "Today" should probably be rounded all round or not at all. Do we want to select just today? That's not really a range.

The rest of the stuff on the range selector looks great though, I like it! 🍪

iheartrachie:
Since the stuff I noticed isn't related to this PR so go ahead and :shipit: We'll fix the CSS stuff later.

PR #353, opened 2016-06-23
Column: Change from individual props to objects

jmophoto:
This changes the breakpoint*, column*, offset* and order* to objects. 

Also renames columns to span


jtanner:
I like the changes to the api 👍

cerinman:
NIce!  :shipit:

PR #355, opened 2016-06-23
DateTimePicker: Fix error with midnight and noon

jmophoto:
When converting an am/pm date in non-chrome browsers, there was a fault in my logic when it came to midnight and noon.

- Sets `12` hour with am to `0`
- Sets non-`12` hours with pm to the value + `12`


lukeschunk:
:shipit:

PR #356, opened 2016-06-24
luke/deprecated icons and added line-of-credit

lukeschunk:
This fixes the `line-of-credit` Icon, and deprecates the `key` and `bike` which Derek told me are no longer needed. 

Are we currently using Key or Bike anywhere that we should be concerned about? I'm going to do a quick check but let me know if you know of somewhere.


paniclater:
:shipit:

cerinman:
Does it hurt anything to leave the `key` and `bike` icons in place?  If we're absolutely set on removing them for a valid reason then maybe we should first add a console warning to the `Icon` component saying they will be removed in the next major release rather than just yanking them out?

paniclater:
Good thinking. Probably better to only remove icons if they are faulty or outdated.

lukeschunk:
i'm pretty sure that the `bike` icon is now being used for `kids` - should we keep both for now and have a duplicate icon? I agree there's not a ton of harm in leaving them but maybe @derek-boman has other thoughts

derek-boman:
I agree it doesn't hurt anything. But in the spirit of cleanness shouldn't
we look at phasing out Bike?

DB

On Jun 27, 2016, at 10:14 AM, Luke Schunk <notifications@github.com> wrote:

i'm pretty sure that the bike icon is now being used for kids - should we
keep both for now and have a duplicate icon? I agree there's not a ton of
harm in leaving them but maybe @derek-boman <https://github.com/derek-boman>
has other thoughts

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/mxenabled/mx-react-components/pull/356#issuecomment-228794496>,
or mute the thread
<https://github.com/notifications/unsubscribe/AQYoY4qspo-l1AJPmqFA20YdlmukrQU2ks5qP_bJgaJpZM4I92AC>
.

lukeschunk:
Okay - for now I've left in the `bike` and `key` icons - I talked to Tommy about this offline and we decided that since we just bumped till `v4.x.x` it's not worth doing a major version bump.

cerinman:
:shipit:

PR #357, opened 2016-06-28
Row component: Spread props to component div

cerinman:
I have a need to add an onClick handler to a grid row internally.  This PR makes that possible.  If we don't want to spread props then I can change this to just add an onClick prop.

### DIFF
--- a/src/components/grid/Row.js
+++ b/src/components/grid/Row.js
@@ -9,7 +9,7 @@ const Row = React.createClass({
     const styles = this.styles();
 
     return (
-      <div style={styles.component}>
+      <div {...this.props} style={styles.component}>
         {this.props.children}
       </div>
     );


src/components/grid/Row.js
```diff
@@ -27,4 +28,4 @@ const Row = React.createClass({
   }
 });
 
-module.exports = Row;
+module.exports = Radium(Row);
```
jmophoto: I know Radium doesn't play well with mobile browsers. Would this just be 'ignored'?

lukeschunk:
:shipit:

src/components/grid/Row.js
cerinman: Currently I think the radium issue is mostly around webviews and I think we're concerning ourselves to much with that sub set since it is covered by the mobile mx component library.

PR #358, opened 2016-06-29
Move breakpoint resolutions to style constants

cerinman:
I'm starting to see a need to hide/show/change text and other elements based upon resolution as the grid system is only a start to making things responsive.  

This PR will make our lives a bit easier by doing the following.

- Moves the default break point resolutions to the `Style` constants file for easy reuse/access
- Adds a `getWindowSize` function to the `Style` constants file that returns a string telling you what break point the screen resolution currently is in. (ported from @jmophoto's `_getScreenResolution` function the `Column` component)
- Updates the `Column` component to use the break points now contained within the `Style` constants file.

NOTE:  I didn't replace @jmophoto's `_getScreenResolution` function in the `Column` component with the one from the `Style` constants because he is merging in a break point prop to allow for custom functionality for the `Column` component.

### DIFF
--- a/src/constants/Style.js
+++ b/src/constants/Style.js
@@ -1,4 +1,10 @@
 module.exports = {
+  BreakPoints: {
+    large: 1200,
+    medium: 750,
+    small: 320
+  },
+
   Colors: {
     // GRAYSCALE
     ASH: '#ACB0B3',
--- a/src/constants/Style.js
+++ b/src/constants/Style.js
@@ -129,5 +129,19 @@ module.exports = {
     const b = parseInt(color.slice(5, 7), 16);
 
     return 'rgba(' + r + ', ' + g + ', ' + b + ', ' + opacity + ')';
+  },
+
+  getWindowSize () {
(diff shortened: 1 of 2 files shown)


src/constants/Style.js
```diff
@@ -1,4 +1,10 @@
 module.exports = {
+  BreakPoints: {
+    large: 1200,
+    medium: 750,
+    small: 320
```
paniclater: Maybe uppercase to match the other constants: `LARGE` ?

cerinman: I started down that path but then noticed the `Column` component merges a `breakpoints props with this `BreakPoint` object to keep the `Column` component dynamic.  Changing the shape of that prop to match the up case here would be a breaking change so I thought I would leave it lower case for now.

I'm assuming that `Object.assign` wouldn't merge two keys if one was upper case and the other lower case.  If I'm wrong on that assumption please let me know.

paniclater: 👍  :shipit:

PR #359, opened 2016-06-30
Calendar: Set current date to minimum date if no selected date

jmophoto:
the current date is set to the selected date or today in the calendar. This adds the minimum date to that list. 

This allows for the calendar to be advanced to the minimum date.

This also adds `day` to the check for minimum date, otherwise if you send a timestamp of today at 4, the minimum date would not be selectable.


redperk:
:shipit:

PR #360, opened 2016-06-30
Update DonutChart.js

derek-boman:
Default label is the donut chart is difficult to read, this makes the text more legible.


lukeschunk:
:shipit:

PR #362, opened 2016-07-05
Simple Input: add icon prop

jmophoto:
This adds an icon prop to the simple input component following the pattern of the date/time inputs and buttons.

- add baseColor prop for active border and icon
- add icon prop
- add styles prop that gets merged with the component style
- set baseColor to default to PRIMARY
- use onBlur and onFocus to set active border color
- remove radium

This should be backwards compatible with a warning when a `style` prop is used.

Also updates the SearchInput component to use the icon prop.

Input with Icon and SearchInput
screen shot 2016-07-05 at 11 21 24 am

Deprecation warning
screen shot 2016-07-05 at 11 21 35 am

Testing internal uses
screen shot 2016-07-05 at 11 37 57 am
screen shot 2016-07-05 at 11 48 06 am
screen shot 2016-07-05 at 11 58 39 am

### DIFF
--- a/src/components/SearchInput.js
+++ b/src/components/SearchInput.js
@@ -2,12 +2,10 @@ const React = require('react');
 const Radium = require('radium');
 
 const Input = require('./SimpleInput');
-const Icon = require('./Icon');
-
-const StyleConstants = require('../constants/Style');
 
 const SearchInput = React.createClass({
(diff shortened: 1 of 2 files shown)


iheartrachie:
:shipit:

Issue #363, opened 2016-07-06
Make all components responsive


iheartrachie:
The following components need to be made responsive:

- [ ] BarChart
- [x] DateRangePicker (priority) - Tommy
- [x] DateTimePicker (https://github.com/mxenabled/mx-react-components/pull/367)
- [x] DisplayInput (Dan)
- [ ] TimeBasedLineChart
- [x] Drawer (#361)

derek-boman:
Are more of these done? I've seen some updates to the drawer and barchart. Is there more to do there still before we can check those off?

iheartrachie:
I think for now these are good.

PR #364, opened 2016-07-06
Grid system for DateTimePicker Component

iheartrachie:
Addresses part of https://github.com/mxenabled/mx-react-components/issues/363
This adds the grid system to the DateTimePicker component to make it responsive. The only changes needed were to the inputs as the calendar on the date portion is already small enough for the smallest viewport. There are upcoming change to the grid system that should be deployed before this PR is reviewed. 

#### Large
screen shot 2016-07-06 at 3 12 25 pm

#### Medium
screen shot 2016-07-06 at 3 12 35 pm

#### Small
screen shot 2016-07-06 at 3 13 47 pm

### DIFF
--- a/src/components/DateTimePicker.js
+++ b/src/components/DateTimePicker.js
@@ -231,6 +231,7 @@ const DatePicker = React.createClass({
           </Column>
         </Row>
       </Container>
+
     );
   },
 
--- a/src/components/DateTimePicker.js
+++ b/src/components/DateTimePicker.js
@@ -3,6 +3,7 @@ const moment = require('moment-timezone/builds/moment-timezone-with-data.min');
 
 const Calendar = require('./Calendar');
 const Column = require('./grid/Column');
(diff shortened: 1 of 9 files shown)


src/components/DateTimePicker.js
```diff
@@ -147,85 +149,100 @@ const DatePicker = React.createClass({
...
     return (
-      <div style={styles.component}>
-        <div
-          onBlur={this._handleDateBlur}
-          onClick={this._handleDateClick}
-          onFocus={this._handleDateFocus}
-          ref='dateSelect'
-          style={Object.assign({}, styles.selectWrapper, this.state.showCalendar ? styles.activeSelectWrapper : null)}
-          tabIndex={0}
+      <Row style={styles.component}>
```
cerinman: Now that I've updated the grid system, you'll need to require in the new `Container` component and use it to wrap this row.

PR #365, opened 2016-07-06
Grid fix

cerinman:
This switches the Grid system over to using Bootstrap rather than detecting window size via a listener and using flex.  It also adds the new `Container` component that is made to wrap rows and columns.

I'll spare everyone the extra screen shots but here are at least two.

#### Before
screen shot 2016-07-06 at 4 21 02 pm

#### After
screen shot 2016-07-06 at 4 21 43 pm

### DIFF
--- a/src/components/grid/Column.js
+++ b/src/components/grid/Column.js
@@ -1,8 +1,6 @@
 const React = require('react');
 
-const StyleConstants = require('../../constants/Style');
-const defaultSpan = { large: 12, medium: 12, small: 12 };
-const defaultOffset = { large: 0, medium: 0, small: 0 };
+const defaultSpanOffset = { large: 0, medium: 0, small: 0 };
 
 const defaultShape = {
   large: React.PropTypes.number,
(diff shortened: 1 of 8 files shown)


src/components/grid/Column.js
```diff
@@ -12,82 +8,81 @@ const defaultShape = {
...
-    } else if (width >= breakpoints.medium) {
-      windowSize = 'medium';
+    // Column offsets
+    if (small !== 0) {
+      offsets.push('col-sm-offset-' + small);
     }
 
-    return windowSize;
-  },
+    if (medium !== 0 && medium !== small) {
```
jtanner: Why the `medium !== small` condition? Why would you not want to add the offset when they're equal?

cerinman: It's the way the Bootstrap grid system works.  You don't add every class to the element if things are equal.  It just takes the one supplied value and uses it for small, medium, and large.  I believe it would most like work with out the check but then we would end up with a lot of classes when we don't have to.

jtanner:
This looks good to me 👍

src/components/grid/Column.js
```diff
@@ -12,82 +8,81 @@ const defaultShape = {
 
 const Column = React.createClass({
   propTypes: {
-    breakpoints: React.PropTypes.shape(defaultShape),
-    columnCount: React.PropTypes.number,
     offset: React.PropTypes.shape(defaultShape),
-    order: React.PropTypes.shape(defaultShape),
```
iheartrachie: Why remove `order`? @jmophoto was this being used by us anywhere?

cerinman: I didn't see an immediate way to implement it and I didn't think we were using it anywhere yet so I pruned it for the time being to get things moving.

jmophoto: We aren't using it right now. I can see a need, but that can be added later.

jmophoto:
Looks good! :shipit:

PR #366, opened 2016-07-07
Gauge component

vintagepenguin:
### From App Demo: `Donut Chart vs. CresentChart`
screen shot 2016-07-06 at 6 39 17 pm

I'm thinking of changing the math on how the segments are setup to be `((2*Math.PI)/360)` instead of going with the `+` `-` of `(Math.PI / 180 )`. 

Shout out to @ryanbjones on helping me with the Math to get the data point to go to the right spot.

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -18,7 +18,7 @@ const {
   DonutChart,
   Drawer,
   FileUpload,
-  Gauge,
+  CresentChart,
   Icon,
   Loader,
   Modal,
@@ -704,7 +704,7 @@ const Demo = React.createClass({
         <br /><br />
         <div style={{ textAlign: 'center' }}>
           <div style={{ display: 'inline-block', padding: 20 }}>
-            <Gauge
+            <CresentChart
               activeOffset={5}
               animateOnHover={true}
               animationDuration={750}
@@ -720,17 +720,17 @@ const Demo = React.createClass({
               ]}
               defaultLabelText='Total Users'
               defaultLabelValue='300'
-              id='gauge-1'
+              id='cresentChart-1'
               numberOfSegments={50}
             />
           </div>
           <div style={{ display: 'inline-block', padding: 20 }}>
-            <Gauge
+            <CresentChart
               activeOffset={5}
               arcWidth={30}
               data={[]}
               height={200}
-              id='gauge-2'
+              id='cresentChart-2'
               numberOfSegments={10}
               showDataLabel={false}
               width={200}
--- a/demo/app.js
+++ b/demo/app.js
@@ -209,7 +209,6 @@ const Demo = React.createClass({
           selected: false
         }
       ],
-      gaugeData: [],
       icon: {
         value: 'accounts',
         displayValue: 'Accounts'
@@ -248,8 +247,7 @@ const Demo = React.createClass({
             name: 'Data Point 3',
             value: 200
           }
-        ],
-        gaugeData: []
+        ]
       });
     }, 3000);
   },
@@ -722,8 +720,8 @@ const Demo = React.createClass({
               ]}
               defaultLabelText='Total Users'
               defaultLabelValue='300'
-              numberOfSegments={50}
               id='gauge-1'
+              numberOfSegments={50}
             />
           </div>
           <div style={{ display: 'inline-block', padding: 20 }}>
@@ -732,8 +730,8 @@ const Demo = React.createClass({
               arcWidth={30}
               data={[]}
               height={200}
-              numberOfSegments={10}
               id='gauge-2'
+              numberOfSegments={10}
               showDataLabel={false}
               width={200}
             />
--- a/demo/app.js
+++ b/demo/app.js
@@ -722,6 +722,7 @@ const Demo = React.createClass({
               ]}
               defaultLabelText='Total Users'
               defaultLabelValue='300'
+              numberOfSegments={50}
               id='gauge-1'
             />
           </div>
@@ -731,6 +732,7 @@ const Demo = React.createClass({
               arcWidth={30}
               data={[]}
               height={200}
+              numberOfSegments={10}
               id='gauge-2'
               showDataLabel={false}
               width={200}
--- a/demo/app.js
+++ b/demo/app.js
@@ -726,15 +726,15 @@ const Demo = React.createClass({
             />
           </div>
           <div style={{ display: 'inline-block', padding: 20 }}>
-              <Gauge
-                activeOffset={5}
-                arcWidth={30}
-                data={[]}
-                height={200}
-                id='gauge-2'
-                showDataLabel={false}
-                width={200}
-              />
+            <Gauge
+              activeOffset={5}
+              arcWidth={30}
+              data={[]}
+              height={200}
+              id='gauge-2'
+              showDataLabel={false}
+              width={200}
+            />
           </div>
         </div>
         <br /><br />
--- a/demo/app.js
+++ b/demo/app.js
@@ -18,7 +18,7 @@ const {
   DonutChart,
   Drawer,
   FileUpload,
-  CrescentChart,
+  Gauge,
   Icon,
   Loader,
   Modal,
@@ -704,7 +704,7 @@ const Demo = React.createClass({
         <br /><br />
         <div style={{ textAlign: 'center' }}>
           <div style={{ display: 'inline-block', padding: 20 }}>
-            <CrescentChart
+            <Gauge
               activeOffset={5}
               animateOnHover={true}
               animationDuration={750}
@@ -712,6 +712,7 @@ const Demo = React.createClass({
               arcWidth={15}
               chartTotal={300}
               data={[]}
+              dataPointColors={['#FFA510']}
               dataPoints={[
                 {
                   name: 'Data Dot 1',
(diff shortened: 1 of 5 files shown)


src/components/Gauge.js
```diff
@@ -0,0 +1,263 @@
...
+
+    const segments = [
+      { startAngle: firstSegment, endAngle: endOfFirstSegment, padAngle: 0.02 },
+      { startAngle: secondSegment, endAngle: endOfSecondSegment, padAngle: 0.02 },
+      { startAngle: thirdSegment, endAngle: endOfThirdSegment, padAngle: 0.02 },
+      { startAngle: fourthSegment, endAngle: endOfFourthSegment, padAngle: 0.02 },
+      { startAngle: fifthSegment, endAngle: endOfFifthSegment, padAngle: 0.02 },
+      { startAngle: sixthSegment, endAngle: endOfSixthSegment, padAngle: 0.02 }];
+
+    return segments.map((point, i) => {
```
cerinman: We don't use index for keys any more due to a tricky issue around React not knowing how to diff keyed elements correctly.  We now use a property from current object if possible and only fall back to using index if there is no better scenario.

In this case you can add a new key/value pair the the segments objects.  Something like `id` or `name` that you can then use for the key on the group element below.

```javascript
{ id: 'firstSegment', startAngle: firstSegment, endAngle: endOfFirstSegment, padAngle: 0.02 }

<g key={point.id}>
```

src/components/Gauge.js
```diff
@@ -0,0 +1,263 @@
...
+    const secondSegment = -90 * (Math.PI / 180);
+    const endOfSecondSegment = -45 * (Math.PI / 180);
+    const thirdSegment = -45 * (Math.PI / 180);
+    const endOfThirdSegment = 0;
+    const fourthSegment = 0;
+    const endOfFourthSegment = 45 * (Math.PI / 180);
+    const fifthSegment = 45 * (Math.PI / 180);
+    const endOfFifthSegment = 90 * (Math.PI / 180);
+    const sixthSegment = 90 * (Math.PI / 180);
+    const endOfSixthSegment = 135 * (Math.PI / 180);
```
cerinman: Can there only ever be six segments to the gauge?  Hardcoding each segment seems like we're constraining ourselves.  Can this be more dynamic?  Does it need to be more dynamic?

src/components/Gauge.js
```diff
@@ -0,0 +1,263 @@
...
+      padAngle: 0.02,
+      showBaseArc: true,
+      showDataLabel: true,
+      width: 150
+    };
+  },
+
+  getInitialState () {
+    return {};
+  },
```
cerinman: If initial state returns an empty object then there is no need to include the function.  You can just remove it.

demo/app.js
```diff
@@ -246,7 +248,8 @@ const Demo = React.createClass({
             name: 'Data Point 3',
             value: 200
           }
-        ]
+        ],
+        gaugeData: []
```
cerinman: If guageData isn't changing from the initial state then I would say we don't need to include this here on `componentDidMount`

src/components/Gauge.js
```diff
@@ -0,0 +1,263 @@
...
+    return !_isEqual(this.props, nextProps) || !_isEqual(this.state, nextState);
+  },
+
+  _setupD3Functions (props) {
+    const dataSets = props.data.map(item => {
+      return item.value;
+    });
+    const startAngle = (-135 * (Math.PI / 180));
+    const endAngle = (135 * (Math.PI / 180));
+    const pie = d3.layout.pie().sort(null).padAngle(props.padAngle).endAngle(endAngle);
```
cerinman: Is making the `.sort(null)` call here even required since you're passing `null`?

src/components/Gauge.js
```diff
@@ -0,0 +1,263 @@
...
+
+    return dataPoints.map((dataPoint, index) => {
+      const percentOfTotal = dataPoint / this.props.chartTotal;
+      const endAngle = (percentOfTotal * 270) - 135;
+
+      const dataPointArc = d3.svg.arc()
+        .outerRadius(this.state.radius - this.props.activeOffset)
+        .innerRadius(this.state.radius - this.props.arcWidth)
+        .startAngle(endAngle * (Math.PI / 180))
+        .endAngle(endAngle * (Math.PI / 180));
```
cerinman: I'm seeing 180 used a lot through out the component.  Can we maybe make this a state variable that is set on `getInitialState`?  Less work to maintain down the road.

src/components/Gauge.js
```diff
@@ -0,0 +1,263 @@
...
+      if (this.props.children) {
+        return (
+          <div
+            className='mx-gauge-data'
+            onClick={this._handleClick}
+            style={styles.center}
+          >
+            {this.props.children}
+          </div>
+        );
```
cerinman: Since this is the only if statement in the true branch of the outer if statement, can we just combine them?

```javascript
if (this.props.showDataLabel && this.props.children) {
```

src/components/Gauge.js
```diff
@@ -0,0 +1,263 @@
...
+
+  render () {
+    const position = 'translate(' + this.props.width / 2 + ',' + this.props.height / 2 + ')';
+    const fontSize = Math.min(this.props.width, this.props.height) * 0.2;
+    const styles = this.styles();
+
+    return (
+      <div
+        className='mx-gauge'
+        style={[styles.component, this.props.style, { fontSize, height: this.props.height, width: this.props.width }]}
```
cerinman: If we're not using Radium for it's hover/focused/browser prefixing then we've decided to not use it in a component.  In stead, we've been using `Object.assign` to merge styles together like this.

```javascript
style={Object.assign({}, styles.component, this.props.style, { fontSize, height: this.props.height, width: this.props.width })}
```

src/components/Gauge.js
```diff
@@ -0,0 +1,263 @@
...
+        marginTop: 5
+      },
+      value: {
+        fontWeight: 300
+      }
+    };
+  }
+});
+
+module.exports = Radium(Gauge);
```
cerinman: I think we can kill Radium usage for this file.  I haven't seen it being used for anything but merging styles.  We can do that with `Object.assign` as I noted in my previous comment.

src/components/Gauge.js
```diff
@@ -0,0 +1,263 @@
...
+      center: {
+        position: 'absolute',
+        top: '50%',
+        left: '50%',
+        textAlign: 'center',
+        transform: 'translate(-50%, -50%)'
+      },
+      label: {
+        color: StyleConstants.Colors.ASH,
+        fontSize: '0.4em',
```
cerinman: We're trying to use font sizes form the style constants to keep things consistent.  https://github.com/mxenabled/mx-react-components/blob/master/src/constants/Style.js  If you have questions on which size to use then you can get with Derek from product.

src/components/Gauge.js
```diff
@@ -0,0 +1,263 @@
...
+        marginTop: 5
+      },
+      value: {
+        fontWeight: 300
```
cerinman: Same thing here.  There is a `Fonts` object in the `Style` constants file that we use if we need to bold or then something.  Again, we might get with @derek-boman to take a look to make sure he is ok with font sizes, weights, padding, and margin sizes.

src/components/Gauge.js
vintagepenguin: There design only had 6 pieces, but you're right. I think this should be dynamic. I can do that--the math is pretty easy.

src/components/Gauge.js
vintagepenguin: Yeah I like that. 👍

src/components/Gauge.js
vintagepenguin: I'm 👍  for that

src/components/Gauge.js
vintagepenguin: Cool, I'll do that.

src/components/Gauge.js
jtanner: I would also include `Math.PI /`, e.g.
```javascript
const radiansMultiplier = Math.PI / 180;
...
.endAngle(endAngle * radiansMultiplier);
```

phantomxc:
So this is more of a question on the functionality and usability of this information. Is this representing an actual gauge where 300 users can be presented in as a percentage of total users? A few examples of what I'm talking about:

image

http://bl.ocks.org/msqr/3202712

Or is this a doughnut chart with the bottom section missing?

cerinman:
Good question @phantomxc.  The screen shots of what you have @vintagepenguin does make it feel like this isn't a gauge.  That could effect the naming of the component.

src/components/Gauge.js
vintagepenguin: yeah we can remove it 👍

vintagepenguin:
@phantomxc @cerinman I got the name from this issue: https://git.moneydesktop.com/product/Insight-and-Target/issues/338

Right now, it's basically a donut component with the bottom missing right now.

vintagepenguin:
@cerinman @cerinman @jtanner I updated it so that it can be done dynamically. Here's a screenshot of what it looks like: 

screen shot 2016-07-07 at 8 09 30 pm

cerinman:
I think this should be named something different then.  I have no doubt that we'll add an actual gauge here at some point and if this one is name gauge then it will make things difficult.  I'm not sure what to call this one but if it is that specific to our internal application then maybe it should be moved to the internal application and named accordingly.

paniclater:
How about Crescent Chart?

derek-boman:
I think this is looking good. 

Just curious what else we're thinking we would use `Gauge` for? Just did a quick search and this looks pretty similar to other UI people are calling `Gauge`: https://www.google.com/search?q=gauge+ui&espv=2&biw=1152&bih=715&source=lnms&tbm=isch&sa=X&ved=0ahUKEwiZjaez6-vNAhUELmMKHSxgAPQQ_AUIBigB

jtanner:
👍

cerinman:
Ok, I just had an offline conversation with @vintagepenguin about the naming here.  Based upon that and @derek-boman's comment above, we're going to go back to naming this `Gauge`.  

The segments on the chart are there as only reference values for the ball position and the component can be drawn with a single segment which does mimic a gauge similar to the screen shot that @phantomxc linked above.  The use case for Insight/Target(adding segments) was just throwing me and making me think this was something different.

I'll ship this after the name change.

PR #367, opened 2016-07-07
Date time responsive

iheartrachie:
Addresses part of https://github.com/mxenabled/mx-react-components/issues/363
This adds the grid system to the DateTimePicker component to make it responsive. The only changes needed were to the inputs as the calendar on the date portion is already small enough for the smallest viewport. There are upcoming change to the grid system that should be deployed before this PR is reviewed. 

#### Large
screen shot 2016-07-06 at 3 12 25 pm

#### Medium
screen shot 2016-07-06 at 3 12 35 pm

#### Small
screen shot 2016-07-06 at 3 13 47 pm

### DIFF
--- a/src/components/DateTimePicker.js
+++ b/src/components/DateTimePicker.js
@@ -3,6 +3,7 @@ const moment = require('moment-timezone/builds/moment-timezone-with-data.min');
 
 const Calendar = require('./Calendar');
 const Column = require('./grid/Column');
+const Container = require('./grid/Container');
 const Icon = require('./Icon');
 const Row = require('./grid/Row');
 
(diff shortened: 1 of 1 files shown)


iheartrachie:
@cerinman I ended up cluttering up my last PR so here's a clean up with the Container component added. Thanks for your help on this!

src/components/DateTimePicker.js
```diff
@@ -145,87 +148,96 @@ const DatePicker = React.createClass({
...
-            <div style={styles.timeDisplay}>
-              {this.props.selectedDate ? moment.unix(this.props.selectedDate).format(this.props.timeFormat) : this.props.timePlaceholder}
-            </div>
-          )}
-          {this.props.timezoneFormat ? (
-            <div style={styles.timezone}>
-              {this._getTimezone(this.props.selectedDate)}
+      <Container fluid={true}>
+        <Row>
+          <Column span={{ large: inputSize, medium: inputSize, small: 12 }}>
```
cerinman: We're putting all of our span values in to a span function similar to the styles function.  Then you can use it the same way.  `const spans = this.spans();  spans.timeColumn`

iheartrachie: I like that. Much cleaner! Updated my PR.

src/components/DateTimePicker.js
```diff
@@ -145,87 +148,118 @@ const DatePicker = React.createClass({
...
+              ) : null}
+            </div>
+          </Column>
+        </Row>
+      </Container>
     );
   },
 
+  spans () {
+    return Object.assign({}, {
```
cerinman: Why the use of `Object.assign` here?  It doesn't look like your merging with anything other than just an empty object.

cerinman:
:shipit:

PR #368, opened 2016-07-08
DateRangePicker grid system/css update

cerinman:
Partially addresses https://github.com/mxenabled/mx-react-components/issues/363

This PR updates the `DateRangePicker`

- Now uses the new grid system
- Makes some of the css dynamic based upon the window size
- Changes the date format based upon the window size
- Makes the `Column` component `fluid` prop true by default as it's the more widely used case.

### Demo Screen shots
screen shot 2016-07-08 at 10 18 00 am
screen shot 2016-07-08 at 10 18 18 am

### Internal Usage screen shots
screen shot 2016-07-08 at 10 18 43 am
screen shot 2016-07-08 at 10 18 55 am

### DIFF
--- a/src/components/DateRangePicker.js
+++ b/src/components/DateRangePicker.js
@@ -420,22 +420,24 @@ const DateRangePicker = React.createClass({
       //Calendar Styles
       optionsWrapper: {
         backgroundColor: StyleConstants.Colors.WHITE,
-        border: isLargeOrMediumWindowSize ? '1px solid ' + StyleConstants.Colors.FOG : 'none',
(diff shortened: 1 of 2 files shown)


jmophoto:
:shipit:

cerinman:
I neglected to test the case where the default range options aren't shown.  Once I did that I had a bug that I needed to fix.  This needs one more quick review.

### Without default range options shown
screen shot 2016-07-08 at 10 54 12 am
screen shot 2016-07-08 at 10 54 24 am

phantomxc:
:shipit:

jmophoto:
:shipit:

PR #369, opened 2016-07-08
Column: fix offset

jmophoto:
There was an issue resetting the column offset for medium and large if small was set.

`offset: { small: 2 }`
Sets the offset of small, medium and large to 2, as expected.

`offset: { small: 2, medium: 0}`
Expected to set small to two and medium and large to 0, but didn't.

This was because of the check for `=== 0` in medium and large.

- Remove defaults for offset
- Remove `|| 0` from constants
- Check if prop exists with `>=` since just checking the prop would return false with `0`

This was tested in the Budget action bar where the offset is used.

@cerinman

### DIFF
--- a/src/components/grid/Column.js
+++ b/src/components/grid/Column.js
@@ -14,7 +14,7 @@ const Column = React.createClass({
 
   getDefaultProps () {
     return {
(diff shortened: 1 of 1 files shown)


paniclater:
:shipit:

PR #370, opened 2016-07-11
Make Drawer component wrapper div fixed positioned

cerinman:
### Issue

The drawer could be scrolled off screen due to being positioned absolute.

### Solution

The makes the component wrapper div of the Drawer component fixed positioned.  It also adds a z-index of 999.  The z-index number was chosen based upon the Drawer component's scrim being at z-index 1000 and it's contents at 1001.

#### Screen shot
screen shot 2016-07-11 at 10 07 35 am


redperk:
:shipit: 🍪

PR #372, opened 2016-07-12
Change the Column span prop so it behaves more like Bootstrap

jtanner:
This fixes a problem where the spans were defaulted to 0 when only one
or two of the spans were provided as props which caused the hidden-*
class to be applied causing surprising behavior.

Before:
`<Column span={ large:6, medium:6, small:6 }>`

After:
`<Column span={small:6}>`
This will result in a span of `{ large:6, medium:6, small:6 }`

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -452,19 +452,23 @@ const Demo = React.createClass({
               >
                 <div style={styles.responsiveDiv}>1</div>
               </Column>
+            </Row>
+            <Row>
               <Column span={{ large: 4, medium: 3, small: 10 }} >
                 <div style={styles.responsiveDiv}>2</div>
               </Column>
               <Column
-                span={{ large: 4, medium: 9 }}
+                span={{ large: 4, medium: 6, small: 0 }}
               >
                 <div style={styles.responsiveDiv}>3</div>
               </Column>
               <Column
-                span={{ large: 4 }}
+                span={{ large: 4, medium: 9, small: 2 }}
               >
                 <div style={styles.responsiveDiv}>4</div>
               </Column>
+            </Row>
+            <Row>
               <Column
                 order={{ medium: -1 }}
                 span={{ large: 3, medium: 6, small: 9 }}
@@ -473,10 +477,12 @@ const Demo = React.createClass({
               </Column>
               <Column
                 order={{ medium: -2 }}
-                span={{ large: 3, medium: 6, small: 9 }}
+                span={{ large: 9, medium: 6, small: 3 }}
               >
                 <div style={styles.responsiveDiv}>6</div>
               </Column>
+            </Row>
+            <Row>
               <Column
                 span={{ large: 6 }}
               >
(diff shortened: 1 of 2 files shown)


jmophoto:
So, this is what we had before, but with Tommy's changes we changed to follow Bootstraps pattern of only declaring the first property and the rest will follow. So a `small` of 6 will result in a `medium` and `large` of six also. 

I like this way better, though. Caused a lot of confusion for me wondering why widths that were not implicitly set were being `auto-magically` done.

paniclater:
I also prefer this way -- but there is definitely something to be said for following bootstrap's convention esp. now that we're using their css.

cerinman:
Yeah, I was following the Bootstrap convention.  I lean towards leaving it the way it is.  I think it is more confusing this way as it doesn't match normal css short hand convention but I will bow to the masses if everyone hates it.

src/components/grid/Column.js
```diff
@@ -15,30 +15,34 @@ const Column = React.createClass({
   getDefaultProps () {
     return {
       offset: {},
-      span: { large: 12, medium: 12, small: 12 }
+      span: { large: 12 }
```
jtanner: ^ to give a hint that not all are required

jtanner:
After looking at this some more I think the problem lies with [setting a default of 0](https://github.com/mxenabled/mx-react-components/blob/30849faacd85d012f371efcb69a77a9a558158bd/src/components/grid/Column.js#L24) and then [adding the `hidden-*` class](https://github.com/mxenabled/mx-react-components/blob/30849faacd85d012f371efcb69a77a9a558158bd/src/components/grid/Column.js#L30).

I've amended my commit to reflect this new finding.

It looks like @jmophoto already fixed the offsets [here](https://github.com/mxenabled/mx-react-components/commit/30849faacd85d012f371efcb69a77a9a558158bd#diff-e08bcf18e26bf986765255f1a2c623be).

jtanner:
@jmophoto @paniclater @cerinman this is ready for another look.

paniclater:
:shipit:

demo/app.js
```diff
@@ -452,19 +452,23 @@ const Demo = React.createClass({
...
                 <div style={styles.responsiveDiv}>1</div>
               </Column>
+            </Row>
+            <Row>
               <Column span={{ large: 4, medium: 3, small: 10 }} >
                 <div style={styles.responsiveDiv}>2</div>
               </Column>
               <Column
-                span={{ large: 4, medium: 9 }}
+                span={{ large: 4, medium: 6, small: 0 }}
```
jtanner: I used `small: 0` to verify that the `hidden-sm` class is used.

jtanner:
@paniclater @cerinman can I get another look at this? I did some testing with the demo page and found that `if (small) { ...` was not inserting the `hidden-sm` class. I fixed it and also updated the demo page.

cerinman:
:shipit:

PR #374, opened 2016-07-14
Color gauge value label

vintagepenguin:
Running into a problem where the first color value for the segment should be a different color than the center value. This allows you to pass in a value so that the two are not the same. 

### Demo of Problem
screen shot 2016-07-13 at 6 37 06 pm

### Demo of Solution...`yes the solution is hot pink. #80sAreComingBack`

screen shot 2016-07-13 at 6 39 24 pm

Once this goes up, I'll have to update this PR: robinhood #1178

@cerinman @jtanner @paniclater


src/components/Gauge.js
```diff
@@ -17,6 +17,7 @@ const Gauge = React.createClass({
     dataPointRadius: React.PropTypes.number,
     dataPoints: React.PropTypes.array,
     defaultLabelText: React.PropTypes.string,
+    defaultLabelValueColor: React.PropTypes.string,
```
lukeschunk: sort the props alphabetically so Travis passes and this should be good

cerinman: I would also change the name of this prop to `defaultLabelColor` as well.  Adding the word `value` to it makes it read weird.

vintagepenguin: I made the distinction because the value & the text beneath it are usually two different colors. For instance, in the image above and in the product designs they want the value, `300` to be a different color than the text beneath it `Total Users`. 

But the name is a little bit strange. What do you think of changing it to be `defaultValueColor`?

vintagepenguin:
##### Screenshot showing that you can modify both text and number colors

screen shot 2016-07-14 at 1 52 24 pm

cerinman:
:shipit:

PR #375, opened 2016-07-14
Remove illeagal attributes causing react warnings

jmophoto:
Remove {...this.props} on buttonGroup and button containing divs and the fit attribute on the svg.

Checking https://developer.mozilla.org/en-US/docs/Web/SVG/Element/svg, I don't see that `fit` is a valid attribute on an svg.

I tested these changes in Serenity and everything appears to be functioning correctly.


redperk:
:shipit:

PR #379, opened 2016-07-25
Allows BarChart to accept negative values

tumentumurchudur:
This PR allows users to pass both negative and positive numbers to the BarChart component. This feature is needed for internal use.

screen shot 2016-07-25 at 2 39 53 pm

### DIFF
--- a/src/components/BarChart.js
+++ b/src/components/BarChart.js
@@ -59,10 +59,8 @@ const Rect = React.createClass({
 
   _handleMouseOver (label, value, x, y) {
     const animateDuration = 500;
-    const labelMargin = 200;
-    const yPos = value < 0 ? labelMargin : y;
 
-    this.props.onHover(label, value, x, yPos);
+    this.props.onHover(label, value, x, y);
 
     this.setState({
       hovering: true,
@@ -150,7 +148,7 @@ const BarChart = React.createClass({
   },
 
   _renderBar (yScale, xScale, value, index) {
-    const height = Math.abs(yScale(value) - yScale(0));
+    const height = value < 0 ? Math.abs(yScale(value) - yScale(0)) : yScale(value);
     const x = xScale(index);
     const barWidth = xScale.rangeBand();
     const y = value > 0 ? this.props.height - height : 0;
@@ -179,9 +177,12 @@ const BarChart = React.createClass({
     const data = this.props.data.map(d => {
       return d.value;
     });
+    let negativeBarsDrawn = false;
+    let positiveBarsDrawn = false;
+    const y0 = Math.max(Math.abs(d3.min(data)), Math.abs(d3.max(data)));
 
     const yScale = d3.scale.linear()
-      .domain([d3.min(data), d3.max(data)])
+      .domain([-y0, y0])
       .range([0, height]);
 
     const xScale = d3.scale.ordinal()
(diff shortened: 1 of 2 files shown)


redperk:
Is there any way we can get a screen shot of how Helios handles negative values? I just wonder how this compares to Helios since this will be used on mobile devices.

paniclater:
@derek-boman Can you confirm this is good from a design perspective?

Code looks good to me --  so I'll ship as soon as you confirm.

derek-boman:
I'd like to match the behavior in Helios on this. See screenshot, talk to me if you have further questions.

screenshot_20160726-083403

tumentumurchudur:
@derek-boman Here is what the barchart looks like based on our discussion. Let me know if it works.

screen shot 2016-07-26 at 3 03 14 pm

derek-boman:
Yes this looks pretty good. A few observations:

- [ ] What is going on with the 625.25? Is that a tool tip? I think we should match Insight bar chart tooltips.
- [ ] The labels beneath still don't look centered with the bars to me. They should also be Small, Ash if they are not already. 
- [ ] Bar should be brand color in Accounts drawer.

tumentumurchudur:
@derek-boman Take a look at the following screenshot and let me know if the design is good to go.

- The tooltip is displayed and formatted in the parent component. In other words, this component does not handle the logic to show the tooltips
- The labels should be centered
- On bar hover, brand color is used

screen shot 2016-07-27 at 4 38 48 pm

tumentumurchudur:
Ready for review again.

- [ ] What is going on with the 625.25? Is that a tool tip? I think we should match Insight bar chart tooltips.
- [ ] The labels beneath still don't look centered with the bars to me. They should also be Small, Ash if they are not already. 
- [ ] Bar should be brand color in Accounts drawer.

screen shot 2016-07-28 at 9 51 57 am

cerinman:
Code looks good.  @derek-boman are you good with the changes now?

src/components/BarChart.js
```diff
@@ -89,6 +95,8 @@ const Rect = React.createClass({
         onClick={this.props.onClick.bind(null, this.props.value, this.props.label)}
         onMouseOut={this._handleMouseOut}
         onMouseOver={this._handleMouseOver.bind(null, this.props.label, this.props.value, this.props.x, y)}
+        rx={3}
+        ry={3}
```
jtanner: What do the `3` values do here? Do they correspond to `const shift = this.props.value > 0 ? '3px' : '-3px'` above? If so it may be good to give it a name.

tumentumurchudur: It moves the bars down a little to hide the bottom rounded corners. We only want the top of the bars rounded. It is named `shift`. Do you want me to change it to something little more descriptive?

redperk:
:shipit: From a code perspective it looks good. Well done on figuring out the complex d3 stuff! 🍪  I went ahead and gave it a shipit, but @derek-boman, speak up if you have any concerns from a design perspective.

src/components/BarChart.js
tumentumurchudur: `rx={3}` adds rounded corners to `rect`.

jtanner: No, I think it's fine for now.

derek-boman:
Looks good.

Issue #373, opened 2016-07-13
Display Input / Responsive Revisions


derek-boman:
The Display Input component has some problems adapting to small screens. This should also set some patterns how other Display Menus (date picker, category selectors, tag lists, etc) behave to add consistency to forms and "Details" pages.

**Current "Transaction Details" page on mobile:**
![screen shot 2016-07-13 at 10 57 41 am](https://cloud.githubusercontent.com/assets/17180771/16812186/3b1ab81e-48e9-11e6-911c-15b267917482.png)

**Prototype "Transaction Details" with revisions:**
Mobile: https://marvelapp.com/3ag56fe#13715726
Desktop: https://marvelapp.com/hfji73#13772514

**Frontify spec:** https://app.frontify.com/d/KHQZnlj1XMIO/web-patterns#/inputs/form-elements/input

**New Design:**
![display inputs](https://cloud.githubusercontent.com/assets/17180771/16812887/62243568-48ec-11e6-8f66-527c47de4962.png)
![display input - mobile](https://cloud.githubusercontent.com/assets/17180771/16812892/674d6000-48ec-11e6-8155-ed244080b8f2.png)

**Construction**
![input construction](https://cloud.githubusercontent.com/assets/17180771/16812898/6c8107ca-48ec-11e6-8d7e-f94aff9c5e7b.png)
![input construction - mobile](https://cloud.githubusercontent.com/assets/17180771/16812899/70759080-48ec-11e6-9563-4d51206db9f7.png)

vintagepenguin:
- [ ] Responsive 
- [ ] Refactor Validations

derek-boman:
Here are some additional specs.

![typography](https://cloud.githubusercontent.com/assets/17180771/17229930/0c7b2f58-54d7-11e6-9be9-693db843c877.png)
![padding](https://cloud.githubusercontent.com/assets/17180771/17229931/0c7e2050-54d7-11e6-96bc-6307588e1da3.png)
![padding - mobile](https://cloud.githubusercontent.com/assets/17180771/17229929/0c7ab1f4-54d7-11e6-8722-d645d198a657.png)

vintagepenguin:
Closing this since Tommy's PR resolved it 👍

derek-boman:
@vintagepenguin: can we close this then?

PR #380, opened 2016-07-29
Display input tweaks/fixes

cerinman:
https://github.com/mxenabled/mx-react-components/issues/373

This PR makes some changes and fixes to the DisplayInput component.  The last set of changes made it responsive by implementing the grid system but @vintagepenguin and I noticed some issues and inconsistencies with the component internally after the upgrade.

The following was done after discussion with @vintagepenguin @derek-boman and @paniclater.

- Allow children used in the `DisplayInput` to be used in place of the input.  This allows us to use the `DisplayInput` for things like account selects and category selects on forms.  Currently we use a separate internal component for these things but no one has time to maintain two sets of styles to keep things looking consistent.
- Adds a consistent padding/margin to match product designs to the various parts of the component
- Uses the `StyleConstants.Spacing` numbers for the above padding/margins
- Only renders the status column if it exists.  Before we were rendering the column with empty content so it was taking up space and made the `DisplayInput` look weird.
- Drops empty column used for spacing and instead uses offset to achieve same goal.
- Uses `flex` and `alignItems` to correctly align the label and input/children.
- Corrects the label color to be `charcoal`.  It was black but we don't use black for anything.  I also removed the color black from the `StyleConstants`.
- Adds a `relative` prop to the `Column` component.  Bootstrap adds a position relative to columns which works in almost all cases.  The issue we were running into internally was that using the display input to wrap a select component that needed to show it's options outside it's parent wrapper was impossible.  By adding this prop we allow ourselves to set that position to `static` were needed.  The prop defaults to true the position will continue to be relative unless we set it false.  All of our testing has shown that having the position set to `static` does no harm to the `DisplayInput` in question or it's surrounding neighbors.
- Drops the fixed height of the `DisplayInput`.  Having a fixed height was causing positioning issues and it was decided after discussing the issue with @derek-boman that it would be better to drop the fixed height and let the padding position things as intended rather than trying to do one off padding/margin styles to try and align things.  We'll need to keep this in mind when using the `DisplayInput` to wrap things like selectors if we want those `DisplayInputs` to maintain the same height as those that use the input.

#### Screen Shots

screen shot 2016-07-29 at 8 49 06 am
screen shot 2016-07-29 at 8 49 19 am
screen shot 2016-07-29 at 8 49 27 am
screen shot 2016-07-29 at 8 50 04 am
screen shot 2016-07-29 at 8 50 23 am
screen shot 2016-07-29 at 8 51 38 am

### DIFF
--- a/src/components/DisplayInput.js
+++ b/src/components/DisplayInput.js
@@ -166,7 +166,7 @@ const DisplayInput = React.createClass({
 
       status: {
         paddingBottom: StyleConstants.Spacing.XSMALL,
-        paddingLeft: StyleConstants.Spacing.SMALL,
+        paddingLeft: isLargeOrMediumWindowSize ? StyleConstants.Spacing.SMALL : StyleConstants.Spacing.XSMALL,
         paddingRight: StyleConstants.Spacing.SMALL,
         paddingTop: StyleConstants.Spacing.XSMALL
       },
--- a/src/components/DisplayInput.js
+++ b/src/components/DisplayInput.js
@@ -65,7 +65,7 @@ const DisplayInput = React.createClass({
             ) : null }
 
             <Column relative={!hasChildren} span={inputColumn}>
-              {this.props.children ? (
+              {hasChildren ? (
                 <div style={styles.children}>
                   {this.props.children}
                 </div>
@@ -136,19 +136,13 @@ const DisplayInput = React.createClass({
       inputWrapper: {
         alignItems: 'center',
         display: 'flex',
-        paddingBottom: StyleConstants.Spacing.SMALL,
-        paddingLeft: StyleConstants.Spacing.SMALL,
-        paddingRight: StyleConstants.Spacing.SMALL,
-        paddingTop: StyleConstants.Spacing.SMALL
+        padding: StyleConstants.Spacing.SMALL
       },
 
       children: {
         alignItems: 'center',
         display: 'flex',
-        paddingBottom: StyleConstants.Spacing.SMALL,
-        paddingLeft: StyleConstants.Spacing.SMALL,
-        paddingRight: StyleConstants.Spacing.SMALL,
-        paddingTop: StyleConstants.Spacing.SMALL
+        padding: StyleConstants.Spacing.SMALL
       },
 
       labelText: {
--- a/src/components/DisplayInput.js
+++ b/src/components/DisplayInput.js
@@ -125,9 +125,7 @@ const DisplayInput = React.createClass({
         border: '1px solid transparent',
         fontSize: StyleConstants.FontSizes.LARGE,
         height: isLargeOrMediumWindowSize ? '100%' : '75%',
-        paddingBottom: isLargeOrMediumWindowSize ? 10 : 0,
-        paddingTop: 0,
-        paddingLeft: textIndent,
+        paddingBottom: isLargeOrMediumWindowSize ? 20 : 0,
         textAlign: 'left',
         width: '100%',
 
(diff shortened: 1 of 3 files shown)


jmophoto:
Looks good. :shipit:

src/components/DisplayInput.js
```diff
@@ -54,23 +55,31 @@ const DisplayInput = React.createClass({
...
-              />
+            <Column relative={!hasChildren} span={inputColumn}>
+              {this.props.children ? (
+                <div style={styles.children}>
+                  {this.props.children}
+                </div>
+              ) : (
+                <div style={styles.inputWrapper}>
+                  <input
+                    {...this.props}
```
cerinman: Yeah I didn't go down that rabbit hole on purpose for this PR but I agree we need to do some clean up around stuff like that.

src/components/DisplayInput.js
```diff
@@ -54,23 +55,31 @@ const DisplayInput = React.createClass({
...
-            <Column span={inputColumn}>
-              <input
-                {...this.props}
-                key='input'
-                label={this.props.label}
-                style={styles.input}
-                type='text'
-              />
+            <Column relative={!hasChildren} span={inputColumn}>
+              {this.props.children ? (
```
tumentumurchudur: no biggie but why not use `hasChildren` here?

src/components/DisplayInput.js
```diff
@@ -128,28 +133,42 @@ const DisplayInput = React.createClass({
...
       },
 
-      label: {
-        ':focus': {
-          color: this.props.primaryColor
-        }
+      inputWrapper: {
+        alignItems: 'center',
+        display: 'flex',
+        paddingBottom: StyleConstants.Spacing.SMALL,
```
tumentumurchudur: I would condense it
`padding: StyleConstants.Spacing.SMALL`

src/components/DisplayInput.js
```diff
@@ -128,28 +133,42 @@ const DisplayInput = React.createClass({
...
+        paddingLeft: StyleConstants.Spacing.SMALL,
+        paddingRight: StyleConstants.Spacing.SMALL,
+        paddingTop: StyleConstants.Spacing.SMALL
+      },
+
+      children: {
+        alignItems: 'center',
+        display: 'flex',
+        paddingBottom: StyleConstants.Spacing.SMALL,
```
tumentumurchudur: same here

src/components/DisplayInput.js
```diff
@@ -84,20 +93,20 @@ const DisplayInput = React.createClass({
         </div>
 
         <Row>
```
tumentumurchudur: Do we really want the `status` message on a new row? I think it would be more intuitive to display it inside the `input` field on the far right. What do you think?

cerinman: That change was made after discussion with @derek-boman.  It was originally in the same row and positioned to the right as you suggest.  I think the main problem is that it's position to the right doesn't work for mobile.

src/components/DisplayInput.js
cerinman: Agreed.

src/components/DisplayInput.js
cerinman: Yep, that makes more sense.  I'll get it fixed.

src/components/DisplayInput.js
jmophoto: The only concern I had with the short hand is if we have to override one. That may not be an issue, though or we can use the short hand to do that. It just makes the use a little more coupled because we have to know all the padding and make sure we pass in the same amounts for the ones we don't want to override.

src/components/DisplayInput.js
vintagepenguin: Having `status` on the new row makes it more flexible for `small`, `medium`, and `large` views. 👍

cerinman:
@tumentumurchudur changes made.  Can I get another review when you have a sec.

tumentumurchudur:
:shipit:

PR #384, opened 2016-08-02
Change log update for version 4.2.17

cerinman:
No docs required updating and the Icon docs are auto updated via the App constants file that was already updated with the new icons.


redperk:
I'm guessing there isn't a travis for gh-pages

cerinman:
nope

redperk:
:shipit:

Issue #385, opened 2016-08-03
Donut Chart Animation Broken


paniclater:
The donut chart animation works correctly on initial mount, but then does not work on subsequent renders.

paniclater:
This has been verified fixed.

PR #386, opened 2016-08-03
Fixing Chart Animation with New Props

paniclater:
<h2>The Run Down:

Issue: https://github.com/mxenabled/mx-react-components/issues/385

`componentWillReceiveProps` is called after shouldComponentUpdate, but before render. It is also not called on the initial render -- which is why only subsequent animations were busted.

The animation functions needed to be called **after** render.

`componentDidUpdate` is called after render and putting the animation calls there fixed the issue. It is also not called on the initial render -- which means it doesn't break the initial animation.

Notice in the broken version only some of the arcs animate, while in the fixed version all of the arcs animate every time. 

**Bonus** this fixes both roll and bounce animations!

--

<h3>Broken:
ezgif com-video-to-gif 4

--

<h3>Fixed:
ezgif com-video-to-gif 4

--

Gigantic Shout Out to :cookie::cookie::eggplant:@cerinman & @phantomxc:eggplant::cookie::cookie: for helping me in my moments of despair. I was completely stumped without their help.

### DIFF
--- a/src/components/DonutChart.js
+++ b/src/components/DonutChart.js
@@ -80,22 +80,22 @@ const DonutChart = React.createClass({
   },
 
   componentWillReceiveProps (newProps) {
-    if (!_isEqual(this.props.data, newProps.data)) {
-      this._setupD3Functions(newProps);
-      this._animateChart();
-    }
+    this._setupD3Functions(newProps);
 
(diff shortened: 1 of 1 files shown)


jtanner:
:shipit: 🎉

lukeschunk:
🍪

redperk:
🍪

PR #387, opened 2016-08-03
Add support for running jekyll locally to view doc changes

jtanner:
If you're reviewing this please checkout the branch and follow the README instructions to get jekyll running.


paniclater:
Excited to try this tomorrow!

README.md
```diff
@@ -5,3 +5,16 @@ npm install mx-react-components
...
 ## Usage
 [See Documentation](http://moneydesktop.github.io/mx-react-components/)
+
+## Development
+
+To view your documentation changes before committing you can run the following commands:
+
+```
+npm install
+npm run build
```
jtanner: This command (which runs `webpack`) doesn't watch, do you think it would be a better workflow to have two console sessions running: one for webpack watch and one for jekyll server?

paniclater:
@jtanner: how about just adding one more script to the package.json:

`"dev": "npm install && npm run build && bundle install && bundle exec jekyll serve"`

that way we can have one process to run it locally?

README.md
paniclater: Here's an idea -- if we install and save `webpack-dev-server` then use the command `webpack-dev-server --inline --hot` in place of the jeckyll stuff then we actually don't need jeckyll installed locally and get hot module replacement.

iheartrachie:
I'm very excited for this! I agree with Ryan's comments about running webpack inline (didn't know we could do that) \o/ I think falling back to having two console sessions is nbd anyway. Thanks for your work on this!

jtanner:
I spoke with @paniclater offline and he's going to push up a different PR that uses `webpack-dev-server` with hot reloading which I think is a much better workflow than using jekyll.

jtanner:
See #389

PR #390, opened 2016-08-03
Preventing Opening Animation From Running on Every Render

paniclater:
<h2>The Run Down:

Issue: https://github.com/mxenabled/mx-react-components/issues/385

Moving the animation to componentDidMount caused the opening animation to run on every hover, click, etc. This protects against that.

--

<h3>Broken:
ezgif com-video-to-gif 5

--

<h3>Fixed:
ezgif com-video-to-gif 6


redperk:
:shipit:

redperk:
:shipit:

PR #392, opened 2016-08-04
Add onItemClick Prop to SimpleSelect

vintagepenguin:
Gives you the ability to pass a function to be called when you click on a specific item. 

## ⚡  LIVE DEMO ⚡ 

[https://gyazo.com/1504054ca9833fb550e9b8c613634cfd](https://gyazo.com/1504054ca9833fb550e9b8c613634cfd)


cerinman:
I don't think this is needed.  You can already pass in a function for each menu item.

PR #394, opened 2016-08-04
DateTimePicker: Fix for IE

jmophoto:
Two IE issues:

1. `String.includes` is not supported in IE. Changed that to `String.indexOf`
2. The onDateSelect was not triggered when the calendar was added to the page. Changed to the pattern used in the DatePicker with `display` hide/show.

### DIFF
--- a/src/components/DateTimePicker.js
+++ b/src/components/DateTimePicker.js
@@ -114,13 +114,13 @@ const DatePicker = React.createClass({
     const value = e.target.value.toLowerCase().replace('.', '');
 
     const time = value.split(':');
-    const am = value.includes('am');
-    const pm = value.includes('pm');
+    const am = value.indexOf('am');
+    const pm = value.indexOf('pm');
     const minute = Number(time[1].substring(0, 2));
     let hour = Number(time[0]);
 
-    hour = am && hour === 12 ? 0 : hour;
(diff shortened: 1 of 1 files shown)


src/components/DateTimePicker.js
```diff
@@ -114,13 +114,13 @@ const DatePicker = React.createClass({
...
-    const am = value.includes('am');
-    const pm = value.includes('pm');
+    const am = value.indexOf('am');
+    const pm = value.indexOf('pm');
     const minute = Number(time[1].substring(0, 2));
     let hour = Number(time[0]);
 
-    hour = am && hour === 12 ? 0 : hour;
-    hour = pm && hour !== 12 ? hour + 12 : hour;
+    hour = am >= 0 && hour === 12 ? 0 : hour;
```
jtanner: I think `am` and `pm` should remain booleans, can you move `>= 0` up to the `const` declarations?

cerinman: I like that idea.

```javascript
const am = value.includes('am') >= 0;

hour = am && hour === 12 ? 0 : hour;
```

jtanner:
:shipit:

src/components/DateTimePicker.js
jmophoto: Agreed. Changed that and pushed.

PR #395, opened 2016-08-04
Add Tabs Component for Full-Width || Mobile

vintagepenguin:
## Full-Width

[https://gyazo.com/ccd281d2b0a9b35b7e886754b276461f](https://gyazo.com/ccd281d2b0a9b35b7e886754b276461f)

## Mobile

[https://gyazo.com/a2b63362b4ec5bb1dbb00d03a06dace2](https://gyazo.com/a2b63362b4ec5bb1dbb00d03a06dace2)


vintagepenguin:
## GH-PAGES PR: 
https://github.com/mxenabled/mx-react-components/pull/396

src/components/Tabs.js
```diff
@@ -0,0 +1,203 @@
...
+          style={[styles.tab, this.state.selectedTab === _index && selectedTabStyle]}
+        >
+          {tab}
+        </span>
+      );
+    });
+  },
+
+  _renderTabMenu () {
+    const selectedTab = this.props.tabs[this.state.selectedTab];
```
tumentumurchudur: I would rename this to `selectedTabName`.

src/components/Tabs.js
```diff
@@ -0,0 +1,203 @@
...
+
+  componentWillReceiveProps (nextProps) {
+    if (nextProps.selectedTab !== this.state.selectedTab) {
+      this.setState({
+        selectedTab: nextProps.selectedTab
+      });
+    }
+  },
+
+  _handleMenuClick () {
```
tumentumurchudur: Can this be renamed to `_toggleMenu`?

src/components/Tabs.js
```diff
@@ -0,0 +1,203 @@
...
+
+const Icon = require('./Icon');
+const SimpleSelect = require('./SimpleSelect');
+const StyleConstants = require('../constants/Style');
+
+const Tabs = React.createClass({
+  propTypes: {
+    brandColor: React.PropTypes.string,
+    onTabSelect: React.PropTypes.func.isRequired,
+    selectedBottomBorderPadding: React.PropTypes.number,
```
tumentumurchudur: Can this prop be determined based on the screen size?

src/components/Tabs.js
```diff
@@ -0,0 +1,203 @@
...
+            menuStyles={styles.menu}
+            onScrimClick={this._handleMenuClick}
+            showItems={this.state.showMenu}
+          />
+        ) : null}
+      </div>
+    );
+  },
+
+  _tabItems () {
```
tumentumurchudur: I feel like you can just remove this function and move its logic to `_renderTabMenu` function.

src/components/Tabs.js
```diff
@@ -0,0 +1,203 @@
...
+    return windowSize === 'medium' || windowSize === 'large';
+  },
+
+  _renderTabs () {
+    const selectedTabStyle = {
+      borderBottom: '2px solid ' + this.props.brandColor,
+      paddingBottom: this.props.selectedBottomBorderPadding
+    };
+    const styles = this.styles();
+
```
phantomxc: I think we should allow a style props and object assign with the defined function.

src/components/Tabs.js
vintagepenguin: `renderTabMenu` should be in charge of rendering the items, but not building the `_tabItems`. That's why I split them out. Splitting these out also makes it more discoverable for anyone jumping in.

src/components/Tabs.js
vintagepenguin: Yeah I like that name 👍

src/components/Tabs.js
tumentumurchudur: Are you going to address this?

src/components/Tabs.js
tumentumurchudur: That is fine. Can you rename it to `_buildTabItems` then?

src/components/Tabs.js
```diff
@@ -0,0 +1,204 @@
...
+const Tabs = React.createClass({
+  propTypes: {
+    activeTabStyles: React.PropTypes.object,
+    brandColor: React.PropTypes.string,
+    onTabSelect: React.PropTypes.func.isRequired,
+    selectedTab: React.PropTypes.number,
+    showBottomBorder: React.PropTypes.bool,
+    tabs: React.PropTypes.array.isRequired,
+    useTabsInMobile: React.PropTypes.bool
```
tumentumurchudur: Can this be determined from the screen size?

vintagepenguin: This was something that @derek-boman wanted because he said that the normal case is that we would want the `menu` in mobile but there is one case where we would not want the `menu` in mobile but we would want to use `tabs` in mobile instead.

tumentumurchudur:
:shipit:

PR #397, opened 2016-08-07
DateRangePicker - Close calendar on date range selection complete

cerinman:
Closes the calendar in the `DateRangePicker` on date range completion or if a default range is
selected and the new prop `closeCalendarOnRangeSelect` is set true.  `closeCalendarOnRangeSelect` defaults to false.

close-on-range-select


jmophoto:
:shipit:

PR #378, opened 2016-07-22
Button - disable Hover prop to make mobile friendly 

lukeschunk:
potential way to disable hover to make mobile friendly

PR #398, opened 2016-08-08
Fix spelling for Prop in Tabs demo

vintagepenguin:
## WHY THE FRANK DOES THIS PR EXIST? 
When the prop `useTabsInMobile={true}` then the Hover state is applied. Hover, in mobile, is applied by clicking. So the color in mobile will change on click which is not something we, or other people, would probably want. 

### Color Not Changing onClick in Mobile Anymore 👍 
[https://gyazo.com/a10eaa6579abc2c74f832c0d43877ada](https://gyazo.com/a10eaa6579abc2c74f832c0d43877ada)

### Hover Still Workin' in Full Screen
[https://gyazo.com/3f98179d9c388c4f7e8ca78acf6f3a0f](https://gyazo.com/3f98179d9c388c4f7e8ca78acf6f3a0f)

* I forgot the `s` on `activeTabStyles` prop so the styles weren't getting displayed. I also made this mistake in the `gh-pages`. So I figured I'd fix it now.

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -763,7 +763,7 @@ const Demo = React.createClass({
 
         <div style={{ marginLeft: '30%', width: '100%' }}>
           <Tabs
-            activeTabStyle={{ paddingBottom: 25 }}
+            activeTabStyles={{ paddingBottom: 25 }}
             onTabSelect={this._handleTabSelect}
             selectedTab={this.state.selectedTab}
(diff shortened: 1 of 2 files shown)


lukeschunk:
I handled a similar thing like this: https://github.com/mxenabled/mx-react-components/pull/378

lukeschunk:
:shipit:

PR #402, opened 2016-08-11
move padding so it can be easily overridden

lukeschunk:
Moved the padding so I can override the style in the DatePickerButtonGroup internally.


iheartrachie:
:shipit:

PR #405, opened 2016-08-16
Make the close fn public for closing the Drawer externally

jtanner:
This will allow users to externally close a `Drawer` with the animation!

_Thanks to [this guide](https://facebook.github.io/react/tips/expose-component-functions.html) for the tip on how to do it._

ezgif com-video-to-gif

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -349,6 +349,10 @@ const Demo = React.createClass({
     });
   },
 
+  _handleCloseDrawerClick () {
+    this.refs.drawer.close();
+  },
+
   _onHideDrawer () {
     this.setState({
       showDrawer: false,
@@ -532,9 +536,15 @@ const Demo = React.createClass({
               headerStyle={{ backgroundColor: '#fff' }}
               navConfig={navConfig}
               onClose={this._onHideDrawer}
+              ref='drawer'
               title='This is the drawer component'
             >
               <div style={{ padding: 20, fontFamily: 'Helvetica, Arial, sans-serif' }}>Insert Custom Content Here</div>
+              <Button
+                onClick={this._handleCloseDrawerClick}
+              >
+                Close Drawer
+              </Button>
             </Drawer>
           </div>
         ) : null}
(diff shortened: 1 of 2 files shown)


redperk:
Do you have any screenshots or a gif?

src/components/Drawer.js
```diff
@@ -67,7 +67,12 @@ const Drawer = React.createClass({
     }
   },
 
-  _handleCloseClick () {
+  /**
+   * Animate the Drawer closed and then call the onClose callback.
+   *
+   * @returns {Promise} that is resolved when the animation finishes
+   */
+  close () {
```
tumentumurchudur: I would rename it back to `_handleCloseClick`. That is our convention to name functions inside components.

demo/app.js
```diff
@@ -532,9 +536,15 @@ const Demo = React.createClass({
               headerStyle={{ backgroundColor: '#fff' }}
               navConfig={navConfig}
               onClose={this._onHideDrawer}
+              ref='drawer'
```
redperk: So it's the ref attribute that exposes drawer component?

jtanner: Yes

src/components/Drawer.js
jtanner: @tumentumurchudur using the `_` prefix is a common convention in javascript for private functions. I think it might be confusing for other javascript developers to call what looks like a private function so I would argue that it's better to keep it as `close`. Would you prefer having two functions a private `_handleCloseClick` which is used internally and called by the public `close` function?

Take a look at the `animate` function in this example from the React docs: https://facebook.github.io/react/tips/expose-component-functions.html

jtanner:
@redperk I've added a gif of the demo app.

src/components/Drawer.js
tumentumurchudur: Oh I didn't realize that `close` was an exposed function. It is good then. 👍

redperk:
:shipit:

redperk:
Also, as a side note, it's funny to me that they are using `index` as their `key` in the official docs.

redperk:
:shipit:

PR #406, opened 2016-08-16
DateRangePicker: added isRelative prop, added xsmall window size

lukeschunk:
This enables us to pass a prop that changes relative positioning on a some of the date range picker outer divs in order to position it easier in mobile. I also added another window size, `xsmall` that we can check for when deciding which props to pass, etc. 

@cerinman lmk your thoughts.

### DIFF
--- a/src/constants/Style.js
+++ b/src/constants/Style.js
@@ -140,6 +140,8 @@ module.exports = {
       windowSize = 'large';
     } else if (width >= breakPoints.medium) {
       windowSize = 'medium';
+    } else if (width <= breakPoints.small) {
+      windowSize = 'xsmall';
     }
 
     return windowSize;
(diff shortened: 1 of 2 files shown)


src/constants/Style.js
```diff
@@ -140,6 +140,8 @@ module.exports = {
       windowSize = 'large';
     } else if (width >= breakPoints.medium) {
       windowSize = 'medium';
+    } else if (width <= breakPoints.small) {
+      windowSize = 'xsmall';
```
jtanner: I like it 👍 

I just added `getWindowSize` to the docs (#411). Please add something about the new `xsmall` size when you add your changes to the change log.

lukeschunk: sounds good :)

cerinman:
:shipit:

PR #407, opened 2016-08-16
Stripped out the isMobile check for now

redperk:
### Issue
The `isMobile` check was messing up the dropdown in the `Select.js` component.

### Solution
I ripped it out of the code for now, so we are implementing our own styled dropdown instead of allowing the device OS to display it's own dropdown. The `Select.js` component is being used in many places so eventually we will want to come back and address this in the future.


redperk:
@cerinman This is ready if you could take a look.

tumentumurchudur:
:shipit:

PR #408, opened 2016-08-16
Reposition the Drawer when the window is resized

### LINKED ISSUE #361: Drawer component should re-render on window resize
Steps to reproduce the issue:
* Have the Drawer component visible
* Resize the page
* Notice that the component doesn't update it's position

jtanner:
When the window was resized the Drawer would stay in the same position until closing and reopening. This fixes the problem by re-animating the Drawer, but with a duration of 0, which turned out to be the cleanest solution.

ezgif com-video-to-gif 2

Fixes #361 
Refs #363


paniclater:
:shipit:

PR #409, opened 2016-08-16
Allow for different Drawer sizes

jtanner:
I.e. The half-drawer

Allow customization of the size of a Drawer and at different breakpoints.

This PR will:
* expose breakPoints and maxWidth props
* use StyleConstants.BreakPoints for default breakPoints
* use breakPoints for media queries

ezgif com-video-to-gif 1

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -356,6 +356,18 @@ const Demo = React.createClass({
     });
   },
 
+  _handleShowSmallDrawer () {
+    this.setState({
+      showSmallDrawer: true
+    });
+  },
+
+  _handleHideSmallDrawer () {
+    this.setState({
+      showSmallDrawer: false
+    });
+  },
+
   _handleSpinnerIconOnlyClick () {
     this.setState({
       spinnerIconOnlyIsActive: !this.state.spinnerIconOnlyIsActive
@@ -535,6 +547,18 @@ const Demo = React.createClass({
               title='This is the drawer component'
             >
               <div style={{ padding: 20, fontFamily: 'Helvetica, Arial, sans-serif' }}>Insert Custom Content Here</div>
+              <Button onClick={this._handleShowSmallDrawer}>
+                Toggle Small Drawer
+              </Button>
+              {this.state.showSmallDrawer ? (
+                <Drawer
+                  breakPoints={{ large: 750, medium: 500, small: 320 }}
+                  maxWidth={480}
+                  onClose={this._handleHideSmallDrawer}
+                  showScrim={false}
+                  title='Small Drawer'
+                />
+              ) : null}
             </Drawer>
           </div>
         ) : null}
--- a/src/components/Drawer.js
+++ b/src/components/Drawer.js
@@ -50,15 +50,13 @@ const Drawer = React.createClass({
       return this.props.animateLeftDistance + '%';
     }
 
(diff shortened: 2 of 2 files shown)


demo/app.js
```diff
@@ -535,6 +547,18 @@ const Demo = React.createClass({
               title='This is the drawer component'
             >
               <div style={{ padding: 20, fontFamily: 'Helvetica, Arial, sans-serif' }}>Insert Custom Content Here</div>
+              <Button onClick={this._handleShowSmallDrawer}>
+                Toggle Small Drawer
+              </Button>
+              {this.state.showSmallDrawer ? (
+                <Drawer
+                  breakPoints={{ large: 750, medium: 500, small: 320 }}
```
jtanner: See [src/constants/Style.js#L2](https://github.com/mxenabled/mx-react-components/blob/0a62871140adbb6f8364025e67d4fe0c45791882/src/constants/Style.js#L2) for the defaults.

src/components/Drawer.js
```diff
@@ -50,20 +58,19 @@ const Drawer = React.createClass({
...
     const windowWidth = window.innerWidth;
 
-    if (windowWidth >= maxResolutionBreakPoint) {
-      //Resolution - 960 from the left
-      return windowWidth - greaterThan1200ComponentWidth;
-    } else if (windowWidth <= minResoultionBreakPoint) {
+    if (windowWidth >= this.props.breakPoints.large) {
+      //Resolution - maxWidth
+      return windowWidth - this.props.maxWidth;
+    } else if (windowWidth <= this.props.breakPoints.medium) {
```
paniclater: definitely better names here 👍

src/components/Drawer.js
```diff
@@ -152,11 +159,11 @@ const Drawer = React.createClass({
         backgroundColor: StyleConstants.Colors.PORCELAIN,
         boxShadow: StyleConstants.ShadowHigh,
 
-        '@media (max-width: 750px)': {
+        [`@media (max-width: ${this.props.breakPoints.medium}px)`]: {
```
paniclater: I haven't seen this bracketed notation style yet.

iheartrachie: I'm not sure if Ryan is talking about the square brackets or the curly brackets. Why are the square brackets needed?

cerinman: He is doing string interpolation here so he has to surround the string in brackets in order for the style object to behave correctly.  At least that is how I interpret this.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

iheartrachie: I made the realization of this ^ on the drive home (second best place to do thinking). I like using string interpolation and I think this is a really creative implementation of it!

iheartrachie:
:shipit:

src/components/Drawer.js
jtanner: Thank you @iheartrachie! The bracket syntax is called _Computed property names_ and is new in ES2015.

See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#Computed_property_names

We're doing something similar in raja, take a look at `Icon` in `serenity/constants/Category.js`.

PR #412, opened 2016-08-17
Drawer.close needs to return a Promise like it says in the comment :)

jtanner:
#405 made `close` public, but it didn't return a `Promise` like I said it would in the docs. This fixes that discrepancy.


redperk:
:shipit:

PR #417, opened 2016-08-19
Allow DisplayInput to be "focused" programmatically

jtanner:
We need a way to manually focus the `DisplayInput` for custom inputs to match the behavior of other `DisplayInput`s.

Does this seem like a reasonable approach?

__Here's an example use case__ (I was using these changes for the screen shot):


paniclater:
:shipit:

PR #418, opened 2016-08-22
React 15.3.1 update

cerinman:
The new version of React that was just released makes several perf improvements for development builds. We would have picked up the new version with out bumping the version number in the package.json file but I wanted to make it implicit that we upgraded.

There are still a few warnings that we have from the 15.3.0 update that removed the white list for filtering out non HTML attributes when spreading props over an HTML element that we should address in individual PRs as time allows.

https://github.com/facebook/react/releases/tag/v15.3.1


jtanner:
:shipit:

PR #422, opened 2016-08-23
SearchInput: Adds focusOnLoad prop

tumentumurchudur:
This PR allows users to focus their cursor in the text field when `SimpleInput` and `SearchInput` components load through `focusOnLoad` prop.


cerinman:
:shipit:

PR #423, opened 2016-08-24
Adds documentation to PR422

tumentumurchudur:
- Adds docs to PR422 which allows users to focus their cursor in the text field on load in the SimpleInput and SearchInput components. I skipped the version `4.4.36` trying to publish before pushing it. Sorry about that.

### DIFF
--- a/components/Changelog.js
+++ b/components/Changelog.js
@@ -5,7 +5,12 @@ const Changelog = React.createClass({
     return (
       <div>
         <h1>Change Log</h1>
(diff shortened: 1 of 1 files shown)


paniclater:
Does the demo need to be changed at all?

tumentumurchudur:
No, this PR just adds a prop that puts the cursor in the input field on load.

paniclater:
:shipit:

PR #426, opened 2016-09-05
Spreading props fix for warning introduced in React 15.0

cerinman:
### Issue

React now warns when applying properties to html elements that should not be there according to HTML standards.  We spread props across inputs in a few places so we were triggering the warning

See https://fb.me/react-unknown-prop for details

### Solution

This PR introduces a new prop for the components effected.

elementProps: Type: `Object` Default: `{}`

We now use the above prop to pass the properties that we want to apply to the HTML element within the component.  This allows us to separate this properties from the component properties and prevent the warning.  

Components effected

- DisplayInput
- Icon
- RajaIcon
- SimpleInput

I've also updated the DisplayInput in the demo app.  It was the only component there that needed it's usage updated.  Please refer to it on how to handle internally as we move forward with this change.

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -1019,12 +1019,14 @@ const Demo = React.createClass({
 
         <br /><br />
         <DisplayInput
+          elementProps={{
+            onBlur: this._handleInputStatusMessage,
+            onFocus: this._handleInputFocus,
+            onMouseOut: this._handleInputHideHint,
+            onMouseOver: this._handleInputShowHint
+          }}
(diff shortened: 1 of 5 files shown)


lukeschunk:
:shipit: 

also want to post this: https://medium.com/@mbostock/prereleases-and-npm-e778fc5e2420#.7ktks31xq i remember when I researched release candidates there are a couple potential gotchas

jtanner:
@cerinman thank you for taking care of these warnings! 🍪

PR #427, opened 2016-09-05
Fixes style errors in demo app

cerinman:
Fixes letter spacing number as string and padding short hand errors while running the demo app.


lukeschunk:
:shipit:

PR #429, opened 2016-09-06
Updated the SimpleInput input reference to new style.

phantomxc:
There was an issue trying to reference the input ref before it was made available. Updated to the recommended callback approach: https://facebook.github.io/react/docs/more-about-refs.html


cerinman:
:shipit:

PR #430, opened 2016-09-08
Aphrodite- Drawer Upgrade

cerinman:
https://github.com/Khan/aphrodite
### Issue

Older versions of Radium are causing an `Unknown Prop` warnings with React 15.2.x +.  The 0.18.x versions of Radium resolve the issue but they now require that you wrap your application in its `StyleRoot` component if you use media queries.  

This presents an issue as we don't want users of `mx-react-components` to have to require in Radium and wrap their application in a `StyleRoot` component.  We could wrap the effected components in the StyleRoot component our selves but it looks like the `StyleRoot` component for Radium is injecting CSS style sheets similar to how `Aphrodite` works but in a way that seems more cumbersome.
### Solution

I decided to make this issue a reason to test out Aphrodite.  This PR converts the Drawer component to Aphrodite as a test.  All feed back is welcome.
### Gotchas
- Combining internal style objects with styles from props must now happen in the `StyleSheet.create` call in the styles function.  The resulting javascript object from `StyleSheet.create` prevents us from doing `Object.assign` to merge styles directly in render functions.
- We need to use the no-important version of Aphrodite so that it does not mark all our styles as `!important`.  `const { StyleSheet, css } = require('aphrodite/no-important');`  From the Aphrodite git hub page: `By default, Aphrodite will append !important to style definitions. This is intended to make integrating with a pre-existing codebase easier. If you'd like to avoid this behaviour, then instead of importing aphrodite, import aphrodite/no-important. Otherwise, usage is the same:`
- The injecting of the css style sheets seems to add a very small amount of time on load and because of this I had to delay the drawer animation by 50 ms so that it would run smoothly.  This was not a big deal since Velocity provides the `delay` option.  The increase is also not noticeable by the human eye.  It just seems to be enough to cause a race condition with the initial animation.
- Now that we have a proper style sheet, we have to be aware of nesting html elements that css does not support.  The example here is that I had to move from using the `<p>` tag to using a div because I was getting css errors for the demo app about nesting divs inside `<p>` tags.

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -551,17 +551,17 @@ const Demo = React.createClass({
               ref='drawer'
               title='This is the drawer component'
             >
-              <p style={{ padding: 20, fontFamily: 'Helvetica, Arial, sans-serif' }}>
+              <div style={{ padding: 20, fontFamily: 'Helvetica, Arial, sans-serif' }}>
                 Insert Custom Content Here
-              </p>
+              </div>
 
-              <p style={{ padding: 10 }}>
+              <div style={{ padding: 10 }}>
                 <Button onClick={this._handleCloseDrawerClick}>
                   Close Drawer
                 </Button>
-              </p>
+              </div>
 
-              <p style={{ padding: 10 }}>
+              <div style={{ padding: 10 }}>
                 <Button onClick={this._handleShowSmallDrawer}>
                   Toggle Small Drawer
                 </Button>
@@ -574,7 +574,7 @@ const Demo = React.createClass({
                     title='Small Drawer'
                   />
                 ) : null}
-              </p>
+              </div>
             </Drawer>
           </div>
         ) : null}
--- a/src/components/Drawer.js
+++ b/src/components/Drawer.js
@@ -108,6 +108,8 @@ const Drawer = React.createClass({
   },
 
   _renderNav () {
+    const styles = this.styles();
+
     return this.props.navConfig ? (
       <nav className={css(styles.nav)}>
         <Button
@@ -130,33 +132,14 @@ const Drawer = React.createClass({
   },
 
   render () {
-    const dynamicStyles = StyleSheet.create({
-      backArrow: {
-        [`@media (max-width: ${this.props.breakPoints.medium}px)`]: {
-          paddingLeft: 10
-        }
-      },
-      component: Object.assign({}, {
-        [`@media (max-width: ${this.props.breakPoints.medium}px)`]: {
-          width: '100%'
-        },
-        [`@media (min-width: ${this.props.breakPoints.large}px)`]: {
-          width: this.props.maxWidth
-        }
-      }, this.props.style),
-      content: this.props.contentStyle,
-      header: this.props.headerStyle,
-      scrim: {
-        backgroundColor: this.props.showScrim ? StyleConstants.Colors.SCRIM : 'transparent'
-      }
-    });
+    const styles = this.styles();
 
     return (
       <div className={css(styles.componentWrapper)}>
-        <div className={css(styles.scrim, dynamicStyles.scrim)} onClick={this.close} />
-        <div className={css(styles.component, dynamicStyles.component)} ref={(ref) => (this._component = ref)}>
-          <header className={css(styles.header, dynamicStyles.header)}>
-            <span className={css(styles.backArrow, dynamicStyles.backArrow)}>
+        <div className={css(styles.scrim)} onClick={this.close} />
+        <div className={css(styles.component)} ref={(ref) => (this._component = ref)}>
+          <header className={css(styles.header)}>
+            <span className={css(styles.backArrow)}>
               <Button
                 icon='arrow-left'
                 onClick={this.close}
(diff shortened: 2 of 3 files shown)


src/components/Drawer.js
```diff
@@ -99,6 +93,7 @@ const Drawer = React.createClass({
   _animateComponent (transition, extraOptions) {
     const el = this._component;
     const options = Object.assign({
+      delay: 50,
```
paniclater: Is this going to need to be longer in wrapped web views or mobile browsers with slower javascript execution?

src/components/Drawer.js
```diff
@@ -131,31 +126,31 @@ const Drawer = React.createClass({
           type='base'
         />
       </nav>
-    ) : <div style={styles.nav} />;
+    ) : <div className={css(styles.nav)} />;
   },
 
   render () {
     const styles = this.styles();
```
paniclater: Is it possible to run this in `componentWillMount` or `componentDidMount` and make it available on state?

src/components/Drawer.js
cerinman: Valid question.  Not sure but I would guess yes.  We might have to do some testing to dial it in.  I can increase it to 100 ms if we feel a longer time would be better initially.

src/components/Drawer.js
paniclater: Oh, I see, because we're dynamically doing stuff in the render we can't modify it because the stylesheet is already created.

redperk:
Will users have to require in Aphrodite now?

src/components/Drawer.js
paniclater: Actually it looks like in this case that wouldn't be an issue, and it may be possible to use inline styles in combination with Aphrodite style sheets for the dynamic values we do need?

cerinman: Correct

cerinman:
@redperk this will not force the user to require Aphrodite.  This will be an internal dependency in `mx-react-components` just like Radium currently is.

src/components/Drawer.js
cerinman: You could use inline styles and Aphrodite together if needed yes.

src/components/Drawer.js
```diff
@@ -164,8 +159,8 @@ const Drawer = React.createClass({
   },
 
   styles () {
-    return {
-      component: {
+    return StyleSheet.create({
```
paniclater: If we go this route I would rather pass in props as arguments:

```
//function execution
const styles = this.styles(this.props.style, this.props.contentStyle, this.props.headerStyle);

//function definition
styles (style, contentStyle, headerStyle) {
  component: Object.assign({}, {
  }, style),
//etc
}
```

Just personal preference and making functions take in arguments instead of referencing `this` properties often makes them easier to test if we ever needed / wanted to.

src/components/Drawer.js
cerinman: We could also probably put `styles` on state and update it on incoming prop/state change if we so desired.  I can't think of a reason that wouldn't work.

paniclater: I'd like to see if creating the stylesheet on `componentWillMount` and then mixing inline styles when we need dynamic values would reduce / eliminate the need for the delay animation property.

redperk:
I agree about Radium. I don't love the idea of wrapping the app in a styleRoot component or forcing users to require radium.

src/components/Drawer.js
cerinman: I agree that it would make testing easier but then we have to ensure argument order unless we plan to pass in an object but even then it requires the style function know about at least a key name.  I consider this at one point and decided that an internal private function knowing directly about props wasn't a bad thing.

src/components/Drawer.js
cerinman: Let me play around and see what I can come up with.  I'll have you come over and check it out after I get something going.

cerinman: https://github.com/Khan/aphrodite#style-injection-and-buffering was the reasoning for doing the delay.

paniclater: Cool. I think we can use this technique as well to conditionally generate our styles. Would require more of a shift away from how we're used to doing things.

https://github.com/Khan/aphrodite#conditionally-applying-styles

lukeschunk:
i'm a little worried about the injecting of stylesheets increasing loadtime - will this increase if we had more widespread use of aphrodite?

demo/app.js
```diff
@@ -551,17 +551,17 @@ const Demo = React.createClass({
               ref='drawer'
               title='This is the drawer component'
             >
-              <p style={{ padding: 20, fontFamily: 'Helvetica, Arial, sans-serif' }}>
+              <div style={{ padding: 20, fontFamily: 'Helvetica, Arial, sans-serif' }}>
```
lukeschunk: just so i'm understanding - this could have technically still been a `<p>` tag, but the ones below it would not have worked so you changed for consistency/DRYness

cerinman:
I should have chosen my words better.  It doesn't really increase load time.  It's more of a buffer to minimize DOM modifications.  This causes an issue with stuff happening in `componentDidMount` as not all styles have been applied yet. https://github.com/Khan/aphrodite#style-injection-and-buffering

demo/app.js
cerinman: CSS expects certain elements when traversing the DOM tree so that you can do things like `div -> header -> a` to style all links within any header inside a div.  Because of this, it freaks out if you nest elements inside another element that are not supported.  Divs inside of  P tags happen to be one of those.  Since Button is a div, you can't place it in a p tag.  The easiest solution was to just make the p tags divs in this case.

src/components/Drawer.js
```diff
@@ -131,31 +126,31 @@ const Drawer = React.createClass({
...
               />
             </span>
-            <span style={styles.title}>
+            <span className={css(styles.title)}>
               {this.props.title}
             </span>
             {this._renderNav()}
           </header>
-          <div style={Object.assign({}, styles.content, this.props.contentStyle)}>
+          <div className={css(styles.content)}>
```
jtanner: 👍  I like that you moved the concern of merging this and the other style overrides into `styles`, I think it makes the jsx cleaner and when looking at the styles it's apparent which can be overridden.

src/components/Drawer.js
cerinman: I've played around a bit now and here are some thoughts.

#### Mixing inline styles and Aphrodite: 

I was not able to get this to work 100%.  There was some cognitive over head around thinking through how inline styles over write styles in the Aphrodite injected style sheet.  I'm sure we could probably get it to work with some more effort but more complex components might end up being kinda of weird to follow.

#### Dynamic/Static styles all with Aphrodite

This works great with no issues.  The only down side to this was that we had two or more sets of styles.  While not terrible, I like having all the styles for a component together in one location but that is personal preference.  This still requires a delay or setTimeout.

#### Leave as is with a small change to using setTimeout

In the end I still like this approach best.  I've moved to using setTimeout over delay via Velocity and I've also used zero for the time out delay.  I looked into doing a decorator as Cameron suggested but I was coming up short with a good way to do it and in the end it seemed over kill.  I'm open to sitting down with Cameron next week for an education on decorators to see if we can come up with something better than I could.

cerinman:
While we debate this and work through tweaks and changes, I'm going to put up another PR to revert the upgrade to Radium so that I can get my internal PR moving up.  We can live with the Radium unknown prop warning for a while longer.

cerinman:
This has seemed to die.  Closing. We can reopen later if needed.

PR #433, opened 2016-09-12
BarChart: Fixes scaling issue

tumentumurchudur:
This PR fixes an issue with bars in the BarChart not scaling data values correctly. For example: A data value 1200 should be twice as tall as a data value 600 in the bar chart. This was caused by calculating the `domain` and `range` values on the assumption that all data passed as a prop to this component are positive.

Before
screen shot 2016-09-12 at 3 59 38 pm

After
screen shot 2016-09-12 at 3 59 16 pm

### DIFF
--- a/src/components/BarChart.js
+++ b/src/components/BarChart.js
@@ -211,11 +211,26 @@ const BarChart = React.createClass({
       return d.value;
     });
     const gap = 0.3;
-    const y0 = Math.max(Math.abs(d3.min(data)), Math.abs(d3.max(data)));
+    const hasNegative = !!data.filter(value => value < 0).length;
+    const hasPositive = !!data.filter(value => value > 0).length;
+    let domain;
+    let range;
 
+    if (hasNegative && !hasPositive) {
+      domain = [d3.min(data), 0];
+      range = [-height, 0];
+    } else if (!hasNegative && hasPositive) {
+      domain = [0, d3.max(data)];
+      range = [0, height];
+    } else if (hasNegative && hasPositive) {
+      const y0 = Math.max(Math.abs(d3.min(data)), Math.abs(d3.max(data)));
+
+      domain = [-y0, y0];
+      range = [-height, height];
+    }
     const yScale = d3.scale.linear()
-      .domain([-y0, y0])
-      .range([0, height]);
+      .domain(domain)
+      .range(range);
 
     const xScale = d3.scale.ordinal()
       .domain(d3.range(this.props.data.length))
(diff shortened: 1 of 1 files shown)


src/components/BarChart.js
```diff
@@ -211,11 +211,26 @@ const BarChart = React.createClass({
...
+      domain = [d3.min(data), 0];
+      range = [-height, 0];
+    } else if (!hasNegative && hasPositive) {
+      domain = [0, d3.max(data)];
+      range = [0, height];
+    } else if (hasNegative && hasPositive) {
+      const y0 = Math.max(Math.abs(d3.min(data)), Math.abs(d3.max(data)));
+
+      domain = [-y0, y0];
+      range = [-height, height];
```
jtanner: Does doubling the range cause the height of the component to grow larger than the `height` provided in props?

src/components/BarChart.js
```diff
@@ -211,11 +211,26 @@ const BarChart = React.createClass({
...
+    if (hasNegative && !hasPositive) {
+      domain = [d3.min(data), 0];
+      range = [-height, 0];
+    } else if (!hasNegative && hasPositive) {
+      domain = [0, d3.max(data)];
+      range = [0, height];
+    } else if (hasNegative && hasPositive) {
+      const y0 = Math.max(Math.abs(d3.min(data)), Math.abs(d3.max(data)));
+
+      domain = [-y0, y0];
```
jtanner: I'm curious, could this be accomplished with the following?
```javascript
domain = [d3.min(data), d3.max(data)]
range = [0, height]
```

src/components/BarChart.js
tumentumurchudur: Yes, that is a good question. It should be `range=[-height * 0.5, height * 0.5]` but I was having some problems. Let me look into it some more and report back. 👍

src/components/BarChart.js
tumentumurchudur: no, it is little more complicated than that. When we have positive and negative values, the chart needs to have the negative values going downward and the positive values going upward. So to make it work, I have two bar charts like this. The range for the bar chart positioned upside down needs to have negative values in the range.

screen shot 2016-09-13 at 10 19 57 am

jtanner:
:shipit:

tumentumurchudur:
@jtanner fixed the issue with height doubling for a data array that has both negative and positive numbers.

All positive 
screen shot 2016-09-13 at 12 59 32 pm

Both positive and negative
screen shot 2016-09-13 at 1 00 11 pm

All negative
screen shot 2016-09-13 at 1 00 58 pm

PR #434, opened 2016-09-13
Fixes SimpleInput usage in SearchInput component

cerinman:
I missed this one when I was fixing unknown props before because it was required in as `Input` rather than `SimpleInput` so when I searched for `<SimpleInput` I missed this one.


tumentumurchudur:
:shipit:

tumentumurchudur:
:shipit:

PR #432, opened 2016-09-12
Wengert/docs

phantomxc:
Move the docs into the main repository to utilize the new docs folder serving from github. This will have to be turned on manually by an Admin. https://help.github.com/articles/configuring-a-publishing-source-for-github-pages/

- Removed the demo directory as you can now develop against the docs with hot loading! Thanks @jtanner for simplifying what I was doing! 🍪 
- Added a simple script for updating the commit message for doc updates `npm run docs:release`
- Updated the contribution page to reflect the docs changes

PR #435, opened 2016-09-14
Adds new component TextArea

tumentumurchudur:
This PR adds a new component `TextArea`.

## Props
- elementProps: object - allows the parent component to pass props to `textarea` such as onChange, onBlur, placeholder, etc.
- primaryColor: string - border color of the component when it has focus
- rows: number - determines the number of rows
- valid: bool - determines whether the value in the component is valid

screen shot 2016-09-14 at 9 51 03 am


jtanner:
Once you take care of the conflict (demo/app.js was removed in #432) I'll give this a 🚢 .

tumentumurchudur:
Added documentation. Ready for review.

screen shot 2016-09-22 at 4 04 01 pm

src/components/TextArea.js
```diff
@@ -54,7 +54,9 @@ const TextArea = React.createClass({
       >
         <textarea
           {...elementProps}
-          ref='textarea'
+          ref={ref => {
+            this.textarea = ref;
+          }}
```
jtanner: What's the advantage of doing it this way instead of `this.refs.textarea`?

jtanner:
Besides my comment I think this is ready to :ship:

src/components/TextArea.js
tumentumurchudur: `this.refs.textarea` no longer works and throws a console error. It may have to do with the latest version of React.

jtanner:
:shipit: ⚡

PR #437, opened 2016-09-14
Modal: Fixes close icon

tumentumurchudur:
This PR fixes an issue where the close icon does not work in the `Modal` component.

- Pass `onClick` prop to the `Icon` component as a property in the `elementProps` object

### DIFF
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -190,7 +190,9 @@ const Modal = React.createClass({
           {this.props.showCloseIcon ? (
             <Icon
               className='mx-modal-close'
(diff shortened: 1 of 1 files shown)


jtanner:
I think this fix looks good :shipit:

But it made me wonder if there were other instances where `Icon` didn't get updated to use the new `elementProps` pattern that @cerinman added so I did some digging and found that `Icon` still needs to be updated in the following files:

src/components/Calendar.js
src/components/DatePickerFullScreen.js
src/components/TypeAhead.js

I created #438 to keep track of the rest of the components that need to be updated.

PR #440, opened 2016-09-19
Luke/date picker element props

lukeschunk:
Updated Icons with element Props so click handlers work. Also changed cursor style to `pointer` in a couple of places.


cerinman:
:shipit:

Issue #442, opened 2016-09-19
Search Input / Polish


derek-boman:
A few items to tighten up on the Search Input component (or could this just use the Simple Input with a couple additional props?)

![image](https://cloud.githubusercontent.com/assets/17180771/18652125/fbb4d766-7e8c-11e6-8582-c34d5af6eaf5.png)

- [ ] Reset field icon on the right side. `icon: close-solid, fog.
- [ ] Set default font size to Medium
- [ ] Typing text is currently rgb(0,0,0) this should be Charcoal
- [ ] increase `padding-right` on the Search icon from 5px to 7px

lukeschunk:
Do you want the reset field icon on all simple inputs or just the search input. The search input uses the simple input, and wraps it with a couple of additional things.

derek-boman:
I think this is done. Closing.

PR #443, opened 2016-09-20
Search Input - reset icon & other style fixes 

lukeschunk:
Addresses https://github.com/mxenabled/mx-react-components/issues/442 

I was kind of 50/50 on whether to just kill the SearchInput and add more props to the SimpleInput to make this happen, but the reason I decided to keep it is I think it will make a faster dev experience for us && have less potential to break things in existing SimpleInputs.  

searchinput gif


jtanner:
:shipit:

PR #444, opened 2016-09-20
FileUpload: Add image dimension validation

jmophoto:
This adds validation for image dimensions. It also removes Radium, moves the styles to a method and cleans up any style array.

- Add prop object for imageValidation with attributes for exact, max, min and ratio width and height
- Remove array as prop type for style
- Validate file on mount and when receiving new prop
- Add a callback to the `readFile` method that returns the file, height and width
- Add method to validate the image
- Clean-up styles


lukeschunk:
:shipit:

PR #447, opened 2016-09-21
FileUpload: reset input on new props

jmophoto:
This fixes an issue where using the file select after clearing the previous file does not work. When new props are received, set the input value to null, resetting the form and allowing for another file to be selected.

### DIFF
--- a/src/components/FileUpload.js
+++ b/src/components/FileUpload.js
@@ -42,6 +42,8 @@ const FileUpload = React.createClass({
   },
 
   componentWillReceiveProps (newProps) {
+    this._input.value = null;
+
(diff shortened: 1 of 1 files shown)


redperk:
:shipit:

src/components/FileUpload.js
```diff
@@ -42,6 +42,8 @@ const FileUpload = React.createClass({
   },
 
   componentWillReceiveProps (newProps) {
+    this._input.value = null;
```
cerinman: I've been getting warnings in the console when using null for input values.  We might want to use an empty string in this case instead.

cerinman:
:shipit:

PR #449, opened 2016-09-21
Fix tooltip elemementprops

jmophoto:
This moves the mouse event props on the tooltip icon and the tooltip in the modal to the new `elementProps`.


redperk:
:shipit:

PR #451, opened 2016-09-26
Luke/search input update

lukeschunk:
I implemented the new `x` icon to the search input and i wanted to put in a check to not display the X if there isn't a function passed down; this way it won't break things where the `handleResetClick` hasn't been implemented. 

Also going to go through internally and implement `handleResetClick` in the app.

searchclear_gif


lukeschunk:
also @cerinman you mentioned checking if it is a function as well, but i don't think that's necessary since the prop type is already checking for that and will throw a warning if it's something else - lmk what you think

redperk:
:shipit:

PR #453, opened 2016-09-29
handle next click

lukeschunk:
When you click on the forward arrow you don't want to go back in time. 

giphy-38


jmophoto:
:shipit:

timebandits_primary

Issue #452, opened 2016-09-29
Navigating in Docs causes Error 


lukeschunk:
The `Only a react owner can have refs` error sometimes occurs when trying to navigate between components in the docs. 

Haven't determined exactly what puts the app in this state, but click around on a few of the nav links and the error will occur. 

#### Example:
![osnavgif](https://cloud.githubusercontent.com/assets/13579578/18938588/e7c44864-85b6-11e6-9b7b-c7091ad336f3.gif)

PR #454, opened 2016-09-30
Luke/range selector

lukeschunk:
Fixes https://github.com/mxenabled/mx-react-components/issues/452 

Fixes breaking bug in range selector. The refs API was changed in React 15 https://facebook.github.io/react/blog/2016/04/07/react-v15.html. So I changed to a current accepted pattern.


jtanner:
:shipit:

PR #457, opened 2016-10-04
Drawer docs and new props

redperk:
### Issue
The monthly contribution drawer needs to stay open if the scrim is clicked and I needed to remove the close button from the drawer, so I added 2 new props.

- closeOnScrimClick: this defaults to true, but if you set it to false the drawer will stay open if the scrim is clicked

- showCloseButton: this also defaults to true, but if you set it to false the button will be hidden

Since the drawer component doesn't exist in the docs I went ahead and added it. I'm pretty confident about most of the explanations, but I definitely would love some extra eyes to check coherency, grammers and spells.

I played around with the breakPoints prop and I just don't feel like I can explain that prop in the docs so I will need some help there. breakpoints takes in a `small` property but it doesn't appear to even be used in the code.

@paniclater @cerinman

### DIFF
--- a/docs/components/DrawerDocs.js
+++ b/docs/components/DrawerDocs.js
@@ -8,7 +8,9 @@ const DrawerDocs = React.createClass({
 
   getInitialState () {
     return {
-      demoDrawerOpen: false
+      demoDrawerOpen: false,
+      currentPage: 3,
+      totalPages: 8
     };
   },
 
@@ -24,12 +26,30 @@ const DrawerDocs = React.createClass({
     });
   },
 
+  _handlePreviousClick () {
+    if (this.state.currentPage > 1) {
+      this.setState({
+        currentPage: this.state.currentPage - 1
+      });
+    }
+  },
+
+  _handleNextClick () {
+    if (this.state.currentPage < this.state.totalPages) {
+      this.setState({
+        currentPage: this.state.currentPage + 1
+      });
+    }
+  },
+
   _renderDrawer () {
     const styles = this.styles();
 
     return (
       <Drawer
+        breakPoints={{ large: 1200, medium: 1100, small: 1320 }}
         contentStyle={styles.content}
+        navConfig={{ label: this.state.currentPage + ' of ' + this.state.totalPages, onNextClick: this._handleNextClick, onPreviousClick: this._handlePreviousClick }}
         onClose={this._handleDrawerClose}
         title='Demo Drawer'
       >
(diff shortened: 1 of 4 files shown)


docs/components/DrawerDocs.js
```diff
@@ -0,0 +1,159 @@
...
+      });
+    }
+  },
+
+  _renderDrawer () {
+    const styles = this.styles();
+
+    return (
+      <Drawer
+        breakPoints={{ large: 1200, medium: 1100, small: 1320 }}
```
jtanner: @redperk I looked at the code in the `Drawer` component and you're right, the small breakpoint is not used. I think you should remove it from this example.
```javascript
breakPoints={{ large: 1200, medium: 1100 }}
```

docs/components/DrawerDocs.js
```diff
@@ -0,0 +1,159 @@
...
+          Demo Drawer
+        </Button>
+        {this.state.demoDrawerOpen && this._renderDrawer()}
+
+        <h3>Usage</h3>
+        <h5>animateLeftDistance<label>Number</label></h5>
+        <p>This number represents the percent of the screen visible between the left edge of the screen and the left edge of the drawer.</p>
+
+        <h5>breakPoints<label>Object</label></h5>
+        <p></p>
```
jtanner: Here is a rough description of the breakpoints:

_When the screen size is:_
__`>=` large:__ drawer width is `maxWidth`
__`<` large and `>` medium:__ drawer width is 20% from left side
__`<=` medium:__ drawer width is 100%

redperk:
@jtanner @paniclater @cerinman I added Joe's suggestions for the breakpoints in the docs. Thanks Joe! This should be ready to go now.

docs/components/DrawerDocs.js
```diff
@@ -78,7 +80,16 @@ const DrawerDocs = React.createClass({
...
         <h5>breakPoints<label>Object</label></h5>
-        <p></p>
+        <p>This object takes 2 properties:</p>
+        <ul style={styles.unorderdLists}>
+          <li style={styles.listItem}><h5 style={styles.listItem}>large <label>Number</label></h5></li>
+          <li style={styles.listItem}><h5 style={styles.listItem}>medium <label>Number</label></h5></li>
+        </ul>
+        <p>When the screen size is: <br />
+          &#62;&#61; large: drawer width is maxWidth <br />
+          &#60;	large and > medium: drawer width is 20% from left side <br />
```
jtanner: Looks like you need to change `>` to `&#62;`. Other than that I think this is 👍  to 🚢 .

redperk: Good catch!

jtanner:
:shipit:

PR #458, opened 2016-10-11
changelog

lukeschunk:
Was about to deploy the OS  and needed to get the changelog update reviewed. 

As a side note i think the work flow for updating the changelog is a little weird, because you either need to update right as you're merging in, or add to a non-changelog PR - either way it's a little weird to guess what version you'll be releasing

### DIFF
--- a/docs/components/Changelog.js
+++ b/docs/components/Changelog.js
@@ -5,6 +5,18 @@ const Changelog = React.createClass({
     return (
       <div>
         <h1>Change Log</h1>
+        <h3>Release Canidate 5.0.0-rc.15</h3>
+        <ul>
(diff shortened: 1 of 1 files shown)


docs/components/Changelog.js
```diff
@@ -5,6 +5,18 @@ const Changelog = React.createClass({
     return (
       <div>
         <h1>Change Log</h1>
+        <h3>Release Canidate 5.0.0-rc.15</h3>
```
redperk: Candidate

redperk:
:shipit:

redperk:
Good catch on all the previous typos

PR #459, opened 2016-10-12
Added styles props to drawer and textarea

redperk:
This PR follows the pattern set in the SimpleInput component. Merging all the local styles with the styles prop that is passed in. This gives the user of the OS component access to all the CSS.


jmophoto:
:shipit: with a bagel.

jtanner:
:shipit:

PR #461, opened 2016-10-25
Update the medium breakpoint to 768

phantomxc:
After speaking with @cerinman @derek-boman, we've opted to update the medium breakpoint to 768px.


jtanner:
:shipit:

PR #462, opened 2016-10-31
Expose D3 components

cerinman:
This PR exposes the D3 components in the d3 folder so that I can use them internally for building out a new chart.  No doc updates for these components as we don't intend to broadcast that they are available at this time.  I'm sure we will at some point but that is low priority.

I also added a `circleColor` prop to the CirclesGroup component.


phantomxc:
:shipit:

PR #466, opened 2016-11-07
Fancy new slider component

redperk:
This PR adds a new slider component and docs to our OS components.

ezgif com-video-to-gif 7

### DIFF
--- a/docs/components/SimpleSliderDocs.js
+++ b/docs/components/SimpleSliderDocs.js
@@ -26,17 +26,18 @@ const SimpleSliderDocs = React.createClass({
         </h1>
 
         <h3>Demo</h3>
-        <SimpleSlider onPercentChange={this._handleSliderChange} />
+        <div>{this.state.percent}</div>
+        <SimpleSlider onPercentChange={this._handleSliderChange} percent={this.state.percent} />
 
         <h3>Usage</h3>
-        <h5>defaultValue <label>Number</label></h5>
-        <p>This prop moves the slider to the assigned postion. It is passed in as a number between 0 and 1. So 0.75 would put the slider at 75% of the line.</p>
+        <h5>percent <label>Number</label> <em>Required</em></h5>
+        <p>This prop moves the slider to the assigned postion. It is passed in as a number between 0 and 1. So 0.75 would put the slider at 75% of the total distance.</p>
 
         <h5>disabled <label>bool</label></h5>
-        <p>When true the slider is disabled. The default is false.</p>
+        <p>When true, the slider is disabled. This defaults to false.</p>
 
         <h5>onPercentChange <label>func</label></h5>
-        <p>When the slider is moved your callback function is called and the current percent is passed as a param. The current percent is a number between 0 and 1.</p>
+        <p>When the slider is moved, your callback function is called and the current percent is passed as a param. The current percent is a number between 0 and 1.</p>
 
         <h5>selectedColor <label>string</label></h5>
         <p>This sets the slider bar color.</p>
@@ -46,6 +47,7 @@ const SimpleSliderDocs = React.createClass({
         <Markdown>
           {`
             <SimpleSlider
+              percent={this.state.percent}
               selectedColor='#359BCF'
               onPercentChange={this._myCallbackFunction}
             />
--- a/docs/components/SimpleSliderDocs.js
+++ b/docs/components/SimpleSliderDocs.js
@@ -53,8 +53,6 @@ const SimpleSliderDocs = React.createClass({
             />
           `}
         </Markdown>
-
-
       </div>
     );
   }
--- a/docs/components/SimpleSliderDocs.js
+++ b/docs/components/SimpleSliderDocs.js
@@ -1,0 +1,39 @@
+const React = require('react');
+
+const { SimpleSlider } = require('mx-react-components');
+
+const Markdown = require('components/Markdown');
+
+const SimpleSliderDocs = React.createClass({
+  render () {
+    return (
+      <div>
+        <h1>
+          Simple Slider
+          <label>A simple slider bar</label>
+        </h1>
+
+        <h3>Demo</h3>
+        <SimpleSlider defaultValue={5} />
+
+        <h3>Usage</h3>
+        <h5>test <label>test</label></h5>
+        <p></p>
+
+
+        <h3>Example</h3>
+        <Markdown>
+          {`
+            <SimpleSlider
+            />
+          `}
+        </Markdown>
+
+
+      </div>
+    );
+  }
+});
+
+module.exports = SimpleSliderDocs;
+
--- a/docs/components/SimpleSliderDocs.js
+++ b/docs/components/SimpleSliderDocs.js
@@ -5,6 +5,18 @@ const { SimpleSlider } = require('mx-react-components');
 const Markdown = require('components/Markdown');
 
 const SimpleSliderDocs = React.createClass({
+  getInitialState () {
+    return {
+      percent: 0
+    };
+  },
+
+  _handleSliderChange (newPercent) {
+    const percent = newPercent;
+
+    this.setState({ percent });
+  },
+
   render () {
     return (
       <div>
@@ -14,17 +26,28 @@ const SimpleSliderDocs = React.createClass({
         </h1>
 
         <h3>Demo</h3>
-        <SimpleSlider defaultValue={5} />
+        <SimpleSlider onPercentChange={this._handleSliderChange} />
 
         <h3>Usage</h3>
-        <h5>test <label>test</label></h5>
-        <p></p>
+        <h5>defaultValue <label>Number</label></h5>
+        <p>This prop moves the slider to the assigned postion. It is passed in as a number between 0 and 1. So 0.75 would put the slider at 75% of the line.</p>
+
+        <h5>disabled <label>bool</label></h5>
+        <p>When true the slider is disabled. The default is false.</p>
+
+        <h5>onPercentChange <label>func</label></h5>
+        <p>When the slider is moved your callback function is called and the current percent is passed as a param. The current percent is a number between 0 and 1.</p>
+
+        <h5>selectedColor <label>string</label></h5>
+        <p>This sets the slider bar color.</p>
 
 
         <h3>Example</h3>
         <Markdown>
           {`
             <SimpleSlider
+              selectedColor='#359BCF'
+              onPercentChange={this._myCallbackFunction}
             />
           `}
         </Markdown>
--- a/docs/components/SimpleSliderDocs.js
+++ b/docs/components/SimpleSliderDocs.js
@@ -11,9 +11,7 @@ const SimpleSliderDocs = React.createClass({
     };
   },
 
-  _handleSliderChange (newPercent) {
-    const percent = newPercent;
-
+  _handleSliderChange (percent) {
     this.setState({ percent });
   },
 
--- a/docs/components/SimpleSliderDocs.js
+++ b/docs/components/SimpleSliderDocs.js
@@ -36,7 +36,7 @@ const SimpleSliderDocs = React.createClass({
         <h5>disabled <label>bool</label></h5>
         <p>When true, the slider is disabled. This defaults to false.</p>
 
-        <h5>onPercentChange <label>func</label></h5>
+        <h5>onPercentChange <label>func</label> <em>Required</em></h5>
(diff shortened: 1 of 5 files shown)


docs/components/SimpleSliderDocs.js
```diff
@@ -0,0 +1,62 @@
...
+    return {
+      percent: 0
+    };
+  },
+
+  _handleSliderChange (newPercent) {
+    const percent = newPercent;
+
+    this.setState({ percent });
+  },
```
jmophoto: You can shorten this.

```
_handleSliderChange (percent) {
    this.setState({ percent });
}
````

File: src/components/SimpleSlider.js [hunk 2]
```diff
@@ -0,0 +1,170 @@
...
+    disabled: React.PropTypes.bool,
+    onPercentChange: React.PropTypes.func.isRequired,
+    percent: React.PropTypes.number.isRequired,
+    selectedColor: React.PropTypes.string,
+    styles: React.PropTypes.object
+  },
+
+  getDefaultProps () {
+    return {
+      selectedColor: '#359BCF'
```
jmophoto: `selectedColor: StyleConstants.Colors.PRIMARY`

src/components/SimpleSlider.js
```diff
@@ -0,0 +1,170 @@
...
+    const component = ReactDOM.findDOMNode(this.rangeSelectorRef);
+    const componentStyles = window.getComputedStyle(component);
+    const width = parseInt(componentStyles.width, 0);
+    const leftPixels = this.props.percent * width;
+
+    this.setState({ width, leftPixels });
+  },
+
+  _handleMouseEvents (e) {
+    const clientX = e.touches ? e.touches[0].clientX : e.clientX;
```
jmophoto: I like this check for mobile. Stealing it for use in the bubble component. 🍪

src/components/SimpleSlider.js
```diff
@@ -0,0 +1,170 @@
...
+  },
+
+  getDefaultProps () {
+    return {
+      selectedColor: '#359BCF'
+    };
+  },
+
+  getInitialState () {
+    const disabled = this.props.disabled || false;
```
jmophoto: It doesn't seem like this needs to be set on the state. You can just use `this.props.disabled` in your checks. If it isn't provided, it would return false.

src/components/SimpleSlider.js
```diff
@@ -0,0 +1,170 @@
...
+    });
+  },
+
+  render () {
+    const styles = this.styles();
+
+    return (
+      <div style={styles.component}>
+        <div
+          onMouseLeave={this.state.disabled ? _noop() : this._handleDragEnd}
```
jmophoto: See comment out setting `disabled` to state. Thinking you could just use the prop. Also, you can just use 'null' instead of importing Lodash's `_noop()`

src/components/SimpleSlider.js
```diff
@@ -0,0 +1,170 @@
...
+
+  styles () {
+    return _merge({}, {
+      component: {
+        position: 'relative',
+        fontFamily: StyleConstants.FontFamily
+      },
+      range: {
+        padding: '25px 0',
+        margin: '0 15px'
```
jmophoto: Spacing constant, here and other places in the styles.

src/components/SimpleSlider.js
```diff
@@ -0,0 +1,170 @@
...
+      },
+      track: {
+        height: '1px',
```
jmophoto: Remove `px`

src/components/SimpleSlider.js
```diff
@@ -0,0 +1,170 @@
...
+      },
+      trackHolder: {
+        padding: '15px 0',
+        cursor: this.state.disabled ? 'default' : 'pointer'
+      },
+      toggle: {
+        width: '20px',
+        height: '20px',
+        borderRadius: '100%',
+        background: '#fff',
```
jmophoto: Color constants.

redperk:
@jmophoto @phantomxc Ready for another look

jmophoto:
Looks good!

:shipit:

jtanner:
🍪 Nice work!

src/components/SimpleSlider.js
```diff
@@ -0,0 +1,175 @@
...
+    this.setState({ leftPixels });
+  },
+
+  _setDefaults () {
+    const component = ReactDOM.findDOMNode(this.rangeSelectorRef);
+    const componentStyles = window.getComputedStyle(component);
+    const width = parseInt(componentStyles.width, 0);
+    const leftPixels = this.props.percent * width;
+
+    this.setState({ width, leftPixels });
```
tumentumurchudur: It is good to list  all the state variables in `getInitialState` method.

redperk: @tumentumurchudur Does that look better?

src/components/SimpleSlider.js
```diff
@@ -0,0 +1,176 @@
...
+    if (currentPercent < 0) {
+      currentPercent = 0;
+    } else if (currentPercent > 1) {
+      currentPercent = 1;
+    }
+
+    this.props.onPercentChange(currentPercent);
+  },
+
+  _handleTrackMouseDown (e) {
```
tumentumurchudur: I don't think this function is needed. You can just use `this._handleMouseEvents` on line 103.

src/components/SimpleSlider.js
```diff
@@ -0,0 +1,176 @@
...
+          ref={(ref) => {
+            this.rangeSelectorRef = ref;
+          }}
+          style={styles.range}
+        >
+          <div
+            onMouseDown={this.props.disabled ? null : this._handleTrackMouseDown}
+            style={styles.trackHolder}
+          >
+            <div style={styles.track}></div>
```
tumentumurchudur: Can this be `<div style={styles.track} />`?

src/components/SimpleSlider.js
```diff
@@ -0,0 +1,176 @@
...
+            <div style={styles.selected}></div>
+          </div>
+          <div
+            onMouseDown={this.props.disabled ? null : this._handleDragStart}
+            onTouchStart={this.props.disabled ? null : this._handleDragStart}
+            style={styles.toggle}
+          >
```
tumentumurchudur: same here.
```javascript
<div
  ...
/>

File: src/components/SimpleSlider.js [hunk 13]
```diff
@@ -0,0 +1,176 @@
...
+    });
+  },
+
+  render () {
+    const styles = this.styles();
+
+    return (
+      <div style={styles.component}>
+        <div
+          onMouseLeave={this.props.disabled ? null : this._handleDragEnd}
```
tumentumurchudur: instead of typing `this.props.disabled` over and over again, can we move 
`const { disabled } = this.props;` to the top and then use `disabled` below?

src/components/SimpleSlider.js
```diff
@@ -0,0 +1,176 @@
...
+        top: '50%',
+        transform: 'translateY(-50%)',
+        WebkitTransform: 'translateY(-50%)',
+        zIndex: 1
+      }
+    }, this.props.styles);
+  }
+});
+
+module.exports = Radium(SimpleSlider);
```
tumentumurchudur: Is `Radium` used in this component?

src/components/SimpleSlider.js
```diff
@@ -0,0 +1,176 @@
...
+      width: 0
+    };
+  },
+
+  componentDidMount () {
+    this._setDefaults();
+  },
+
+  componentWillReceiveProps (newProps) {
+    const leftPixels = (newProps.percent * this.state.width);
```
tumentumurchudur: You may need to minimize the number of renders here
```javascript
if (this.props.percent !== nextProps.percent) {
  ...
}
```

src/components/SimpleSlider.js
```diff
@@ -0,0 +1,176 @@
...
+const _merge = require('lodash/merge');
+
+const StyleConstants = require('../constants/Style');
+
+const SimpleSlider = React.createClass({
+  propTypes: {
+    disabled: React.PropTypes.bool,
+    onPercentChange: React.PropTypes.func.isRequired,
+    percent: React.PropTypes.number.isRequired,
+    selectedColor: React.PropTypes.string,
```
tumentumurchudur: Should this be named `brandColor`?

src/components/SimpleSlider.js
```diff
@@ -0,0 +1,176 @@
...
+  componentWillReceiveProps (newProps) {
+    const leftPixels = (newProps.percent * this.state.width);
+
+    this.setState({ leftPixels });
+  },
+
+  _setDefaults () {
+    const component = ReactDOM.findDOMNode(this.rangeSelectorRef);
+    const componentStyles = window.getComputedStyle(component);
+    const width = parseInt(componentStyles.width, 0);
```
tumentumurchudur: I wonder if you could use `const width = component.clientWidth`.

src/components/SimpleSlider.js
```diff
@@ -0,0 +1,176 @@
...
+    disabled: React.PropTypes.bool,
+    onPercentChange: React.PropTypes.func.isRequired,
+    percent: React.PropTypes.number.isRequired,
+    selectedColor: React.PropTypes.string,
+    styles: React.PropTypes.object
+  },
+
+  getDefaultProps () {
+    return {
+      selectedColor: StyleConstants.Colors.PRIMARY
```
tumentumurchudur: since `disabled` is not required, I would add that in here
`disabled: false`

src/components/SimpleSlider.js
redperk: Sadly yes, I am forced to use it for the different cursors. Each browser has it's on implementation and Radium handles it all.

src/components/SimpleSlider.js
```diff
@@ -0,0 +1,174 @@
...
+    };
+  },
+
+  componentDidMount () {
+    this._setDefaults();
+  },
+
+  componentWillReceiveProps (newProps) {
+    if (this.props.percent !== newProps.percent) {
+      const leftPixels = (newProps.percent * this.state.width);
```
tumentumurchudur: No need for curly braces. `const leftPixels = newProps.percent * this.state.width;`

src/components/SimpleSlider.js
```diff
@@ -0,0 +1,174 @@
...
+    let cursorStyle = 'grab';
+
+    if (this.props.disabled) {
+      cursorStyle = 'not-allowed';
+    }
+    if (this.state.dragging) {
+      cursorStyle = 'grabbing';
+    }
+
+    return _merge({}, {
```
tumentumurchudur: Can we use `Object.assign({}, { ` instead?

redperk: I tired `Object.assign` the problem is that it doesn't do a deep merge, if you update 1 thing it just wipes out all the rest.

src/components/SimpleSlider.js
```diff
@@ -0,0 +1,174 @@
...
+      component: {
+        position: 'relative',
+        fontFamily: StyleConstants.FontFamily
```
tumentumurchudur: This looks wrong to me. Did you mean `StyleConstants.Fonts.REGULAR`?

redperk: That was left over from the range selector, but it's not needed here.

src/components/SimpleSlider.js
```diff
@@ -0,0 +1,173 @@
...
+    const width = component.clientWidth;
+    const leftPixels = this.props.percent * width;
+
+    this.setState({ width, leftPixels });
+  },
+
+  _handleMouseEvents (e) {
+    const clientX = e.touches ? e.touches[0].clientX : e.clientX;
+    const leftSpace = ReactDOM.findDOMNode(this.rangeSelectorRef).getBoundingClientRect().left;
+    let currentPercent = ((clientX - leftSpace) / this.state.width);
```
tumentumurchudur: No need for outer curly braces
`let currentPercent = (clientX - leftSpace) / this.state.width;`

src/components/SimpleSlider.js
```diff
@@ -0,0 +1,173 @@
...
+
+  componentWillReceiveProps (newProps) {
+    if (this.props.percent !== newProps.percent) {
+      const leftPixels = newProps.percent * this.state.width;
+
+      this.setState({ leftPixels });
+    }
+  },
+
+  _setDefaults () {
```
tumentumurchudur: What do you think about removing this function and then moving its content directly into `componentDidMount` method? I think it is easier to see what is going on.

tumentumurchudur:
Looks good. :shipit:

PR #467, opened 2016-11-07
hide datepicker icons on xsmall devices, simplify flex rules

paniclater:
- Only display calendar and caret icons on devices large enough to support them
- Simplifies flex rules a little bit
- enforces datepicker height to match style guide

BEFORE:
--
screen shot 2016-11-07 at 3 58 24 pm

AFTER:
--
iPhone 6+
iphone6plus
iPhone 6
iphone6
iPhone 5
iphone5
Nexus 6p
nexus6p
nexus 5x
nexus5x
iPad:
screen shot 2016-11-07 at 3 17 06 pm

### DIFF
--- a/src/components/DateRangePicker.js
+++ b/src/components/DateRangePicker.js
@@ -298,11 +298,12 @@ const DateRangePicker = React.createClass({
     return (
       <div style={styles.component}>
         <div onClick={this._toggleCalendar} style={styles.selectedDateWrapper}>
-          <Icon
-            size={20}
-            style={styles.selectedDateIcon}
-            type='calendar'
-          />
+          {StyleConstants.getWindowSize() !== 'xsmall' &&
+            <Icon
+              size={20}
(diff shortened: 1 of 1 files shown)


iheartrachie:
:shipit:

src/components/DateRangePicker.js
```diff
@@ -416,18 +419,16 @@ const DateRangePicker = React.createClass({
       // Selected Date styles
       selectedDateWrapper: {
         alignItems: 'center',
-        cursor: 'pointer',
         display: 'flex',
-        justifyContent: 'space-between',
-        position: this.props.isRelative ? 'relative' : 'static'
+        height: 20,
+        justifyContent: 'space-between'
```
paniclater: position rule & cursor was not necessary because it is already on the component level rules.

Without height the component was using vertical padding to resize to the weird sizes.

PR #468, opened 2016-11-07
Fix refs - use call back pattern over strings

cerinman:
Since moving Docs into the master branch, we've been having issues with using strings as refs within the Docs.  This PR fixes that issues by updating our usage to the call back pattern that Facebook favors over strings.  The docs should now continue to work after visiting the docs for a component that uses refs.


jmophoto:
:shipit:

PR #473, opened 2016-11-16
Simple Select: add scrimClickOnSelect to props.

jmophoto:
This adds a `scrimClickOnSelect` that will trigger a scrim click on select if set to true. This allows for a pseudo auto-close functionality. Set to false be default.


iheartrachie:
:shipit:

PR #475, opened 2016-11-16
Let the browser know that Button should act like a button

jtanner:
This change lets the browser and other accessibility tools (e.g. Vimium) know that the `Button` component is clickable.

Related to #30


lukeschunk:
🍪

paniclater:
:shipit:

PR #476, opened 2016-11-28
Handles IE drag cursor issues

redperk:
IE does not support `drag` or `dragging` cursor icons, so I added a check to see if the browser is IE.

Gif:
ezgif com-video-to-gif 9

### DIFF
--- a/src/components/SimpleSlider.js
+++ b/src/components/SimpleSlider.js
@@ -2,6 +2,7 @@ const React = require('react');
 const ReactDOM = require('react-dom');
 const Radium = require('radium');
 const _merge = require('lodash/merge');
+const browser = require('bowser');
 
 const StyleConstants = require('../constants/Style');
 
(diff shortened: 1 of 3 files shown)


src/components/SimpleSlider.js
```diff
@@ -114,11 +115,18 @@ const SimpleSlider = React.createClass({
   styles () {
     let cursorStyle = 'grab';
```
jmophoto: This is feeling like it should be a method that returns a cursorStyle. 

```
getCursorStyle () {
  if (this.props.disabled) {
   return 'not-allowed';
  } else  if (browser.msie) {
    return 'pointer';
  } else {
    return this.state.dragging ? 'grabbing' : 'grab'
 }
}
```

jmophoto:
:shipit: before chrome breaks stuff again!

jmophoto:
:shipit:

jmophoto:
:shipit:

PR #477, opened 2016-12-01
D3 CircleGroup: Add ability to have an overlay for circles

cerinman:
I have a need internally to be able to render a 0 opacity circle overlay with a larger radius over the visible circles to make clicking simpler in both desktop and mobile.  This PR does this by..

- Adding a new `useCircleOverlay` prop that defaults to false.
- Adding a new `circleOverlayRadius` prop that defaults to 6.
- If `useCircleOverlay` is true then we add a second circle with an opacity of 0 that has a radius equal to the `circleOverlayRadius` prop.
- Adds the onClick handler to the overlay and removes it from visible circle.
- Also fixed a console error around incorrect style property for `strokeWidth`

### Screen shot of the invisible overlay 

screen shot 2016-12-01 at 9 40 34 am

### DIFF
--- a/src/components/d3/CirclesGroup.js
+++ b/src/components/d3/CirclesGroup.js
@@ -48,7 +48,7 @@ const CirclesGroup = React.createClass({
   },
 
   render () {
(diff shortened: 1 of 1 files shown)


jtanner:
:shipit:

PR #478, opened 2016-12-08
Chart utils getDataMinMaxValues function fix

cerinman:
The axis domain for charts that use the AxisGroup d3 component was off by a power of 1.  This was due to the fact that if the min or max value was a negative number then the negative character was included in the length check done after converting the number to a string.

giphy 26

I also have added the ChartUtils to the module.exports so that I can use them internally rather than duplicating the min max value code.

### DIFF
--- a/src/utils/Chart.js
+++ b/src/utils/Chart.js
@@ -10,15 +10,8 @@ const Chart = {
       return d[axis];
     });
 
-    //remove negative character if it exists
-    let maxString = Math.ceil(max).toString();
-    let minString = Math.floor(min).toString();
-
-    maxString = maxString.replace('-', '');
-    minString = minString.replace('-', '');
-
-    const maxDigits = maxString.length - 1;
-    const minDigits = minString.length - 1;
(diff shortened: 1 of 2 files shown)


src/utils/Chart.js
```diff
@@ -10,8 +10,16 @@ const Chart = {
...
-    const minDigits = Math.floor(min).toString().length - 1;
+    //remove negative character if it exists
+    let maxString = Math.ceil(max).toString();
+    let minString = Math.floor(min).toString();
+
+    maxString = maxString.replace('-', '');
+    minString = minString.replace('-', '');
+
+    const maxDigits = maxString.length - 1;
+    const minDigits = minString.length - 1;
```
jmophoto: Thinking these could all be chained.
`const maxDigits = Math.ceil(max).toString().replace('-', '').length - 1;`

Also, instead of the replace, you could do `Math.abs(Math.ceil(max)).to....` I tihink.

jmophoto:
:shipit:

PR #479, opened 2016-12-14
Chart utils tweaks

cerinman:
We were running into issues where the internal widgets axis ranges/ticks were not matching the mobile applications ranges/ticks.  To solve the problem, I've ported the mobile c++ code over and revamped the Chart Utils file.  Eventually I'll port more of the mobile teams code but for now this allows our line charts to be in parity with theirs.

This only has an effect on the TimeBasedLineChart in the open source but will also effect the internal NetWorth chart.  I'll have screen shots in the internal PR showing how the graphs now look but the gist is that the y range and tick values are dialed in closer so that the line changes between points more dramatically.

@jmophoto Once this goes live you can start using it in Insight/Target as well.

### DIFF
--- a/src/components/TimeBasedLineChart.js
+++ b/src/components/TimeBasedLineChart.js
@@ -272,9 +272,17 @@ const TimeBasedLineChart = React.createClass({
   },
 
   _yRangeContainsZero () {
-    const minMaxValues = ChartUtils.getDataMinMaxValues(this.props.data, 'y');
+    const max = d3.max(this.props.data, d => {
+      return d.y;
+    });
 
-    return minMaxValues.min <= 0 && minMaxValues.max >= 0;
+    const min = d3.min(this.props.data, d => {
+      return d.y;
+    });
+
+    const tickSpec = ChartUtils.getAxisTickSpecification(min, max);
+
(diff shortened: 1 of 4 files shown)


src/components/TimeBasedLineChart.js
```diff
@@ -272,9 +272,17 @@ const TimeBasedLineChart = React.createClass({
...
 
   _yRangeContainsZero () {
-    const minMaxValues = ChartUtils.getDataMinMaxValues(this.props.data, 'y');
+    const max = d3.max(this.props.data, d => {
+      return d.y;
+    });
+
+    const min = d3.min(this.props.data, d => {
+      return d.y;
+    });
```
jtanner: What do you think about using the new shorthand style implicit returns to make these (and the ones in the other files) a tiny bit easier to read?
```javascript
const max = d3.max(this.props.data, d => d.y);
const min = d3.min(this.props.data, d => d.y);
```

redperk:
:shipit:

src/components/TimeBasedLineChart.js
cerinman: yep, I'm good with that.  I'll make the change.

Issue #460, opened 2016-10-20
New Component / Needle


derek-boman:
The Needle is a new component used to indicate a UI element the user should interact with next. An example use case would be: a first-time user coming to a Goals app but hasn't added any goals yet.

**A few requirements:**
1. It is not clickable itself but should point to something that is.  
2. The Needle subtly bounces up and down (or side-to-side, depending on its orientation). It moves 10px from its starting position with a easing-curve of `cubic-bezier(0.6, -0.5, 0.2, 1.5)`
3. Color default should be `BANANA`.
4. Developer should be able to set the orientation.

**SVG File:** 
https://moneydesktop.box.com/s/vmzj8daauvwqg3a8xx3azz6l1vauel05

**Example:**
![needle_example](https://cloud.githubusercontent.com/assets/17180771/19569126/cd3c4c88-96b1-11e6-8e71-e5f679485a8e.gif)

jtanner:
Will this be separate from or part of an on-boarding tour?

derek-boman:
The current application is in an onboarding tour but I wouldn't want to limit it to just that.

Maxnelson997:
First time contributing to this repo. I'll start with this.

derek-boman:
Thanks @Maxnelson997 for your help. Happy to have you here. I don't mean to rush you but we have a fairly urgent need for this, when do you think you might have something?

Maxnelson997:
It's pretty much done, I did it last night. Though this is my very first react component and all my finals are on Monday and Tuesday. So I will most likely have it done Wednesday.

derek-boman:
Thanks for the update. Can't wait to see it. We appreciate your contribution.

paniclater:
@Maxnelson997 So excited to see you contributing! Please hit me up here or on the UtahJS slack if you have questions!

Maxnelson997:
Click the picture, it's a video of the component. This is the desired functionality correct?
[![MX Needle](https://cloud.githubusercontent.com/assets/18063554/21040040/76893184-bd9f-11e6-9fa3-e78a78e8b775.png)](https://youtu.be/IDTE3_btlwI "MX Needle")

cerinman:
This has been finished.  Closing.

PR #482, opened 2016-12-15
Needle Component 

Maxnelson997:
Needle Component. #460

### DIFF
--- a/src/components/Needle.js
+++ b/src/components/Needle.js
@@ -1,0 +1,106 @@
+
+const React = require('react');
+
+const ReactDOM = require('react-dom');
+
+const _merge = require('lodash/merge');
+
+const Needle = React.createClass({
+  propTypes: {
+    direction: React.PropTypes.string
+  },
+
+  getDefaultProps () {
+    return {
+      direction: 'up'
+    };
+  },
+
+  componentDidMount () {
+    const el = ReactDOM.findDOMNode(this);
+
+    let rotation = 25;
+
+    const direction = this.props.direction;
+
+      //sets the rotation
+    switch (direction) {
+      case 'up':
+        rotation = 0;
+        break;
+      case 'right':
+        rotation = 90;
+        break;
+      case 'down':
+        rotation = 180;
+        break;
+      case 'left':
+        rotation = 270;
+        break;
+    }
+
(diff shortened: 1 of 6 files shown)


cerinman:
Hi @Maxnelson997!  Thanks for taking the time to submit code to our open source.  I'm going to open a review and leave some feed back on your code but I wanted to say thank you first.  It's always nice to have help on the open source!  👍 🚀

src/components/Needle.js
```diff
@@ -0,0 +1,106 @@
...
+      }
+
+      if (up) {
+        distance += 1;
+      }
+      if (!up) {
+        distance -= 1;
+      }
+    }, 30);
+  },
```
cerinman: We have been using the animation library `Velocity` for animating things. The Drawer component might be a good reference to see how we use it.  Lets see if it makes sense here.

src/components/Needle.js
```diff
@@ -0,0 +1,106 @@
...
+
+  render () {
+    const styles = this.styles();
+
+    return (
+      <div className='mx-needle' style={Object.assign({}, styles.component)}></div>
```
cerinman: Here is where you would add the callback ref.  If we switch to using the `Icon` component it would actually be on that element in stead.  Here is an example of it.

```javascript
<div ref={(ref) => this.icon = ref} >
```

You can then access this element in your code else where by simply referencing this.icon.

src/components/Needle.js
```diff
@@ -0,0 +1,106 @@
...
+  styles () {
+    return _merge({}, {
+      component: {
+        position: 'relative',
+        display: 'inline-block',
+        backgroundImage: 'url("needle.svg")',
+        height: '50px',
+        src: 'needle.svg',
+        width: '50px'
+      }
```
cerinman: So here you could add a style for the svg icon and call your new `_getIconRotation` function to get the rotation string.

Maxnelson997:
@cerinman Awesome thank you for pointing these things out, I'll check it out as soon as I can and i'll reach out if I have any questions to prevent garbage code from getting in the code base haha.

paniclater:
Thank you for your contribution, but closing due to stale request.

PR #483, opened 2016-12-15
Button Component: New primaryInverse type

cerinman:
This PR adds a new button type to the button component.  `primaryInverse`.  It's purpose is to allow for situations where the background color of the parent element is already a primary color.  The primaryOutline partially solved this type of problem but not to the extent that is needed in some cases. The main differences being, the outline, and the hover/active color.  primaryInverse remains in the white scale while the primaryOutline hover is the primary color.

screen shot 2016-12-15 at 10 37 51 am

### DIFF
--- a/docs/components/ButtonDocs.js
+++ b/docs/components/ButtonDocs.js
@@ -87,7 +87,7 @@ const ButtonDocs = React.createClass({
 
         <h5>type <label>String</label></h5>
         <p>Default: 'primary'</p>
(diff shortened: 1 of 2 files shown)


docs/components/ButtonDocs.js
```diff
@@ -30,7 +30,7 @@ const ButtonDocs = React.createClass({
       <div>
         <h1>
           Button
-          <label>A standard button with 6 available styles.</label>
+          <label>A standard button with 7 available styles.</label>
```
jtanner: Good catch 👍

lukeschunk:
:shipit:

Issue #487, opened 2017-01-05
New Component / Message Box


derek-boman:
This component will allow for messaging to the user for information, errors, or warnings.

![info medium](https://cloud.githubusercontent.com/assets/17180771/21698644/21ad0e62-d355-11e6-9c0f-7a29d29292d0.png)
![warning mobile](https://cloud.githubusercontent.com/assets/17180771/21698626/09199348-d355-11e6-8eb4-f44bd882c417.png)

Here's the spec: https://app.frontify.com/d/KHQZnlj1XMIO/web-patterns#/general/messaging

lukeschunk:
we have this component now - closing

PR #490, opened 2017-01-10
Luke/message box component Issue 487

lukeschunk:
Create the message Box component seen [here](https://github.com/mxenabled/mx-react-components/issues/487)

The message box has the option to be expandable - in order to handle this, i allowed the component to take in children - this makes it more reusable - we could add more buttons & contents if need be. 

messagebox


docs/components/MessageBoxDocs.js
```diff
@@ -0,0 +1,83 @@
...
+          <div style={styles.messageBoxContents}>
+            <div style={styles.messageBoxText}>This is some message box text.</div>
+            <Button style={styles.button}>Its Button Time</Button>
+          </div>
+        </MessageBox>
+        <h3>Usage</h3>
+        {
+          <div>
+            <h5>children<label>Node</label></h5>
+            <p>If the MessageBox is expandable, children will be the contens of the expanded box.</p>
```
jtanner: *contents

src/components/MessageBox.js
```diff
@@ -0,0 +1,96 @@
...
+    children: React.PropTypes.node,
+    color: React.PropTypes.string,
+    expandable: React.PropTypes.bool,
+    icon: React.PropTypes.string,
+    title: React.PropTypes.string
+  },
+
+  getInitialState () {
+    return {
+      isOpen: false
```
jtanner: I think this should have an `isOpen` prop as well.

src/components/MessageBox.js
```diff
@@ -0,0 +1,96 @@
...
+  render () {
+    const styles = this.styles();
+
+    return (
+      <div className='mx-message-box' style={styles.component}>
+        <div onClick={this._toggleMessageBox} style={styles.header}>
+          <div style={styles.leftHeader}>
+            <Icon
+              elementProps={{
+                onClick: this._toggleMessageBox
```
jtanner: Is the `onClick` handler necessary here since the parent div on line 32 has the same handler?

src/components/MessageBox.js
```diff
@@ -0,0 +1,96 @@
...
+              type={this.props.icon}
+            />
+
+            {this.props.title}
+          </div>
+
+          {this.props.expandable &&
+            <Icon
+              elementProps={{
+                onClick: this._toggleMessageBox
```
jtanner: Same here.

src/components/MessageBox.js
```diff
@@ -0,0 +1,96 @@
+const React = require('react');
+
+const StyleConstants = require('../constants/Style');
+const Icon = require('../components/Icon');
+
+const MessageBox = React.createClass({
+  propTypes: {
+    children: React.PropTypes.node,
+    color: React.PropTypes.string,
+    expandable: React.PropTypes.bool,
```
jtanner: After looking at the design docs on this I think we can get rid of this `expandable` prop and assume that it's expandable if there are children. And also add the `isOpen` prop that I mentioned before.

src/components/MessageBox.js
lukeschunk: good call, i wasn't sure if it should all be clickable at first but i think it's better that way

src/components/MessageBox.js
lukeschunk: 👍 to assuming it's expandable if there are children. 

I'm hesitant to add an isOpen prop, unless we just use it as a value to set the initial state (is that ok practice?) I think using callbacks to setProp values in the parent gets messier - this component is a good use case to manage its own state

jtanner: Good point about having to use callbacks, so never mind about the `isOpen` prop idea and because there isn't a use case for it yet.

lukeschunk:
@jtanner ready for another look - also check out [this line](https://github.com/mxenabled/mx-react-components/pull/490/files#diff-8594106243f200deddd2576ac3f83583R33) lmk what you think

lukeschunk:
also added a second example where i don't pass children: 

screen shot 2017-01-11 at 1 29 28 pm

src/components/MessageBox.js
```diff
@@ -0,0 +1,91 @@
...
+    children: React.PropTypes.node,
+    color: React.PropTypes.string,
+    icon: React.PropTypes.string,
+    styles: React.PropTypes.object,
+    title: React.PropTypes.string
+  },
+
+  getInitialState () {
+    return {
+      isOpen: false
```
jtanner: @derek-boman, @lukeschunk: should the `MessageBox` default to the open state or closed? I think it should be open by default so I'm curious to hear your opinion on it.

lukeschunk: hmm good question. 

On that note, since there is a typical 'message box content' [spec](https://app.frontify.com/d/KHQZnlj1XMIO/web-patterns#/general/messaging) i'm planning on creating a component for that that we can use for default children, but then if we want to pass different children we have that option. what do you think of that idea?

jtanner: Sounds good. What would the component do, add flexbox and font styling like in the demo file?

derek-boman: I'd say open by default.

lukeschunk:
@jtanner made the box open by default and a few other little fixes in my latest commit

PR #491, opened 2017-01-12
PaginationButtons Component

iheartrachie:
This adds the PaginationButtons component and its docs. The component uses a stylized version of the ButtonGroup component and is to be used with paginated content. It supports all button types and custom styling although it is not recommended at this time.

### Example
zbupsasd5l

### Props
`currentPage` **Number** _Required_
`totalPages` **Number** _Required_
`pageRange` **Number** The number of page buttons (not prev/next buttons) to display
`primaryColor` **String** Defaults to `Styles.Colors.PRIMARY`
`onClick` **Function** Returns the number clicked. If prev/next button is clicked, the number returned is -1/+1 the current page number.


phantomxc:
My first thought just from looking at the GIF, I'd like to see the total number of pages and be able to jump to the end.

Random example:
image

iheartrachie:
How does this look, @phantomxc? If it's what you're looking for, I'll push up the code.
BTW, the dots buttons aren't doing anything. Should they?

www

phantomxc:
I think it looks great but don't let me be the only one that dictates how this functions. I also don't think the `...` buttons should do anything. You could maybe gray them out?

src/components/PaginationButtons.js
```diff
@@ -0,0 +1,108 @@
...
+
+const { buttonTypes } = require('../constants/App');
+
+const PaginationButtons = React.createClass({
+  propTypes: {
+    currentPage: React.PropTypes.number.isRequired,
+    onClick: React.PropTypes.func,
+    pageRange: React.PropTypes.number,
+    primaryColor: React.PropTypes.string,
+    styles: React.PropTypes.object,
```
jtanner: Is there a reason this is `styles` instead of `style`? I think this should be renamed to `style` to match the [React naming convention](https://facebook.github.io/react/docs/dom-elements.html#style).

iheartrachie:
I added the gray fill and default cursor but didn't touch the hover color. I can't do much more than that without messing with the styles of the different button types.

cqgrmgr01b

src/components/PaginationButtons.js
```diff
@@ -0,0 +1,186 @@
...
+      } else if (middleSet) {
+        endingPage = startingPage + pageRange - 5;
+      }
+    }
+
+    return endingPage;
+  },
+
+  _addFirstPageButtons () {
+    const style = this.styles();
```
jtanner: Can you override the styles here to remove the hover?

iheartrachie: For some reason the software I'm using to make the gif only shows the default cursor. The page and arrow buttons have a pointer unless they are grayed out or 'disabled'.

src/components/PaginationButtons.js
```diff
@@ -0,0 +1,186 @@
...
+
+    return [{
+      onClick: this._handleButtonClick.bind(null, 1),
+      style: style.component,
+      text: '1'
+    }, {
+      icon: 'kabob_horizontal',
+      style: Object.assign({}, style.component, style.disabled)
```
jtanner: Regarding my previous comment about the hover state of the `...` buttons, I think you can add `type: 'disabled'` after this line to both get the disabled styling and remove the hover state styling. Then you could remove the `disabled` styles and this line would simply be `style: style.component,`.

iheartrachie: Changing the button type would mess with the CSS, including the borders. We want the button group to have a single type to avoid CSS weirdness but maybe that's something we could add in the future?

src/components/ButtonGroup.js
```diff
@@ -14,13 +16,7 @@ const ButtonGroup = React.createClass({
...
     }).isRequired),
     primaryColor: React.PropTypes.string,
-    type: React.PropTypes.oneOf([
-      'base',
-      'disabled',
-      'neutral',
-      'primary',
-      'primaryOutline'
-    ])
+    type: React.PropTypes.oneOf(buttonTypes)
```
jtanner: Good catch on the inconsistencies between here and `Button` 🍺

iheartrachie:
Worked with @jtanner offline and we found a good way to add support for disabled buttons inside of groups. Once [this PR](https://github.com/mxenabled/mx-react-components/pull/494) is merged I'll update this component to use the new feature.

iheartrachie:
Updated the component to include support for disabled buttons inside the group. It's _much_ cleaner now!

page

jtanner:
:shipit:

PR #493, opened 2017-01-16
TimeBasedLineChart Polish

cerinman:
As a part of making an internal widget mobile responsive, I noticed several issues with this chart for alignment.  This PR corrects the issue by doing the following.

- Makes the Demo for the chart in the docs have a dynamic size for testing responsiveness
- Fixes zero line and zero line label position
- Fixes right margin of chart so final datum point doesn't have its label cut off
- Updates `TimeAxisGroup` component to use an array of tick values instead of letting D3 determine them.  We now also filter and map over the data set to generate the required array in the usage of the component within the `TimeBasedLineChart`
- Fixes a couple of minor styles

### Before

screen shot 2017-01-16 at 1 38 58 pm

### After

screen shot 2017-01-16 at 1 41 10 pm

### DIFF
--- a/docs/components/TimeBasedLineChartDocs.js
+++ b/docs/components/TimeBasedLineChartDocs.js
@@ -59,6 +59,7 @@ const TimeBasedLineChartDocs = React.createClass({
           data={lineChartData}
           height={this.state.chartHeight}
           rangeType={'month'}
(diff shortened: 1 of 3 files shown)


iheartrachie:
:shipit:

PR #496, opened 2017-01-19
Adds NotifyOnScrollThreshold Component

cerinman:
#### Props

`onThresholdMet` - function - default(`no operation`): an optional prop callback
`threshold` - number - default(`0.9`): Represents percentage between 0% and 100%

This component take a function as a child and provides values as arguments to the function when it's called.  The values passed are:

- `thresholdMet` - true when the user has scrolled passed the supplied threshold.
- `scrollTop` - the current scroll position.
- `scrollHeight` - the height available to scroll.

See example usage in the docs file for more details.

### Component in action

scroll-notify

### DIFF
--- a/src/components/NotifyOnScrollThreshold.js
+++ b/src/components/NotifyOnScrollThreshold.js
@@ -37,6 +37,11 @@ const NotifyOnScrollThreshold = React.createClass({
   },
 
   render () {
+    // TODO:
+    // 1. Position from bottom passed to children function
+    // 2. Clean up _handleScroll
+    // 3. What happens if the parent isn't scrollable due to lack of content?
+
     return (
       <div ref={(ref) => this.container = ref}>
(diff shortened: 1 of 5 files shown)


src/components/NotifyOnScrollThreshold.js
```diff
@@ -0,0 +1,66 @@
...
+      thresholdMet: false
+    };
+  },
+
+  componentDidMount () {
+    this._handleScroll = _throttle(this._handleScroll, 200);
+
+    this.container.parentElement.addEventListener('scroll', this._handleScroll);
+
+    if (this.container.parentElement.scrollHeight <= this.container.parentElement.clientHeight) {
```
iheartrachie: 🍪

lukeschunk:
@cerinman walked me through this and i dig it 👍

phantomxc:
:shipit:

Issue #495, opened 2017-01-19
Drawer / Revisions to drawer header


derek-boman:
The drawer header currently has the option for sibling navigation in the top right corner. This seemed like a good idea at the time but has not proven very useful. Instead, this space is better when reserved for additional options menus. This is usually a "Kabob" menu using "Simple Select" component but should be left open for the developer how to fill this space according to their needs.

cc / @vintagepenguin 

![image](https://cloud.githubusercontent.com/assets/17180771/22119222/f5b8337a-de37-11e6-99e4-c2322043447b.png)

PR #497, opened 2017-01-19
Drawer Component: Pass Component To the Drawer Header

vintagepenguin:
Issue: https://github.com/mxenabled/mx-react-components/issues/495

## Shout out to @paniclater for a dope pairing session with this. 

# `GIF:` after most recent commit `(1/20/17)`
[https://gyazo.com/4b37d42babd16b15efefa5d190059fc4](https://gyazo.com/4b37d42babd16b15efefa5d190059fc4)

### DIFF
--- a/docs/components/DrawerDocs.js
+++ b/docs/components/DrawerDocs.js
@@ -1,6 +1,6 @@
 const React = require('react');
 
-const { Button, Drawer, SimpleSelect } = require('mx-react-components');
+const { Button, Drawer, HeaderMenu } = require('mx-react-components');
 
 const Markdown = require('components/Markdown');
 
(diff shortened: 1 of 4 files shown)


src/components/Drawer.js
```diff
@@ -24,16 +24,12 @@ const Drawer = React.createClass({
     ]),
     duration: React.PropTypes.number,
     easing: React.PropTypes.array,
+    headerMenu: React.PropTypes.func,
```
phantomxc: This can be `React.PropTypes.element`

paniclater:
This is great! However, you need to add a check so if the prop isn't passed in the old behavior is preserved so it can be a non-breaking change

src/components/Drawer.js
vintagepenguin: ohhhh niccce pro-tip ⚡️

docs/components/DrawerDocs.js
```diff
@@ -140,17 +150,72 @@ const DrawerDocs = React.createClass({
...
+      onClose={this._handleDrawerClose}
       title='Demo Drawer'
     >
-      //Content here
+      // Content Here
+
     </Drawer>
+
+    // Component To Pass As A Prop
+    function HeaderMenu ({ handleButtonClick, handleScrimClick, showSimpleSelectMenu = false }) {
```
paniclater: is this defined 2 times?

paniclater: I also think we should add the header menu as a separate component in `src/components/HeaderMenu.js`

src/components/Drawer.js
```diff
@@ -161,7 +162,9 @@ const Drawer = React.createClass({
               <span style={styles.title}>
                 {this.props.title}
               </span>
-              {this._renderNav()}
+              <div style={styles.headerMenu}>
+                {this.props.headerMenu ? this.props.headerMenu : this._renderNav}
```
paniclater: 👍

phantomxc: this._renderNav() ?

docs/components/DrawerDocs.js
vintagepenguin: it's defined once in the code and once in the docs

src/components/Drawer.js
vintagepenguin: 💯  nicce catch

paniclater:
:shipit:

paniclater:
:shipit:

PR #498, opened 2017-01-23
adding check for navConfig in order for drawer to render

paniclater:
The last drawer release removed a check for `navConfig` before rendering the nav. This was breaking previous implementations that had neither a headerMenu prop or a navConfig. 

This adds the check back in.


lukeschunk:
:shipit:

PR #500, opened 2017-01-26
semibold for MessageBox Header

lukeschunk:
@derek-boman noticed this title/header text should be semibold. 

screen shot 2017-01-25 at 5 32 26 pm


jmophoto:
:shipit:

PR #502, opened 2017-01-26
Wengert/barchart

phantomxc:
This is a very large refactor to the BarChart component.

- Fixes the re-rendering issue.
- Combines positive and negative values into a single SVG
- Removes the spans for labels
- Now accepts xAxis and yAxis elements allowing you to do axis math from d3.
- More style control.
- A few more things...

### DIFF
--- a/docs/components/BarChartDocs.js
+++ b/docs/components/BarChartDocs.js
@@ -154,10 +154,10 @@ const BarChartDocs = React.createClass({
         <p>Default: 500</p>
 
         <h5>xAxis<label>Element</label></h5>
-        <p>A element representing the xAxis. You are incharge of transforming to the correct location.</p>
+        <p>A element representing the xAxis. You are in charge of transforming to the correct location.</p>
 
         <h5>yAxis<label>Element</label></h5>
-        <p>A element representing the yAxis. You are incharge of transforming to the correct location.</p>
+        <p>A element representing the yAxis. You are in charge of transforming to the correct location.</p>
 
         <h3>Example</h3>
         <Markdown lang='js'>
--- a/docs/components/BarChartDocs.js
+++ b/docs/components/BarChartDocs.js
@@ -1,72 +1,31 @@
 const React = require('react');
+const d3 = require('d3');
+const moment = require('moment');
 
 const Markdown = require('components/Markdown');
 
-const { BarChart } = require('mx-react-components');
+const { BarChart, BarTimeXAxis, Styles } = require('mx-react-components');
 
-const chartData = [
-  {
-    color: '#E3E6E7',
-    label: 'Jan',
-    value: 595.45
-  },
-  {
-    color: '#E3E6E7',
-    label: 'Feb',
-    value: 690.65
-  },
-  {
-    color: '#E3E6E7',
-    label: 'Mar',
-    value: 897.40
-  },
-  {
-    color: '#E3E6E7',
-    label: 'Apr',
-    value: 1209.95
-  },
-  {
-    color: '#E3E6E7',
-    label: 'May',
-    value: 1191.00
-  },
-  {
-    color: '#E3E6E7',
-    label: 'Jun',
-    value: 1391.27
-  },
-  {
-    color: '#E3E6E7',
-    label: 'Jul',
-    value: 991.23
-  },
-  {
-    color: '#E3E6E7',
-    label: 'Aug',
-    value: 1101.45
-  },
-  {
-    color: '#E3E6E7',
-    label: 'Sep',
-    value: 1301.45
-  },
-  {
-    color: '#E3E6E7',
-    label: 'Oct',
-    value: 1405.45
-  },
-  {
-    color: '#E3E6E7',
-    label: 'Nov',
-    value: 1600.45
-  },
-  {
-    color: '#E3E6E7',
-    label: 'Dec',
-    value: 2100.45
-  }
+const margins = {
+  top: 50,
+  right: 20,
+  bottom: 40,
+  left: 20
+};
+const width = 500 - margins.right - margins.left;
+const height = 300 - margins.top - margins.bottom;
 
-];
+const chartData = [];
+
+for (let i = 0; i < 12; i++) {
+  const date = moment().add(i, 'M');
+
+  chartData.push({
+    label: date.unix(),
+    value: Math.round(Math.random() * (10000 - -10000) + -10000)
+  });
+}
+
 
 const BarChartDocs = React.createClass({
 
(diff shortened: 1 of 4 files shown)


docs/components/BarChartDocs.js
```diff
@@ -127,97 +111,93 @@ const BarChartDocs = React.createClass({
...
-        <p>A style object that can be passed to style the tooltip.</p>
+        <h5>tooltipFormat<label>Function</label></h5>
+        <p>A function that is called to format the value being displayed in the tooltip.</p>
 
         <h5>width<label>Number</label></h5>
         <p>Width of the SVG; bars will determine their individual width accordingly.</p>
         <p>Default: 500</p>
 
+        <h5>xAxis<label>Element</label></h5>
+        <p>A element representing the xAxis. You are incharge of transforming to the correct location.</p>
```
jmophoto: Space between in and charge.

src/components/BarChart.js
```diff
@@ -1,299 +1,351 @@
...
   getInitialState () {
     return {
-      hovering: false
+      hoveringObj: {},
+      clickedData: {}
     };
   },
 
   shouldComponentUpdate (nextProps, nextState) {
     return !_isEqual(nextProps, this.props) || !_isEqual(nextState, this.state);
```
jmophoto: Do you need to omit the functions like you did in the bar?

jmophoto:
Besides one question and a wording change, this looks good to go. 🍪

src/components/BarChart.js
phantomxc: Since that is more the public API, I don't think we can guarantee that those functions aren't changing? That's my thinking anyways...thoughts?

docs/components/BarChartDocs.js
```diff
@@ -89,9 +48,27 @@ const BarChartDocs = React.createClass({
     });
   },
 
+
   render () {
     const styles = this.styles();
 
+    const ticks = chartData.map(d => {
```
paniclater: Are we cool with one letter function parameters? I am, especially in one liners like this.

src/components/BarChart.js
```diff
@@ -1,299 +1,351 @@
...
-    primaryColor: React.PropTypes.string,
-    value: React.PropTypes.number.isRequired,
+    radius: React.PropTypes.number,
+    value: React.PropTypes.number,
     width: React.PropTypes.number,
-    x: React.PropTypes.number.isRequired,
-    y: React.PropTypes.number.isRequired
-  },
-
-  mixins: [SetIntervalMixin],
```
paniclater: <3 thank you for removing mixins <3

docs/components/BarChartDocs.js
cerinman: In this case I'm cool with them 👍

src/components/BarChart.js
cerinman: Yes, I was very surprised when I saw this.  Very glad to see it removed.  👍

docs/components/BarChartDocs.js
jmophoto: Kind of a standard in D3 to use `d`. All about the branding.

src/components/BarChart.js
cerinman: Hmm interesting question.  I can't imagine a scenario where I would want changing a function prop to trigger a render.

src/components/BarChart.js
```diff
@@ -1,299 +1,351 @@
...
 
-  componentWillReceiveProps (nextProps) {
-    if (nextProps.value !== this.props.value) {
-      this.setState({
-        milliseconds: 0
-      });
-    }
+  shouldComponentUpdate (nextProps) {
+    // check the props that aren't functions
+    return !_isEqual(_omit(nextProps, _functions(nextProps)), _omit(this.props, _functions(this.props)));
```
paniclater: I think this comment is unneccesary, _omit and _functions are pretty clear.

src/components/BarChart.js
```diff
@@ -1,299 +1,351 @@
...
   },
 
-  tick () {
-    this.setState({
-      milliseconds: this.state.milliseconds + 10
-    });
+  componentDidUpdate () {
+    if (this.props.hovering && this.props.animateOnHover) {
```
paniclater: nice job grouping side effects in componetDidUpdate.

src/components/BarChart.js
```diff
@@ -1,299 +1,351 @@
...
+    return `translate(${sx},${sy}) scale(1, ${factor})`;
   },
 
-  _handleMouseOut () {
-    this.setState({
-      hovering: false
-    });
-
-    this.props.onMouseOut();
+  _drawPath ({ x, y, width, height, value, radius }) {
```
paniclater: i think this is great, makes it very clear what's going on in the path.

src/components/BarChart.js
```diff
@@ -1,299 +1,351 @@
...
+          .duration(100)
+          .attr('transform', 'scale(1)');
+    }
   },
 
-  _handleMouseOver (label, value, x, y) {
-    const animateDuration = 500;
+  _getTransformWithScale (factor) {
+    const centerX = 0;
+    const centerY = (this.props.value > 0) ? this.props.y + this.props.height : this.props.y;
```
paniclater: why do you only add height if y is positive?

src/components/BarChart.js
```diff
@@ -1,299 +1,351 @@
...
-        onMouseOut={this._handleMouseOut}
-        onMouseOver={this._handleMouseOver.bind(null, this.props.label, this.props.value, this.props.x, y)}
-        rx={3}
-        ry={3}
-        style={style}
-        width={this.props.width}
-        x={this.props.x}
-        y={y}
+      <path
+        d={this._drawPath({
```
paniclater: so much better than seeing a bunch of logic directly in the d attribute.

src/components/BarChart.js
phantomxc: It has to do with the direction of the transform that needs to happen for the intro animation and the bounce animation.

ezgif com-video-to-gif 15

paniclater: I don't fully understand, but I believe the gif 💯

src/components/BarChart.js
paniclater: Mixins were the main thing preventing me from substituting React for Preact in the open source. I think if we could get rid of all the mixins we could probably substantially lower our bundle size with Preact.

phantomxc:
I believe I fixed all the concerned items.

cerinman:
:shipit:

phantomxc:
Could I get another peak into the last two commits? Found some issues.

src/components/BarChart.js
lukeschunk: oh also switching to ES6 classes might help with the preact switch. just sayin

src/components/BarChart.js
```diff
@@ -168,7 +169,7 @@ const BarChart = React.createClass({
 
       transform = `translate(${x}, ${y})`;
     } else {
-      transform = 'translate(-100, -100)';
+      transform = 'translate(-1000, -1000)';
```
lukeschunk: why the changes here?

phantomxc: This hides the tooltip further out of the SVG viewbox. When on a small screen, there were situations where it should show like a pixel of the tooltip.

lukeschunk:
:shipit:

phantomxc:
Third times a charm right? I didn't account for a proper domain when all the negative numbers were the same. 
༼ノಠل͟ಠ༽ノ ︵ ┻━┻

phantomxc:
Can someone take a look at my last commit?

phantomxc:
image

lukeschunk:
:shipit:

lukeschunk:
:shipit:

PR #504, opened 2017-01-30
Luke/code mod exploration

lukeschunk:
I used a [react code-mod](https://github.com/reactjs/react-codemod) (specifically the class-transform script) to convert our `createClass` to ES6 classes. 

The way the script changes the code means I also had to add the babel `stage-0` preset and the `transform-class-properties` which allows us to use the static class properties. 

I think ES6 class components is something we should adopt, and the codemod is a good way to do it. That said, I'm still doing testing and research around this to make sure it's 👌 

giphy-29

### Benefits 
* [facebook suggests](https://facebook.github.io/react/blog/2015/03/10/react-v0.13.html) the eventual removal of createClass in favor of ES6 classes
* because of this we can use current and future features of react 
* more vanilla JS and less library-specific 'magic'
* This could help us move towards purer components

### Possible trickiness 
* Functions aren't bound to `this` like they were before, from an article I read: 
> With createClass, it’s easy: every property that’s a function is automatically bound by React. Refer to them as this.whateverFn wherever you need to, and this will be set correctly whenever they’re called.

>With ES6 class, it’s trickier: functions are not autobound. 

this change will mean thinking about things a little differently as/if we move forward. 
* The static class properties are really new to me and i'm still wrapping my head around them - not sure if others feel the same way


paniclater:
I think we should definitely embrace this! Good job taking the initiative. The other big syntax thing I would like to change is es6 imports. Static class properties can be a little daunting, but I think after we use them a few times it will become second nature.

paniclater:
Looks like docs bundle got updated as well, not sure if that was by intention or not?

paniclater:
Looks like docs bundle got updated as well, not sure if that was by intention or not?

paniclater:
Otherwise looks good to me. We should run the code mod with our internal projects as well so we don't have competing patterns.

lukeschunk:
@paniclater i'll get rid of the bundle & fix merge conflicts. and yes i'm looking forward to running this internally as well. 🔮 🏀

lukeschunk:
RIP big pr

PR #505, opened 2017-02-01
Fixing Responsive Date Range Picker

paniclater:
Issue:

The date picker will expand to 100% of container on smaller screen sizes / devices, causing the flex wrap of the days and weeks to come completely out of alignment.
This sets a maxWidth to prevent that from happening while maintaining the full width design on xsmall devices / screens.

Before:
ezgif com-video-to-gif 23

After:
ezgif com-video-to-gif 22

iPhone 5:
screen shot 2017-01-31 at 6 06 34 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "dist/Index.js",
   "scripts": {
     "release": "eslint src && babel src --out-dir dist",
-    "dev": "cd docs && yarn && open 'http://localhost:8080' && NODE_ENV=development webpack-dev-server --inline --hot",
+    "dev": "cd docs && yarn && open 'http://localhost:8080' && NODE_ENV=development webpack-dev-server --inline --hot --host 0.0.0.0",
     "docs:release": "cd docs && rm -rf node_modules && yarn && NODE_ENV=production webpack --progress && git add -A && git commit -m \"Docs Version: $(node -p 'require(\"../package.json\").version')\"",
     "test": "eslint src"
   },
--- a/src/components/DateRangePicker.js
+++ b/src/components/DateRangePicker.js
@@ -447,14 +447,15 @@ const DateRangePicker = React.createClass({
         marginTop: isLargeOrMediumWindowSize ? 10 : 5,
         position: 'absolute',
         left: this.props.isRelative ? 'auto' : 0,
-        right: isLargeOrMediumWindowSize ? 0 : 'auto',
+        right: 0,
         zIndex: 10
       },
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -5,7 +5,7 @@
   "main": "dist/Index.js",
   "scripts": {
     "release": "eslint src && babel src --out-dir dist",
-    "dev": "cd docs && yarn && open 'http://localhost:8080' && NODE_ENV=development webpack-dev-server --inline --hot",
+    "dev": "cd docs && yarn && open 'http://localhost:8080' && NODE_ENV=development webpack-dev-server --inline --hot --host 0.0.0.0",
```
paniclater: this just allows us to view the docs from a vm or device via ip instead of just localhost.

src/components/DateRangePicker.js
```diff
@@ -447,13 +447,14 @@ const DateRangePicker = React.createClass({
...
         left: this.props.isRelative ? 'auto' : 0,
-        right: isLargeOrMediumWindowSize ? 0 : 'auto',
+        right: 0,
         zIndex: 10
       },
       calendarWrapper: {
         boxSizing: 'border-box',
         padding: isLargeOrMediumWindowSize ? 20 : 10,
         margin: 'auto',
+        maxWidth: 250,
```
paniclater: this prevents the wrapping issue.

src/components/DateRangePicker.js
```diff
@@ -546,6 +547,7 @@ const DateRangePicker = React.createClass({
         fontSize: StyleConstants.FontSizes.MEDIUM,
         paddingLeft: isLargeOrMediumWindowSize ? 10 : 0,
         paddingTop: 10,
+        maxWidth: 250,
```
paniclater: as does this

src/components/DateRangePicker.js
```diff
@@ -447,13 +447,14 @@ const DateRangePicker = React.createClass({
         marginTop: isLargeOrMediumWindowSize ? 10 : 5,
         position: 'absolute',
         left: this.props.isRelative ? 'auto' : 0,
-        right: isLargeOrMediumWindowSize ? 0 : 'auto',
+        right: 0,
```
paniclater: Really no matter what we want the calendar anchored to the right. On very small devices anchoring to the right 0 and left 0 allows the 100% width effect without the flex wrap breaking.

jmophoto:
Looks like a great change 🍪 :shipit:

PR #507, opened 2017-02-06
SimpleSelect supports children over the items prop.

phantomxc:
If the SimpleSelect has children, render the children instead of the items from the items prop. This will allow me to have a native onClick function to open a window or even populate it with anchor tags.


jmophoto:
:shipit:

PR #510, opened 2017-02-10
luke/Calendar codemod ran

lukeschunk:
Converts the Calendar component's createClass to ES6 classes using the codemod.


jtanner:
:shipit:

PR #511, opened 2017-02-17
Moore/css tweaks for daterangepicker

paniclater:
This PR: 
 - converts `_render*` functions to components with prop type checking.
 - adds a check to display the datepicker calendar and options wrappers at 100% when less than 450px wide.

I talked with @derek-boman and we came up with this as a best solution for issues the datepicker is currently having with smaller devices.

ezgif com-video-to-gif 25


src/components/DateRangePicker.js
```diff
@@ -448,6 +482,7 @@ const DateRangePicker = React.createClass({
         position: 'absolute',
         left: this.props.isRelative ? 'auto' : 0,
         right: 0,
+        width: window.innerWidth < 450 && '100%',
```
cerinman: Will this throw a css error or warning by assigning width to false in the case that it's greater than 450?

src/components/DateRangePicker.js
```diff
@@ -547,7 +582,7 @@ const DateRangePicker = React.createClass({
         fontSize: StyleConstants.FontSizes.MEDIUM,
         paddingLeft: isLargeOrMediumWindowSize ? 10 : 0,
         paddingTop: 10,
-        maxWidth: 250,
+        maxWidth: window.innerWidth > 450 && 250,
```
cerinman: Same question here.

paniclater: I'm not seeing any css errors, the style just gets applied when the condition is met and doesn't get applied when the condition isn't met. Maybe better though is to just default it to the inherited property from div if it isn't applied? I'll do that.
with condition met:
screen shot 2017-02-16 at 6 10 04 pm
with condition not met:
screen shot 2017-02-16 at 6 09 41 pm

paniclater: turns out figuring out how to set those default values was harder than I anticipated. I pushed back up with the `&&` I think its fine since the values are just left out when the object is converted, but I'm open to suggestions on alternate solutions!

cerinman: I'm with leaving it the way it is.  I just thought we just double check cause I wasn't sure how the browser would feel about assigning a boolean to property like that.  Thanks for double checking.

cerinman:
:shipit:

jtanner:
Should these responsive style updates also be applied to `DatePicker`?

paniclater:
Unfortunately this didn't solve all of my problems. I'll be putting up some changes later to get this working.

paniclater:
@jtanner I kinda want to deprecate the other datepickers and add a configuration to this one so it can support single date selections etc.

cerinman:
:shipit:

jtanner:
@paniclater that's an excellent idea

PR #513, opened 2017-02-21
luke/updated color in Select

lukeschunk:
@derek-boman noticed this when working internally. When an option is selected, it had the `PRIMARY` color by default, this makes it the same as what the hover color is.

Issue: 

screen shot 2017-02-21 at 12 57 31 pm


jtanner:
:shipit:

PR #517, opened 2017-02-24
added medium font size

lukeschunk:
When derek wants medium sized font on the message box title, you put a medium sized font on the message box title.


jtanner:
:shipit:

PR #520, opened 2017-03-03
Luke/date picker createclassconversion

lukeschunk:
DatePicker related components and docs createClass conversion. Did some testing on each of these and everything looks 👌


jtanner:
:shipit:

PR #521, opened 2017-03-03
removes margin on buttons in groups

iheartrachie:
This fixes a bug where Safari was adding margins to buttons in groups, breaking the borders.

### Before
screen shot 2017-03-03 at 11 17 41 am

### After
screen shot 2017-03-03 at 11 17 54 am


jtanner:
:shipit:

PR #523, opened 2017-03-04
removes overflow rule

iheartrachie:
Fixes a Safari bug that was cutting off modals inside of drawers on large displays.

## Before
screen shot 2017-03-03 at 4 55 14 pm

## After
screen shot 2017-03-03 at 4 55 31 pm


redperk:
Just to be sure, this doesn't introduce bugs in other browsers right?

iheartrachie:
I tested this in Firefox and Chrome and I didn't see any issues there.

lukeschunk:
:shipit:

PR #528, opened 2017-03-23
Luke/modal accessible

lukeschunk:
First cut at making modal more accessible 
* Changes DOM order - renders the close `<Icon />` last as that is a more natural flow when tabbing through modal 
* makes close icon tabable, and has the screen reader say "Close Modal" when that element is focused 
* Focus Modal component content by default on componentDidMount


lukeschunk:
zxyzxygif

jtanner:
:shipit:

PR #530, opened 2017-03-29
DateRangePicker IE 11 width issue

cerinman:
The options wrapper for the DateRangePicker was set to `inherit` it's parent's width if the window width was larger than 450.  This caused an issue in IE 11 where the the options wrapper would not expand to large enough for it's content.  This PR resolves the issue by setting a max width on the wrapper to 450 and then just using the windows width for width.

### Before

screen shot 2017-03-29 at 11 54 57 am

### After

screen shot 2017-03-29 at 11 47 08 am

screen shot 2017-03-29 at 11 48 04 am

### DIFF
--- a/docs/components/Changelog.js
+++ b/docs/components/Changelog.js
@@ -5,6 +5,13 @@ const Changelog = React.createClass({
     return (
       <div>
         <h1>Change Log</h1>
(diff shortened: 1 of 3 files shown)


jmophoto:
:shipit:

PR #531, opened 2017-03-29
DisplayInput > Sets line height for input to 1

cerinman:
Ensures IE 11 doesn’t clip text in input for letters such as `q` and `p`.

screen shot 2017-03-29 at 2 32 44 pm


vintagepenguin:
pretty straight forward :shipit:

PR #532, opened 2017-03-31
first cut drawer accessibility, focus and arialabel

lukeschunk:
First iteration of making Drawer More Accessible. 

There may be more work to do here. This is the start and does two things: 
* Focuses the Drawer when it is mounted 
* Adds an aria-label and a dialog role so the screenreader will read "{{drawertitle}} dialog" on mount


lukeschunk:
draweraccessibility_1

PR #533, opened 2017-03-31
BarChart: add threshold

jmophoto:
This adds an optional threshold line to the bar chart. 

```
      <BarChart
        data={data}
        margin={margins}
        threshold={threshold}
        xAxis={xAxis}
      />
```

screen shot 2017-03-31 at 5 11 21 pm

### DIFF
--- a/docs/components/BarChartDocs.js
+++ b/docs/components/BarChartDocs.js
@@ -83,6 +83,7 @@ const BarChartDocs = React.createClass({
           data={chartData}
           margin={margins}
           style={styles.chart}
+          threshold={Math.round(Math.random() * 10000)}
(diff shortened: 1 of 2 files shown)


docs/components/BarChartDocs.js
```diff
@@ -83,6 +83,7 @@ const BarChartDocs = React.createClass({
           data={chartData}
           margin={margins}
           style={styles.chart}
+          threshold={Math.round(Math.random() * 10000)}
```
paniclater: Should this be a floor or ceiling so there's predictable results?

paniclater: Actually check that this is just the docs! looks fine.

PR #535, opened 2017-04-06
Bar Chart minBarHeight

iheartrachie:
This adds `minBarHeight` to the `BarChart` component. If a value of 0 is passed to the chart, the chart will display a sliver of the bar, the height being the number used for `minBarHeight`. This sliver includes a tooltip. If no `minBarHeight` is set, so bar is displayed.

## Positive only values
screen shot 2017-04-06 at 10 12 14 am

## Negative only values
screen shot 2017-04-06 at 10 11 35 am

## Positive and negative values
screen shot 2017-04-06 at 10 11 55 am

## No `minBarHeight` passed in
screen shot 2017-04-06 at 10 44 48 am

### DIFF
--- a/src/components/BarChart.js
+++ b/src/components/BarChart.js
@@ -14,7 +14,7 @@ const Bar = React.createClass({
     hasPositive: React.PropTypes.bool,
     height: React.PropTypes.number,
     hovering: React.PropTypes.bool,
-    minHeight: React.PropTypes.number,
+    minBarHeight: React.PropTypes.number,
     onClick: React.PropTypes.func,
     onMouseOut: React.PropTypes.func,
     onMouseOver: React.PropTypes.func,
@@ -65,7 +65,7 @@ const Bar = React.createClass({
   },
 
   _drawPath ({ x, y, width, height, value, radius }) {
-    if (value > 0 || (value === 0 && !this.props.hasNegative)) {
+    if (value >= 0 && !this.props.hasNegative) {
       return 'M' + x + ',' + y +
          'h' + (width - radius) +
          'a' + radius + ',' + radius + ' 0 0 1 ' + radius + ',' + radius +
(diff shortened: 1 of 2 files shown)


jmophoto:
From the screen shots, it looks like the baseline edge of the zero bar on only-positive and -negatives might be concave? Maybe look at the `<=` and `>=` conditionals on lines 68 and 77.

iheartrachie:
@jmo Turns out the minBarHeight can't be less than the radius. I updated the logic and docs to account for this. Since the default value of the border radius is 3, the minBarHeight will probably be 3 in most places.

src/components/BarChart.js
```diff
@@ -10,8 +10,11 @@ const Bar = React.createClass({
   propTypes: {
     animateOnHover: React.PropTypes.bool,
     animationDuration: React.PropTypes.number,
+    hasNegative: React.PropTypes.bool,
+    hasPositive: React.PropTypes.bool,
```
phantomxc: I don't think think this is necessary. It's being computed and stored on the state down below.

iheartrachie: It's on the state for the BarChart, still need to pass the prop down to the Bar component.

jmophoto:
Wondering if you could do a check in the `_drawPath` method.

```
const r = this.props.minHeight ? Math.min(radius, this.props.minHeight) : radius;
```

src/components/BarChart.js
phantomxc: Sorry, forgot about that little booger.

iheartrachie:
Synced up with @jmo offline and came up with a solution for the radius issue. If `minBarHeight` is less than `barRadius` than the radius for that bar is set to `minBarHeight`. In a majority of cases, the bar will be so small that any radius can't be seen anyway but this will at least prevent the weird devil horns we were seeing easlier 😈

jmophoto:
:shipit:

PR #536, opened 2017-04-06
SimpleInput shift+tab accessibility fix

jtanner:
Typing `shift+tab` when the `SimpleInput` had focus was not working, it kept the focus on the input instead of going to the previous focusable element. `tabIndex` on the parent element turned out to be the problem. Removing that attribute and moving the focus/blur handlers to the input fixes the issue.

This also adds `devtool` when developing for better debugging (separate files in the dev tools makes it easier to set breakpoints).


lukeschunk:
:shipit:

PR #537, opened 2017-04-07
added focustrap library and component

lukeschunk:
Adds the [focus-trap-react](https://github.com/davidtheclark/focus-trap-react) library to the modal for accessibility. 

Also looked into other options like this https://github.com/vigetlabs/react-focus-trap 

But I think i like the one I used best.


cerinman:
:shipit:

jtanner:
:shipit:

PR #540, opened 2017-04-12
Adds depratated warn to a styles util

iheartrachie:
Since this is something I'd like to use in the `SearchInput` component I figured I'd move it to a util.


jmophoto:
:shipit:

PR #541, opened 2017-04-12
Luke/VisuallyHidden Button Children - Drawer Implementation

lukeschunk:
This helps make `<Button />` more accessible by allowing to pass children that are visuallyHidden from the user. This is necessary for icon-only buttons - with these buttons currently, there is no indication to the user what the button is. 

Also I added an implementation of this to the `<Drawer />` component's back button 

This seems the recommended way of making accessible Icon-Only buttons. 

References: 
https://egghead.io/lessons/css-accessible-icon-buttons
https://blog.jayway.com/2015/03/03/making-icon-only-buttons-a-little-more-accessible/

The hairiest part of this PR is the `_hasVisibleChildren` method - comment with suggestions on any improvements with that implementation or other ideas.

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -12,6 +12,7 @@ const { buttonTypes } = require('../constants/App');
 class Button extends React.Component {
   static propTypes = {
     actionText: React.PropTypes.string,
+    ariaLabel: React.PropTypes.string,
     icon: React.PropTypes.string,
     isActive: React.PropTypes.bool,
     onClick: React.PropTypes.func,
@@ -58,6 +59,7 @@ class Button extends React.Component {
 
     return (
       <button
+        aria-label={this.props.ariaLabel}
         onClick={this.props.type === 'disabled' ? null : this.props.onClick}
         style={Object.assign({}, styles.component, styles[this.props.type], this.props.style)}
       >
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -34,8 +34,27 @@ class Button extends React.Component {
     return windowSize === 'medium' || windowSize === 'large';
   };
 
+  _hasVisibleChildren = () => {
+    let hasVisibleChildren = false;
+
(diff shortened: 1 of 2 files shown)


vintagepenguin:
:shipit:

src/components/Drawer.js
```diff
@@ -163,7 +163,9 @@ const Drawer = React.createClass({
                     onClick={this.close}
                     primaryColor={this.props.buttonPrimaryColor}
                     type={'base'}
-                  />
+                  >
+                    <span style={styles.visuallyHidden}>Close Drawer</span>
```
jtanner: Looks like this should also have `className='visuallyHidden'` because that's how you're looking for it in Button.js.

src/components/Button.js
```diff
@@ -34,6 +34,26 @@ class Button extends React.Component {
...
+    }
+
+    this.props.children.forEach((child) => {
+      if (child.props && child.props.className !== 'visuallyHidden') {
+        hasVisibleChildren = true;
+      }
+    });
+
+    return hasVisibleChildren;
+  };
```
jtanner: Here's a slightly shorter version:

```javascript
  _childIsVisible = child =>
    !child.props || child.props.className !== 'visuallyHidden';

  _hasVisibleChildren = () => {
    if (!this.props.children) {
      return false;
    }

    if (!Array.isArray(this.props.children)) {
      return this._childIsVisible(children);
    }

    return this.props.children.some(this._childIsVisible);
  };
```

src/components/Drawer.js
lukeschunk: ahh good catch

src/components/Button.js
lukeschunk: ah nice `.some`

lukeschunk:
@jtanner ready for another look

PR #544, opened 2017-04-17
SimpleInput: fix onblur

jmophoto:
With the addition of the `_onBlur` method, any `onBlur` prop sent in would be overridden. This adds a check for that prop and calls the method from the `_onBlur`.


jtanner:
:shipit:

PR #545, opened 2017-04-17
SimpleInputl: add event to onBlur and add onFocus

jmophoto:
This passes the event to the `onBlur` prop and also adds the same check for onFocus.


jtanner:
:shipit:

PR #547, opened 2017-04-17
add tabIndex to modal content so trap-focus is happy and there is alw…

lukeschunk:
…ays focusable content.

`trap-focus-react` throws an error if there is no focusable content. Adding a `tabIndex={0}` to the modal content so even if there is a modal that is only text elements, the content is still focusable.


phantomxc:
:shipit:

PR #548, opened 2017-04-18
adds arialabel prop, focus content on mount

lukeschunk:
After working with this Modal a bit, I think adding an `ariaLabel` prop and focusing the content rather than the outer container of the modal on mount is a better pattern and more accessible.

### DIFF
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -8,6 +8,7 @@ const StyleConstants = require('../constants/Style');
 
 const Modal = React.createClass({
   propTypes: {
+    ariaLabel: React.PropTypes.string,
     buttons: React.PropTypes.arrayOf(React.PropTypes.shape({
       actionText: React.PropTypes.string,
       className: React.PropTypes.string,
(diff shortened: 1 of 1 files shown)


src/components/Modal.js
```diff
@@ -190,12 +192,13 @@ const Modal = React.createClass({
           <div className='mx-modal-scrim' onClick={this.props.onRequestClose} style={Object.assign({}, styles.scrim, styles.overlay, this.props.isRelative && styles.relative)}></div>
           <div
             className='mx-modal-container'
-            ref={ref => this._modal = ref}
             style={Object.assign({}, styles.container, this.props.style)}
           >
             {this._renderTitleBar()}
             <div
+              aria-label={this.props.ariaLabel}
```
jtanner: What does the `aria-label` do in this case? Describe the modal as a whole?

lukeschunk: that is the intent - yes.

jtanner: Do you think the `aria-label` will generally be the same as the modal title? If so, I wonder if using `aria-labelledby` to point to the title would be better because the existing modals we use would not need to change.

lukeschunk: hmm. we could potentially add some logic to do something like this if the title exists but in a lot of places internally we do not use a title.

jtanner:
:shipit:

PR #550, opened 2017-04-19
luke/DonutChart - createClass Conversion

lukeschunk:
Convers `DonutChart.js` and `DonutChartDocs.js` to Class extends syntax.


paniclater:
:shipit:

paniclater:
:shipit:

PR #551, opened 2017-04-19
luke/Drawer - createClass conversion

lukeschunk:
Converts `Drawer.js` and `DrawerDocs.js` to Class extends syntax.


paniclater:
:shipit:

paniclater:
:shipit:

PR #552, opened 2017-04-26
Luke/file upload create class conversion

lukeschunk:
Create Class conversion for the `FileUpload.js` and `FileUploadDocs.js`. 

Also - i noticed that the FIleUpload docs weren't linked in the sidebar. I added it to the docs, and added some text to the demo usage to make it more obvious how to run the demo.


paniclater:
:shipit:

paniclater:
:shipit:

PR #554, opened 2017-04-27
adding calling prop handler display value to date range picker

paniclater:
this allows for the date range select handler to receive the displayValue as a prop so we can maintain month, week, etc selections.


jtanner:
:shipit:

PR #555, opened 2017-04-27
DisplayInput children styles

jtanner:
This PR:
* adds a `childrenStyle` prop to `DisplayInput` for customizing the `children` wrapping `div`
* updates the styles for `children` to match the font and size of the default input

Here's what it looks like in the docs:

screen shot 2017-04-27 at 11 11 03 am


lukeschunk:
:shipit:

PR #556, opened 2017-04-28
SimpleSelect accessibility using Listbox/Option

jtanner:
Pulling in the `Listbox`/`Option` components from our internal raja project. I put it in `components/accessibility/` because I can imagine us adding more accessibility related components.

I also updated `SimpleSelect` to use the new components. I'm happy with how easy it was to hook up. Here's an example of using it with the `enter` and `up`/`down` keys:

ezgif com-video-to-gif

### DIFF
--- a/src/components/SimpleSelect.js
+++ b/src/components/SimpleSelect.js
@@ -3,11 +3,13 @@ const Radium = require('radium');
 const _merge = require('lodash/merge');
 
 const Icon = require('./Icon');
+const { Listbox, Option } = require('./accessibility/Listbox');
 
 const StyleConstants = require('../constants/Style');
 
 const SimpleSelect = React.createClass({
   propTypes: {
+    'aria-label': React.PropTypes.string,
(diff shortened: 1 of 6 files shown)


src/components/SimpleSelect.js
```diff
@@ -48,33 +50,38 @@ const SimpleSelect = React.createClass({
...
   },
 
   render () {
     const styles = this.styles();
 
     return (
       <div style={styles.component}>
-        <div style={styles.menu}>
+        <Listbox
+          aria-label={this.props['aria-label']}
```
lukeschunk: let me know what your thinking on the `aria-label` vs `ariaLabel` is. i mentioned internally i prefer the ariaLabel but I could be convinced if you have strong feelings towards the other

lukeschunk:
just the one comment, otherwise this looks great. 🏀

lukeschunk:
:shipit:

PR #558, opened 2017-05-01
Add esc key handling to SimpleSelect

jtanner:
I tested this with the example in the docs and it's working great 😄


cerinman:
:shipit:

PR #559, opened 2017-05-01
Use a label tag for the DisplayInput label

jtanner:
* Use an actual `label` tag for the label
* Associate the `label` with the `input` using auto generated ids
* Use `aria-labelledby` instead of `label` (which isn't a valid attribute)

Now clicking on the `label` will focus the `input` making `DisplayInput` more accessible.


cerinman:
:shipit:

PR #560, opened 2017-05-02
Luke/react 15.5 upgrade

lukeschunk:
* Updates`eslint-config-mx` `react` and `react-dom` to latest package versions.
* Changes references to `React.PropTypes` to use the `prop-types` npm package instead. 

These changes will throw warnings for create-class and React.PropTypes being deprecated, but I think it's better to chip away at these fixes incrementally. Next step is to update dependencies to get the proptypes warning gone.


docs/components/DrawerDocs.js
```diff
@@ -212,9 +212,9 @@ const DrawerDocs = React.createClass({
     }
 
     HeaderMenu.propTypes = {
-      handleButtonClick: React.PropTypes.func,
-      handleScrimClick: React.PropTypes.func,
-      showSimpleSelectMenu: React.PropTypes.bool
+      handleButtonClick: PropTypes.func,
+      handleScrimClick: PropTypes.func,
+      showSimpleSelectMenu: PropTypes.bool
```
jtanner: `PropTypes` didn't get required in this file.

lukeschunk: This is just markdown for the docs, but i feel like it's probably better to do this in our markdown anyway? maybe out of the scope of this PR though... what do you think

jtanner: Never mind, I didn't realize it was markdown.

jtanner:
Conflicts ⚠️

jtanner:
:shipit:

PR #561, opened 2017-05-02
Rename ariaLabel to aria-label to match React's naming conventions

jtanner:
Rename `ariaLabel` to `aria-label` to match React's naming conventions.


lukeschunk:
:shipit:

PR #564, opened 2017-05-03
loader and loader docs class extends conversion

lukeschunk:
Converts `Loader` and `LoaderDocs` to class extends syntax.


paniclater:
:shipit:

PR #565, opened 2017-05-03
Luke/class extends conversion src

lukeschunk:
Class extends syntax conversion for all components in `/src/components`. Everything here was pretty straightforward - had to do a little bit of manual work in `Listbox` and `DatePickerFullScreen` - check out my last two commits. 

Basically, using a static `state` wont work if we reference props or other class functions from state, so I used the `constructor()` 

Clicked through the docs and tested each component out.

### DIFF
--- a/src/components/BarChart.js
+++ b/src/components/BarChart.js
@@ -26,7 +26,7 @@ class Bar extends React.Component {
     y: PropTypes.number
   };
 
-  componentDidMount() {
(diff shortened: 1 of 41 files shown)


jtanner:
The related docs conversions were done in https://github.com/mxenabled/mx-react-components/pull/566

jtanner:
:shipit:

PR #566, opened 2017-05-04
luke/class extends conversion on /docs/components

lukeschunk:
`createClass` to class extends syntax in `/docs/`. Everything here was straightforward - the only manual change I had to do was in `NotifyOnScrollThresholdDocs` where I added the `constructor` for setting initial state.


jtanner:
:shipit:

PR #569, opened 2017-05-05
yarn && yarn upgrade all the things

jtanner:
We have 2 package.json files, one for the project and one for the docs. I think `yarn dev` should run `yarn && yarn upgrade` for both and `yarn docs:release` should run `yarn` for both. This will ensure that the `yarn.lock` files stay up to date.


lukeschunk:
:shipit:

PR #570, opened 2017-05-05
Fix bug where a custom input id was ignored

jtanner:
Passing `id` with `elementProps` was being ignored (introduced in https://github.com/mxenabled/mx-react-components/pull/559). This fixes the problem by checking for an id before generating a unique id to use with the `label`.


lukeschunk:
:shipit:

PR #571, opened 2017-05-08
Select accessibility updates

jtanner:
* Use `Listbox` for the options' accessibility
* Handle `esc` to close
* Handle `enter`/`space` to open (when Select is focused)
* Give visual indication that the component is focused
* Use css hover instead of state
* Fix propagation bug with `Listbox`

Here's a demo of using only the keyboard to interact with `Select`:

ezgif com-video-to-gif

### DIFF
--- a/src/components/Select.js
+++ b/src/components/Select.js
@@ -1,8 +1,9 @@
 const _isEqual = require('lodash/isEqual');
-const React = require('react');
-const ReactDOM = require('react-dom');
+const keycode = require('keycode');
 const PropTypes = require('prop-types');
 const Radium = require('radium');
+const React = require('react');
+const ReactDOM = require('react-dom');
 
 const Icon = require('./Icon');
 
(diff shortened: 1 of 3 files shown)


src/components/Select.js
```diff
@@ -35,95 +37,43 @@ class Select extends React.Component {
...
-      };
-    } else {
-      return null;
+  _handleKeyDown = (e) => {
+    switch (keycode(e)) {
+      case 'esc':
+        e.preventDefault();
+        e.stopPropagation();
+        this._close();
+        this.component.focus();
```
iheartrachie: Is this needed since _close() calls focus?

jtanner: No, good catch, I meant to remove that 👍

jtanner: I removed it.

iheartrachie:
:shipit:

redperk:
:shipit:

PR #573, opened 2017-05-09
adding npm-watch for dev workflow

paniclater:
I've been frustrated with my workflow while working with this repo in the context of fixing a bug in another repo that uses it.

Here's a workflow that has worked for me but it involves adding another dev dependency: `npm-watch`.

Using [yarn link](https://yarnpkg.com/lang/en/docs/cli/link/) and the npm-watch package you can get another application (for example our internal widgets) to pick up on changes in this repo as you develop. It is a little slow but way better imho than just rerunning webpack all the time.

Maybe others on the team already have a better solution, so just opening it up for discussion!

Description:
1. I added the npm package `npm-watch` and configured it to run the `release` script whenever javascript files in the `src` directory or its children change
1. Then I run `yarn link` in the mx-react-components` top level directory
1. Then I run `yarn link mx-react-components` in the top level directory of the application that depends on the components

At this point now the release script will run eslint and babel on the components every time source files change. The yarn link commands tell yarn to reference the local version of the package (without editing the package.json, adding or removing local deps with `add file` or anything else).

Simply running webpack as normal in the dependent application will pick up changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "eslint-plugin-react": "^5.0.0",
     "live-server": "^0.8.1",
     "node-libs-browser": "^0.5.2",
+    "npm-watch": "^0.1.9",
     "webpack": "^1.0.0",
     "webpack-dev-server": "^1.16.2"
   }
--- a/package.json
+++ b/package.json
@@ -3,11 +3,15 @@
   "version": "5.0.0-rc.87",
   "description": "A collection of generic React UI components",
   "main": "dist/Index.js",
+  "watch" {
+    "release": "src/**/*.js"
+  },
   "scripts": {
     "release": "eslint src && babel src --out-dir dist",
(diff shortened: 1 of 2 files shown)


jtanner:
@paniclater if I have linked this in another repo and run yarn or webpack is there an indication anywhere that I'm using a linked package? The reason I ask is because I don't want to link this and then forget to unlink it and be using an outdated package.

paniclater:
@jtanner I don't think there's a built in way of doing this.

`yarn unlink <package>` is the command that you use to undo the linking.

`yarn link` does create a sym link to the package on your system, so I found this suggestion for keeping track of all the sym links in your node_modules directory:

`ls -l node_modules | grep ^l`

cerinman:
I think @sambev just merged and released something similar.  Lets make sure we don't add duplicate scripts!

jtanner:
Closing now that we have `yarn dev-release`

PR #574, opened 2017-05-09
Tabs component alignment prop and style updates

jtanner:
While updating an internal project I noticed the following sad use of `Tabs` with default styling:
screen shot 2017-05-09 at 2 48 45 pm

The following changes update `Tabs` to be more in line with the [Aggro Design System docs for Tabs](https://company-23585.frontify.com/d/KHQZnlj1XMIO/web-patterns#/components/tabs)
* update the padding, colors, and hover styling
* remove uppercase styling
* add `alignment` prop for optionally aligning the tabs to the center (default is `'left'` for backward compatibility)
* add extra demo example of centered tabs

----
### Example of updated docs:

screen shot 2017-05-09 at 2 40 47 pm

----
### Example of hover changes:

ezgif com-video-to-gif

Issue #480, opened 2016-12-14
Tabs Refactor - Composability & Reusability 


lukeschunk:
Our `<Tabs />` Component could benefit from a rewrite/refactor. 

Right now we have several cases internally where the tabs do not necessarily work with a given use case which causes us to have to create one-off tab-like components. We should be able to use our tabs for all/most use cases. 

What needs to be done I think is more composability, and perhaps two components a `<Tabs>` and an individual `<Tab>`. I have done a bit of research on this but will add to this issue if need be.

derek-boman:
@jtanner has expressed interest in working on this recently. Here's some updated spec: https://company-23585.frontify.com/document/46684#/components/tabs

jtanner:
@derek-boman I created a new issue for adding the "pill" tab type: https://github.com/mxenabled/mx-react-components/issues/567

cerinman:
This has been completed.  Closing.

PR #575, opened 2017-05-10
Refactor Tabs to use a Tab component

jtanner:
This refactors `Tabs` to internally use a `Tab` component. The next step will be to make `Tab` available to use as a child of `Tabs` instead of using the `tabs` prop. See https://github.com/mxenabled/mx-react-components/issues/480

_NOTE: this targets the styling update branch of https://github.com/mxenabled/mx-react-components/pull/574_

### DIFF
--- a/src/components/Tabs.js
+++ b/src/components/Tabs.js
@@ -120,18 +120,12 @@ class Tabs extends React.Component {
 
   render () {
     const styles = this.styles();
+    const useTabs = this._isLargeOrMediumWindowSize() || this.props.useTabsInMobile;
+    const tabsStyles = Object.assign({}, styles.component, (useTabs && styles.tabsContainer), this.props.style);
 
     return (
-      <div style={Object.assign({}, styles.component, this.props.style)}>
-        {this._isLargeOrMediumWindowSize() || this.props.useTabsInMobile ? (
-          <div style={styles.tabsContainer}>
-            {this._renderTabs()}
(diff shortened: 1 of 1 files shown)


src/components/Tabs.js
```diff
@@ -159,31 +190,19 @@ class Tabs extends React.Component {
...
-        width: '100%'
-      }
+      }, this.props.activeTabStyles)
     };
-  };
+  }
 }
 
-module.exports = Radium(Tabs);
+const Tab = Radium(TabWithoutRadium);
```
iheartrachie: Why is `TabWithoutRadium` needed? Also, `TabWithoutRadium` uses Radium so the naming confuses me.

jtanner: The `Tabs` component uses `Tab` which needs Radium for the hover styles. I named it `TabWithoutRadium` to communicate that it still needs to be wrapped with the Radium higher order component before it can be used as the `Tab` component.

Do you think a different name would make it clearer?

jtanner: One option would be to move `Tab` to its own file and wrap it with Radium there. Do you think it would be less confusing that way?

iheartrachie: Oh okay I see what you mean now. Knowing that's the case, I can't really think of a clearer name either 😆  so I'll go ahead and ship this.

:shipit:

PR #576, opened 2017-05-12
Adds OTHER as a category color in Style constants

cerinman:
The other category color was missing from the Style constants Colors and we need it in place internally for theme calculation so this adds it here.  I made the same as the ASH color per instruction from Derek.


jtanner:
:shipit:

PR #578, opened 2017-05-18
Fix Select scrim click

jtanner:
https://github.com/mxenabled/mx-react-components/pull/571 broke the `Select`'s scrim click behavior. This fixes it by halting the click event from bubbling up and then immediately re-opening the drop down.

Also includes some yarn updates.


redperk:
:shipit:

Issue #567, opened 2017-05-04
Add "pill" type to Tabs


jtanner:
We need to add a "pill" type to `Tabs` to match the Aggro design: https://company-23585.frontify.com/document/46684#/components/tabs

Issue #581, opened 2017-05-24
Add *ref to components that can be focused etc.


jtanner:
Recently I was trying to focus a `Button` component while trying to improve the accessibility of another component and found that I couldn't focus it without updating the component. Ultimately I followed the suggestion in https://facebook.github.io/react/docs/refs-and-the-dom.html#exposing-dom-refs-to-parent-components to add a `buttonRef` prop which solved the issue nicely (thanks @phantomxc for finding the article).

I think we should make our other components that can be focused etc. more user friendly by adding a similar prop. Here's a list of components so we can keep track of which have been updated:

- [x] Button #584 
- [x] Date Picker
- [x] Date Picker Full Screen
- [x] Date Range Picker
- [x] Date Time Picker
- [x] Display Input
- [x] File Upload
- [x] Radio Button
- [x] Range Selector
- [x] Search Input
- [x] Select
- [x] Select Full Screen
- [x] Simple Input
- [x] Simple Select
- [x] Simple Slider
- [x] TextArea
- [x] Toggle Switch
- [x] Type Ahead

markstewy:
@jtanner Do we simply need a ref={this.props.myComponentNameRef} added to the **_parent div_** of each of these components along with a new proptype to match? 

Or do some components need multiple refs nested within? 

I've never contributed to open source before, but I would like to become a regular if there is any way I can add value.

jtanner:
@markstewy yes, similar to #584.

We'd love the help if you want to contribute ❤️

jtanner:
Completed with #699

PR #583, opened 2017-05-25
Add a 'pill' type to Tabs

jtanner:

- *Deprecate* the `useTabsInMobile` prop!
  I discussed this with @jmophoto and we thought it would be best to remove this surprising behavior and let the user decide when to switch between tab types.
- Add a `type` prop
- Add `buttonRef` to `Button` for manual focus (see #581)
- Extract `MenuTabs` and `StandardTabs` from `Tabs`

#### Demo

screencast 2017-05-25 12-13-35

#### Docs

screen shot 2017-05-25 at 10 48 04 am


src/components/Tabs.js
```diff
@@ -205,4 +128,184 @@ class TabWithoutRadium extends React.Component {
 
 const Tab = Radium(TabWithoutRadium);
 
+class MenuTabs extends StandardTabs {
```
lukeschunk: i believe this would be the first time we'd be extending from a component (that isn't React's base Component). i sometimes think inheritance gets very confusing, and maybe something like composition would work in this case? [this is a little old, but has some good concepts](https://discuss.reactjs.org/t/best-practices-for-extending-subclassing-components/1820/10)

jtanner: @lukeschunk I tend to favor composition over inheritance, but I felt that this was an excellent use case for inheritance. These polymorphic tab components are not exported and are very similar in behavior so I thought it made sense.

How would you go about refactoring this to use composition? Do you want to experiment with it and throw up a PR against this branch?

jtanner: I'll take a stab at it in another PR too.

jtanner:
I'm going to split this PR into a couple different PRs so it's more focused.

src/components/Tabs.js
paniclater: Its also ok if inheritance is used when it makes sense. While I also tend to not favor it, it is important not to get too dogmatic and use tools when they make sense. 

If we do go this route, maybe just a comment explaining why this pattern gets used here would be helpful for future developers and avoid them copying it by default without understanding the consequences elsewhere.

lukeschunk: i've had issues with inheritance in the past as well, but i agree that it's important not to get too dogmatic - if you think it's an excellent use case that's 🆒 . i'll take a look as well

jtanner:
@paniclater @lukeschunk I've taken some time to experiment with different styles of polymorphism. First I split off #586 from this PR and then I created the following PRs:

#604 
Inheritance - based on this PR

#605
Composition
- `TabsWrapper` manages tab state, styles, and type selection

#606 
Composition 2 (my preference)
- `TabsFactory` manages type selection
- `Tabbable` manages tab state and styles
- Separate files for each component

jtanner:
Closing in favor of #606

PR #586, opened 2017-05-31
Deprecate useTabsInMobile prop and tabs menu type for small screens

jtanner:
Showing the `Select`-like menu when the screen is small was a poorly documented feature and surprising behavior. These changes remove that extra feature and deprecate the `useTabsInMobile` prop making Tabs simpler and more focused by pushing the decision of whether to show a `Select` or `Tabs` to the container.

I discussed this with @jmophoto and we thought it would be best to remove this surprising behavior and let the user decide when to switch between tab types.

This was originally part of #583

### DIFF
--- a/docs/components/TabsDocs.js
+++ b/docs/components/TabsDocs.js
@@ -1,4 +1,5 @@
 const React = require('react');
+const { Link } = require('react-router');
 
 const { Tabs } = require('mx-react-components');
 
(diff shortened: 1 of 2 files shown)


lukeschunk:
:shipit:

jtanner:
@lukeschunk this is ready for another look 👀 

The tabs are now scrollable on small screens.

lukeschunk:
👍 :shipit:

Issue #588, opened 2017-05-31
Add theme support to the components


jtanner:
We need to add theming to our components which will allow customizing more than just the `primaryColor`. 

- [x] move functions in `constants/Style.js` to `utils/Style.js` #589 
- [x] add a theme merge function to `utils/Style.js` #592
- [x] move unused category colors and `LIGHT_*` colors to a different repo
- [x] deprecate unused colors #591 
- [x] add theme prop to all components #648 
- [x] remove primaryColor prop from all components #648 

### New theme names mapping
```javascript
    // Accent Colors
    primary: 'PRIMARY',
    success: 'LIME',
    warning: 'BANANA',
    danger: 'STRAWBERRY',
    attention: 'ORANGE',
    // Light Accent Colors
    light_success: 'LIGHT_LIME',
    light_warning: 'LIGHT_BANANA',
    light_danger: 'LIGHT_STRAWBERRY',
    // Gray Scale
    gray_100: 'PORCELAIN',
    gray_300: 'FOG',
    gray_500: 'ASH',
    gray_700: 'CHARCOAL',
    gray_900: 'SLATE',
```

jtanner:
Here are the components that need the `theme` prop:
- [ ] BarChart.js
- [ ] Button.js
- [ ] ButtonGroup.js
- [ ] Calendar.js
- [ ] DatePicker.js
- [ ] DatePickerFullScreen.js
- [ ] DateRangePicker.js
- [ ] DateTimePicker.js
- [ ] DisplayInput.js
- [ ] DonutChart.js
- [ ] Drawer.js
- [ ] FileUpload.js
- [ ] Gauge.js
- [ ] HeaderMenu.js
- [ ] Icon.js
- [ ] Loader.js
- [ ] Menu.js
- [ ] MessageBox.js
- [ ] Modal.js
- [ ] NotifyOnScrollThreshold.js
- [ ] PageIndicator.js
- [ ] PaginationButtons.js
- [ ] ProgressBar.js
- [ ] RadioButton.js
- [ ] RajaIcon.js
- [ ] RangeSelector.js
- [ ] SearchInput.js
- [ ] Select.js
- [ ] SelectFullScreen.js
- [ ] SimpleInput.js
- [ ] SimpleSelect.js
- [ ] SimpleSlider.js
- [ ] Spin.js
- [ ] Tabs.js
- [ ] TextArea.js
- [ ] TimeBasedLineChart.js
- [ ] ToggleSwitch.js
[...]

PR #589, opened 2017-06-01
Move functions in constants/Style to utils/Style

jtanner:
_Targets `joe/theme-master` branch_

To prepare for adding theme support (see #588) to the components this moves the functions in constants/Style to utils/Style.

Other notable changes:
- `getWindowSize` now takes `BreakPoints` as a parameter because this will need to come from the theme in the future
- `checkForDeprecated` was removed from the utils file so it's not exposed externally
- the `style` prop deprecation warnings were removed for now until we make those style related props consistent across all the components (only 3 of the 12 components that use `styles` had a warning)


lukeschunk:
:shipit:

Issue #587, opened 2017-05-31
Calendar dates are wrong and inconsistent


iheartrachie:
Right now we are using the component width to handle the length of weeks. This is causing weeks to be too long or too short in some cases and messing up the rest of the calendar.

![screen shot 2017-05-31 at 11 01 15 am](https://cloud.githubusercontent.com/assets/1916697/26644055/b9909b1e-45f0-11e7-83ea-5c88b6567458.png)

PR #590, opened 2017-06-01
Calendar week fix

iheartrachie:
Fixes: https://github.com/mxenabled/mx-react-components/issues/587

We were using the component width to set the number of days in a week which was causing the calendar to render weeks with too many or too few days. This PR restricts the week to a constant 7 days, preventing display issues when different CSS is applied. The calendar is now mostly responsive but enough width must be provided for the weeks to render. By default, there is no width.

## After (defualt, no width)
screen shot 2017-06-01 at 2 54 10 pm

## Narrow width - 200px
screen shot 2017-06-01 at 2 53 40 pm

## Wide width - 600px
screen shot 2017-06-01 at 2 53 56 pm

### DIFF
--- a/src/components/Calendar.js
+++ b/src/components/Calendar.js
@@ -149,9 +149,7 @@ class Calendar extends React.Component {
             );
           })}
         </div>
-        <div style={styles.calendarTable}>
-          {this._renderMonthTable()}
-        </div>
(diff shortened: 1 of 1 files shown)


jtanner:
:shipit:

PR #591, opened 2017-06-06
Theme deprecated colors

jtanner:
_Targets: `joe/theme-master` branch_

This renames the colors to the new semantic theme names and removes *most* non-theme colors. See #588.

Here's the updated docs:

screen shot 2017-06-06 at 3 34 32 pm


cerinman:
:shipit:

PR #592, opened 2017-06-07
Add theme to Button

jtanner:

This introduces a pattern for adding a `theme` prop to a component while keeping the `primaryColor` prop in place but deprecated.

Here's an example of the demo with themed buttons:

screen shot 2017-06-07 at 10 09 06 am

### DIFF
--- a/docs/app.js
+++ b/docs/app.js
@@ -41,6 +41,7 @@ const Spin = require('components/SpinDocs');
 const Styles = require('components/StylesDocs');
 const Tabs = require('components/TabsDocs');
 const TextArea = require('components/TextAreaDocs');
+const Theme = require('components/ThemeDocs');
 const TimeBasedLineChart = require('components/TimeBasedLineChartDocs');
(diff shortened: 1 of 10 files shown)


iheartrachie:
I don't know how this relates to other places we are using theme but I find customizing `GRAY_500` or any other color-specific name a little confusing. In the context of buttons, I think it makes more sense to have a `disabled` color variable in the theme. But again, not sure how that plays with other places we're using theme.

jtanner:
@iheartrachie that's a good point. Having a `disabled` color makes sense especially because the theme color names are meant to be more semantic and descriptive of how they're used. Let's discuss it more and if we decide to go that route I'll put up another PR 👌

iheartrachie:
:shipit:

jtanner:
@cerinman @iheartrachie I added some docs for theming. What do you think of the wording? Is is clear enough?

Here's what it looks like:

screen shot 2017-06-08 at 11 22 02 am

jmophoto:
:shipit:

PR #593, opened 2017-06-07
Rachel/input type

iheartrachie:
The `DisplayInput` component was setting the HTML input prop to text. Removing this will allow us to set the type of the HTML input as part of `elementProps`.


lukeschunk:
perhaps we should add a default key on `elementProps` to text?

iheartrachie:
Thanks for @lukeschunk for pointing out the default. Without setting the input type it was causing quite a few bugs.

jtanner:
:shipit:

PR #594, opened 2017-06-08
Add theme to ButtonGroup

jtanner:

* Adds theme to ButtonGroup
* Fixes bug with `adjustColor` where it didn't handle 3 character hex values

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -5,7 +5,6 @@ const Radium = require('radium');
 const Icon = require('./Icon');
 const Spin = require('./Spin');
 
-const StyleConstants = require('../constants/Style');
 const { buttonTypes, themeShape } = require('../constants/App');
 
 const StyleUtils = require('../utils/Style');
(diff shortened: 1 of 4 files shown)


src/components/Button.js
```diff
@@ -58,7 +56,8 @@ class Button extends React.Component {
   };
 
   render () {
-    const styles = this.styles();
+    const theme = StyleUtils.mergeTheme(this.props.theme, this.props.primaryColor);
```
lukeschunk: eventually i think we could kill `this.props.primaryColor` and just use the theme

lukeschunk:
:shipit:

src/components/Button.js
jtanner: Yes, that's the idea 😄 

See https://github.com/mxenabled/mx-react-components/pull/594/files#diff-734f11edfaa69e2a96b8cb8f7f06fb6bR34

And https://github.com/mxenabled/mx-react-components/pull/594/files#diff-ec85bd453c96420d8df62f4d96bd694aR33

lukeschunk: yeah i realized after reviewiing a few more of these that you were way ahead of me 🙃

PR #597, opened 2017-06-09
removed xsmall and in daterangepicker

lukeschunk:
The `xsmall` breakpoint was doing more harm then good, so this removes it and removes the reference to it in `DateRangePicker`.


lukeschunk:
Based on the way the `getWindowSize` is function, we could also probably get rid of the `small` from the breakpoint, since the function sets it to small by default. Let me know if you think we should remove [this](https://github.com/mxenabled/mx-react-components/pull/597/files#diff-88f04cd7239c747481130b6ea9b32c23L5)

src/constants/Style.js
```diff
@@ -155,8 +155,6 @@ module.exports = {
       windowSize = 'large';
     } else if (width >= breakPoints.medium) {
       windowSize = 'medium';
-    } else if (width <= breakPoints.small) {
-      windowSize = 'xsmall';
```
jtanner: Good idea to remove this 👍

jtanner:
@lukeschunk what harm was the `xsmall` size causing?

lukeschunk:
@jtanner sounds like it was causing issues anywhere we weren't accounting for `xsmall`. @phantomxc ran into it in his work internally

jtanner:
:shipit: but let's make sure to clean up any references to `xsmall` in other projects 👌

ztoolson:
:shipit:

PR #601, opened 2017-06-09
Add theme to Drawer

jtanner:

Add theme to Drawer

* clean up docs example
* make `HeaderMenu` work like a real component vs mostly hard-coded


lukeschunk:
:shipit:

PR #605, opened 2017-06-11
Add a 'pill' type to Tabs [COMPOSITION 1]

jtanner:
COMPOSITION style (see discussion in #583)
`TabsWrapper` manages tab state, styles, and type selection

I wasn't satisfied with this version so I created an alternative that I like better: #606

----

- Add a `type` prop
- Extract `StandardTabs` from `Tabs`


jtanner:
Closing in favor of #606

PR #606, opened 2017-06-11
Add a 'pill' type to Tabs [COMPOSITION 2]

jtanner:
COMPOSITION style 2 (see discussion in #583)
- `TabsFactory` manages type selection
- `Tabbable` manages tab state and styles
- Separate files for each component

I think this version turned out the best 😄 

----

- Add a `type` prop
- Extract `StandardTabs` from `Tabs`


paniclater:
I agree, I think this one takes the :cake:.

src/components/Tabs.js
```diff
@@ -1,235 +1,33 @@
...
+const StandardTabs = require('./tabs/StandardTabs');
 
-const TabsPropTypes = {
-  activeTabStyles: PropTypes.object,
-  brandColor: PropTypes.string,
-  onTabSelect: PropTypes.func.isRequired,
-  selectedTab: PropTypes.number,
-  tabs: PropTypes.array.isRequired
+const TabsTypes = {
+  standard: StandardTabs,
```
paniclater: maybe these type constant key names should be all capsed?

src/components/Tabs.js
```diff
@@ -1,235 +1,33 @@
...
-      ':focus': {
-        backgroundColor: StyleConstants.Colors.WHITE,
-        color: props.brandColor
-      },
-      ':active': {
-        backgroundColor: 'transparent'
-      }
-    }
-  };
+  if (!TabsComponent) throw new Error(`Unknown Tabs type: ${type}`);
```
paniclater: I like how this leaves the future open for additional, unforeseen types.

src/components/tabs/PillTabs.js
```diff
@@ -0,0 +1,73 @@
+const React = require('react');
+
+const ButtonGroup = require('../ButtonGroup');
+const Tabbable = require('./Tabbable');
+
+const StyleConstants = require('../../constants/Style');
+
+const PillTabs = Tabbable(({
```
paniclater: I wonder if this pattern would be more approachable for some team members if the component was assigned here:

const Pill = (props) => {};

and then exported with Tabbable at the bottom.

module.exports = Tabbable(Pill);

src/components/tabs/StandardTabs.js
```diff
@@ -0,0 +1,92 @@
...
+
+        ':hover': {
+          color: this.props.brandColor
+        }
+      }, this.props.activeTabStyles)
+    };
+  }
+}
+
+const Tab = Radium(TabWithoutRadium);
```
paniclater: I'm a little confused what is going on here with the without radium biz.

src/components/Tabs.js
jtanner: I'd rather keep these lower case because it matches how we use the `type` prop in other components, e.g. `Button`.

src/components/Tabs.js
jtanner: I should have mentioned that in the description. Making `Tabs` more "open" was one of my goals with this refactoring :)

src/components/tabs/PillTabs.js
jtanner: I like that idea 👍 I'll change it.

src/components/tabs/StandardTabs.js
jtanner: I think Luke was confused about it too in another PR. I think I will move it to it's own file to remove the confusion with the naming.

It's named differently because it needs to be wrapped by `Radium` before it can be used by `StandardTabs` in this same file.

jtanner: I moved it in 076804b. Is it less confusing this way?

src/components/tabs/PillTabs.js
jtanner: Done in c1226e1.

I had to add `*.propTypes = Tabbable.propTypes` to make eslint happy.

jtanner:
@paniclater thanks for the review, I've addressed your feedback so I think this is ready for another look 👀

src/components/tabs/Tabbable.js
```diff
@@ -0,0 +1,71 @@
...
+      }
+    }
+
+    handleTabSelect (selectedTab) {
+      this.props.onTabSelect(selectedTab);
+
+      this.setState({
+        selectedTab
+      });
+    }
```
lukeschunk: where is this called?

jtanner: Good catch! I was supposed to move https://github.com/mxenabled/mx-react-components/pull/606/files#diff-6d37b6931020e71a211c578f514954b6L218 to `render` in this file.

jtanner: Fixed. Take a look at https://github.com/mxenabled/mx-react-components/pull/606/files#diff-3f7be01abc8541a8f5cfb64de7f4105dR45

lukeschunk:
this looks good to me - still getting used to some of the composition patterns but i like it 👍

lukeschunk:
:shipit:

jtanner:
@paniclater do these latest changes look good to you too?

paniclater:
Sorry I missed this. Yes thank you!

paniclater:
:shipit:

PR #622, opened 2017-06-13
Add theme to ToggleSwitch

jtanner:

Add theme to ToggleSwitch

### DIFF
--- a/docs/components/ToggleSwitchDocs.js
+++ b/docs/components/ToggleSwitchDocs.js
@@ -5,6 +5,14 @@ const { ToggleSwitch } = require('mx-react-components');
 const Markdown = require('components/Markdown');
 
 class ToggleSwitchDocs extends React.Component {
+  state = {
+    checked: false
+  }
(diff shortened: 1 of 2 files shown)


docs/components/ToggleSwitchDocs.js
```diff
@@ -5,6 +5,14 @@ const { ToggleSwitch } = require('mx-react-components');
 const Markdown = require('components/Markdown');
 
 class ToggleSwitchDocs extends React.Component {
+  state = {
+    checked: false
+  }
```
lukeschunk: we probably don't need to set this explicitly to false - but if you like to for discoverability that's 🆒 too

lukeschunk:
:shipit:

docs/components/ToggleSwitchDocs.js
jtanner: Yeah I put it there mostly for documentation/discoverability.

PR #623, opened 2017-06-14
Add theme to Tooltip - there were no docs to update

jtanner:

Add theme to Tooltip - there were no docs to update


lukeschunk:
:shipit:

PR #629, opened 2017-06-21
Update FileUpload to validate extensions and show errors

jjcarstens:
I noticed `FileUpload` docs state you could use extensions in the `allowedFileTypes` option, but it wasn't actually filtering.

So this changes that to filter on extension if the type has not already been filtered

A few other changes:
* Skips validating image fields if the `file_type` or `file_size` is already invalid (no point wasting time)
* By default, adds red error messages for each `failedValidationType` so there is feedback to the user why a file won't upload
* If `onFileValidation` callback is present, skips the default and uses the callback instead to allow custom implementation

screen shot 2017-06-21 at 2 03 13 pm

cc/ @lukeschunk @jmo @paniclater

### DIFF
--- a/docs/components/FileUploadDocs.js
+++ b/docs/components/FileUploadDocs.js
@@ -27,7 +27,16 @@ class FileUploadDocs extends React.Component {
         <div style={{ cursor: 'pointer', textAlign: 'center', width: '80%', margin: 'auto' }}>
           <FileUpload
             allowedFileTypes={['image/jpeg', 'text/csv', 'image/png', 'gif']}
-
+            imageValidation={{
+              exactHeight: 200,
+              exactWidth: 400,
+              maxHeight: 400,
+              maxWidth: 800,
+              minHeight: 100,
+              minWidth: 200,
+              ratioHeight: 1,
+              ratioWidth: 2
+            }}
             maxFileSize={3000}
             onFileAdd={this._handleFileChange}
             onFileRemove={this._handleFileChange}
--- a/docs/components/FileUploadDocs.js
+++ b/docs/components/FileUploadDocs.js
@@ -26,21 +26,11 @@ class FileUploadDocs extends React.Component {
         <h3>Demo</h3>
         <div style={{ cursor: 'pointer', textAlign: 'center', width: '80%', margin: 'auto' }}>
           <FileUpload
-            allowedFileTypes={['image/jpeg', 'text/csv', 'image/png']}
-            imageValidation={{
-              exactHeight: 200,
-              exactWidth: 400,
-              maxHeight: 400,
-              maxWidth: 800,
-              minHeight: 100,
-              minWidth: 200,
-              ratioHeight: 1,
-              ratioWidth: 2
-            }}
+            allowedFileTypes={['image/jpeg', 'text/csv', 'image/png', 'gif']}
+
             maxFileSize={3000}
             onFileAdd={this._handleFileChange}
             onFileRemove={this._handleFileChange}
-            onFileValidation={this._handleFileValidation}
             uploadedFile={this.state.uploadedFile}
           >
             Click Here to Upload File (demo).
@@ -100,7 +90,7 @@ class FileUploadDocs extends React.Component {
     }
 
     <FileUpload
-      allowedFileTypes={['image/jpeg', 'text/csv', 'image/png']}
+      allowedFileTypes={['image/jpeg', 'text/csv', 'image/png', 'gif']}
       imageValidation={{
         exactHeight: 200,
         exactWidth: 400,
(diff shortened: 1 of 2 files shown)


src/components/FileUpload.js
```diff
@@ -140,66 +178,73 @@ class FileUpload extends React.Component {
...
-    if (validationMessages.length) {
+    if (failedValidationTypes.length) {
       this.setState({
-        dragging: false
+        dragging: false,
+        failedValidationTypes
       });
 
-      this.props.onFileRemove(this.props.uploadedFile);
+      return this.props.onFileRemove(this.props.uploadedFile);
```
iheartrachie: I'm pretty sure `return` isn't needed here because these prop methods will be called without it. If this method needs to return a value, it needs to be called separately.

I hope someone else on the team will correct me if I'm wrong :)

src/components/FileUpload.js
```diff
@@ -246,6 +291,11 @@ class FileUpload extends React.Component {
               </div>
             ) : (
               <div style={styles.centered}>
+              {this.state.failedValidationTypes.length > 0 ? (
+                  this.state.failedValidationTypes.map(validationType => {
+                    return (<div style={styles.invalidMessage}>{this._getDefaultValidationMessage(validationType)}</div>);
+                  })
+              ) : null }
```
iheartrachie: It's silly but we don't put a space after `null`. I'd only change it though if you're already making another change.

src/components/FileUpload.js
jmophoto: My thought, too.

docs/components/FileUploadDocs.js
```diff
@@ -26,21 +26,11 @@ class FileUploadDocs extends React.Component {
...
-              exactHeight: 200,
-              exactWidth: 400,
-              maxHeight: 400,
-              maxWidth: 800,
-              minHeight: 100,
-              minWidth: 200,
-              ratioHeight: 1,
-              ratioWidth: 2
-            }}
+            allowedFileTypes={['image/jpeg', 'text/csv', 'image/png', 'gif']}
```
jmophoto: Should this be `image/gif` and is it hard G or soft G.

src/components/FileUpload.js
lukeschunk: sorry i think i led you down the wrong path with my explanation of return 😰

docs/components/FileUploadDocs.js
jjcarstens: Actually, this was my lazy attempt to validate an extension instead of a filetype. So it would check `something.gif` as valid.

src/components/FileUpload.js
jjcarstens: So I need `return` on line 236 so `onFileValidation` takes precedence and doesn't use the default error reporting of invalid types. But I don't need the 2 after that which you've commented on. I just added them because the console was complaining about "inconsistent returns" and I just hate seeing red

src/components/FileUpload.js
```diff
@@ -140,66 +178,73 @@ class FileUpload extends React.Component {
...
-        validationMessages.push('file_size');
-      }
+    if (isTooBig) {
+      failedValidationTypes.push('file_size');
+    }
 
-      if (isInvalidType) {
-        validationMessages.push('file_type');
-      }
+    if (isInvalidFileType && isInvalidFileExtension) {
```
jjcarstens: So I had a reason why `&&`..but I can't remember it and further thinking makes me believe I should just do `||`. Going to change

jjcarstens:
O hey..look at that. Without the `return`s it fails the test.

I'm open for suggestions if there is a better way. Otherwise, going to put them back.

src/components/FileUpload.js
```diff
@@ -140,61 +178,68 @@ class FileUpload extends React.Component {
...
+      failedValidationTypes.push('file_type');
+    }
+
+    if (failedValidationTypes.indexOf('file_type') < 0 && failedValidationTypes.indexOf('file_size') < 0) {
+      failedValidationTypes = failedValidationTypes.concat(this._validateImageDimensions(width, height));
+    }
 
-      this.props.onFileValidation(validationMessages);
+    if (this.props.onFileValidation) {
+      return this.props.onFileValidation(failedValidationTypes);
```
iheartrachie: If you're trying to break out from the method, you'll probably need to stick the rest in an else statement to avoid errors.

iheartrachie: Oh and return will still need to be on its own line or else it'll return the function. Since you're not assigning that to anything it's probably NBD but I'd avoid it.

jjcarstens: So you gave me an idea. I can just avoid the `return` and use a slightly larger `if/else` statement to handle each case. Should be ready for review again

iheartrachie:
:shipit:

PR #630, opened 2017-06-21
fix check for invalid type and extension

jjcarstens:
I remembered why I needed `&&` to check this condition.

We're checking `invalidity` with an `if` statement. So in order for this to evaluate, we would need both to be invalid. 

The way it is now a valid extension still gets flagged because it's an invalid file type. 🤦‍♂️ 

cc @iheartrachie @lukeschunk


lukeschunk:
:shipit:

PR #631, opened 2017-06-28
Corrects filter on TimeXAxis ticks prop for TimeBasedLineChart

cerinman:
Showing more than 10 ticks on the x axis can be problematic as dates
can over lap.  Previously we used modulus/3 for just the case where the
type was day and did no correction for month. This allowed over lap in
the case where the date range for the graph was large enough to have
many ticks on the x axis.

We now just calc a modulus value based upon the data’s length / 10 to
ensure that we show the correct number of ticks in all cases.

### Before
screen shot 2017-06-28 at 1 33 47 pm

### After
screen shot 2017-06-28 at 1 34 17 pm


phantomxc:
:shipit:

PR #632, opened 2017-07-01
Luke/Update Linting Related Dependencies and Fix Errors

lukeschunk:
A lot of our libraries were out of date so this updates them. In the process of the update, there were a lot of linting errors, which i mostly fixed by running `eslint --fix`. I wasn't able to include the auto fix option in the webpack build as we're not bundling the main directory with webpack. 

Updated the same as @jtanner did internally.


docs/components/BarChartDocs.js
```diff
@@ -97,13 +97,13 @@ class BarChartDocs extends React.Component {
...
-      value: 10 //number - required, value of bar, determines bar size
-    }]
-  `}
+          {`
+            [{
+              label: 'label', //optional - Anything that can be parsed by d3 as an axis value.
+              color: '#E3E6E7', //string - hexcode or rgb
+              value: 10 //number - required, value of bar, determines bar size
+            }]
+          `}
```
iheartrachie: With the spacing in the markdown blocks it used to including the spacing in the docs and it looked weird. Can you double check that's not happening here?

docs/components/BarChartDocs.js
```diff
@@ -114,7 +114,7 @@ class BarChartDocs extends React.Component {
         <p>Object containing the desired padding on the SVG to account for axis labels and tooltips.</p>
         <p>Default:</p>
         <Markdown lang='js'>
-  {`
+          {`
```
iheartrachie: Why is the indentation of the opening bracket changed but not the closing bracket?

docs/components/BarChartDocs.js
lukeschunk: 👌

docs/components/BarChartDocs.js
lukeschunk: hmm good question, i'm wondering if the eslint autofix is not working as we need it to. looking

lukeschunk:
We currently do markdown like this: 
```
        <Markdown lang='js'>
  {`
    [{
      label: 'label', //optional - Anything that can be parsed by d3 as an axis value.
      color: '#E3E6E7', //string - hexcode or rgb
      value: 10 //number - required, value of bar, determines bar size
    }]
  `}
        </Markdown>
```

However this makes our linting fail the `react-jsx-indent` rule. I tried indenting properly, but then there was weirdness in the way the `marked` library was parsing the string. I looked into fixing that for a bit but couldn't come up with a good fix. 


So instead i've added `/* eslint-disable react/jsx-indent */` on any file where we use `<Markdown />`

paniclater:
I think disabling the eslint rule there is a good idea, can you also add a comment explaining why it is disabled so future devs will know?

lukeschunk:
@paniclater @iheartrachie ready for another look

iheartrachie:
:shipit:

PR #634, opened 2017-07-07
luke/added default aria-label & focusable to Icon, removed usage in Button

lukeschunk:
We decided that having `aria-hidden: true` and `focusable: false` set as defaults on our `<Icon />` SVGs would be a good way to go. 

I added them in default props rather then right on the `<svg>` tag so we can still override if necessary.


jtanner:
:shipit:

PR #635, opened 2017-07-18
Luke/DateRangePicker_Update

lukeschunk:
There's a new design for the `<DateRangePicker />` and this makes those updates. 

Some Highlights: 
* Adds "DateSelectBox"  - these are now part of the `<SelectionPane />` component 
* Adds an `onClose` prop. This will be useful for taking action (ie fetching data) when the date picker is closed
* Factored half the component (the left half on desktop) to a `<SelectionPane />`
* Updated Default Ranges as per design request 
* Adds `<YearSelector />` and `<MonthSelector />` components to calendar to improve UX (ie: less clicks to go back a year)
* Some minor refactoring to make things more discoverable


src/components/DateRangePicker/SelectionPane.js
```diff
@@ -0,0 +1,128 @@
...
+    isLargeOrMediumWindowSize: PropTypes.bool,
+    primaryColor: PropTypes.string,
+    selectedBox: PropTypes.string,
+    selectedEndDate: PropTypes.number,
+    selectedStartDate: PropTypes.number,
+    setCurrentDate: PropTypes.func
+  };
+
+  state = {
+    selectedBox: 'from'
```
paniclater: What are the other options this can be set to? Should this be enumerated in a constants file or object?

src/components/DateRangePicker/SelectionPane.js
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  _handleDateBoxClick = (date, selectedBox) => {
+    this.props.handleDateBoxClick(date, selectedBox);
```
paniclater: I believe we settled on using the `onDoSomething` style for the prop name and the imperative `doSomething` name for the function passed in for prop handlers. Correct me if I'm wrong though, I'm always mixing those up.

src/components/DateRangePicker/SelectionPane.js
```diff
@@ -0,0 +1,128 @@
...
+
+  render () {
+    const styles = this.styles();
+    const { selectedStartDate, selectedEndDate, isLargeOrMediumWindowSize } = this.props;
+
+    return (
+      <div style={styles.container}>
+        <div style={{ borderBottom: isLargeOrMediumWindowSize ? 'none' : '1px solid ' + StyleConstants.Colors.FOG }}>
+          <label style={styles.boxLabel}>From</label>
+          <div onClick={() => this._handleDateBoxClick(selectedStartDate, 'from')} style={Object.assign({}, styles.dateSelectBox, this.props.selectedBox === 'from' ? styles.selectedDateSelectBox : null)}>{selectedStartDate ? moment.unix(selectedStartDate).format('MMM D, YYYY') : 'Select Start Date'}</div>
```
paniclater: Yeah, feels like 'from' should maybe be a constant here.

src/components/DateRangePicker/SelectionPane.js
```diff
@@ -0,0 +1,128 @@
...
+            Select a Range
+          </div>
+          <DefaultRanges {...this.props} styles={styles} />
+        </div>
+      </div>
+    );
+  }
+
+  styles = () => {
+    const { isLargeOrMediumWindowSize } = this.props;
```
paniclater: I think including the isLargeOrMediumWindowSize here directly instead of passing it in via props will make this component more portable.

src/components/DateRangePicker/SelectionPane.js
```diff
@@ -0,0 +1,128 @@
...
+        color: StyleConstants.Colors.PRIMARY,
+        fontFamily: StyleConstants.Fonts.SEMIBOLD,
+        fontSize: StyleConstants.FontSizes.SMALL,
+        padding: `${StyleConstants.Spacing.LARGE}px 0px ${StyleConstants.Spacing.SMALL}px ${StyleConstants.Spacing.LARGE}px`
+      },
+      rangeOptions: {
+        boxSizing: 'border-box',
+        color: StyleConstants.Colors.CHARCOAL,
+        display: 'flex',
+        flexWrap: 'wrap',
```
paniclater: Does using `wrap` here make more sense than having them laid out in rows? Do they need to dynamically wrap?

src/components/DateRangePicker/Selector.js
```diff
@@ -0,0 +1,116 @@
...
+      },
+      calendarHeaderNav: {
+        width: 35,
+        cursor: 'pointer'
+      }
+    };
+  }
+}
+
+class MonthSelector extends React.Component {
```
paniclater: Is there a reason MonthSelector & YearSelector can't be configured into a single component?

src/components/DateRangePicker/SelectionPane.js
```diff
@@ -0,0 +1,128 @@
...
+        width: 35,
+        cursor: 'pointer'
+      },
+
+      boxLabel: {
+        fontFamily: StyleConstants.FontFamily,
+        fontSize: StyleConstants.FontSizes.MEDIUM,
+        color: StyleConstants.Colors.CHARCOAL
+      },
+      dateSelectBox: {
```
paniclater: what is the difference between `dateSelectBox` and `selectedDateSelectBox`?

lukeschunk:
thanks for the review @paniclater also sorry i thought i edited to add the gifs but here they are: 

### Desktop 
desktop _gif

### Mobile
mobile_ gif

src/components/DateRangePicker/Selector.js
lukeschunk: no reason, but I think that this is clearer & more semantic in the parent component https://github.com/mxenabled/mx-react-components/pull/635/files#diff-b5e02cb776f0bf4505e16b8ba45bcf8dR257

src/components/DateRangePicker/SelectionPane.js
lukeschunk: i think the gif will make this more clear, but the `selectedDateSelectBox` is styling for when the component has focus

src/components/DateRangePicker/SelectionPane.js
lukeschunk: the component that the `rangeOptions` style is applied to maps over the ranges. if i did it in rows wouldn't i need to set a new element for every two elements in the ranges? this is nice because no matter the number of items in the range it works nicely: 

screen shot 2017-07-18 at 7 10 42 pm


if you're thinking of something i'm not though lmk

src/components/DateRangePicker/SelectionPane.js
lukeschunk: what would the constant object structure look like?

lukeschunk: i guess i'll just do a mirrored object with `FROM: 'FROM'`

src/components/DateRangePicker/SelectionPane.js
lukeschunk: done

lukeschunk:
@paniclater this is ready for another look. 

* Changed the 2nd DateRangePicker to be called `DateRangePickerTest` 
* adjusted prop names
* used constant for `TO` and `FROM` 
* fixed CSS so spacing between two panes is equal: 

screen shot 2017-07-21 at 11 31 49 am

paniclater:
:shipit: thanks!

PR #639, opened 2017-07-26
Luke/drawer on open ref

lukeschunk:
This adds two callback props to the Drawer, `onOpen` and `beforeClose` 

These are useful so users of the component can trigger UI changes at specific times in the drawer opening/closing lifecycle.


jtanner:
@lukeschunk what do you think potential use cases will be for `beforeClose`?

lukeschunk:
@jtanner for example if you wanted to close or edit another UI element. you can see the internal PR i put up which uses this.

PR #640, opened 2017-08-01
implemented _isEqual props check in Icon shouldComponentUpdate

lukeschunk:
This implements shouldComponentUpdate in the `<Icon />` component. This should fix performance bugs caused from the Icon re-rendering too many times. 

I tried to use [React.PureComponent](https://facebook.github.io/react/docs/react-api.html#react.purecomponent) but that wasn't working as it does a shallow comparison, and elementProps is an object.


cerinman:
:shipit:

PR #643, opened 2017-08-11
Adds the Updated Date Range Picker as the sole date range picker

lukeschunk:
This removes the two components `DateRangePicker` and `DateRangePickerTest` and instead just creates one using the updated design and code. 

All of the difs in `src/components/DateRangePicker.js` have actually already been reviewed, however because it's in the same file as the old file it looks like a massive diff...


paniclater:
:shipit:

cerinman:
:shipit:

PR #645, opened 2017-08-16
Sb/add moment config

sambev:
It would be nice to be able to set values module wide, like moment's default timezone for all components that use it.

This adds a config property to the module with a function that can do that.

It also adds a script for easier local testing.


sambev:
@jtanner I added docs for Config and the new function after your approval. Would you mind taking a look at them?

PR #646, opened 2017-08-18
Fix for common height issues in Drawer component.

sambev:
- Set a static height (50px) to the header
- Use calc to assign dynamic height for the content
- set overflow on the content.

### DIFF
--- a/src/components/Drawer.js
+++ b/src/components/Drawer.js
@@ -1,4 +1,5 @@
 const _isNumber = require('lodash/isNumber');
+const _get = require('lodash/get');
 const PropTypes = require('prop-types');
 const { StyleRoot } = require('radium');
 const FocusTrap = require('focus-trap-react');
@@ -93,6 +94,21 @@ class Drawer extends React.Component {
 
       return Math.max(newLeft, windowWidth - this.props.maxWidth);
     }
+  };
+
+  /**
+   * Figure out the height of the header. This can come from either:
+   *
+   * this.props.headerStyle.height
+   * this.props.header.height
+   *
+   * If neither are set, default to 50px
+   */
+  _getHeaderHeight = () => {
+    return _get(this.props, ['headerStyle', 'height'],
+                _get(this.props,
+                     ['styles', 'header', 'height'],
+                     '50px'));
   };
 
   /**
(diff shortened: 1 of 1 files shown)


jmophoto:
What about making `header_height` a prop with a default value of 50? Then it could be easily overridden if needed.

sambev:
Good point. That would be more extensible than what I have, but the API feels a bit strange since one can already set the height via the prop `headerStyle`.

Which brings up another problem with my approach, it can get nuked by someone passing in `height` via `headerStyle` and the content wont account for it.

https://github.com/mxenabled/mx-react-components/blob/ba7d39e94ef10b2b14556f7245a8de168338826a/src/components/Drawer.js#L171

 Maybe I could just set a default of 50 to `height` in `headerStyle`? That way we don't have two props that can clash and the body content will pick it up?

src/components/Drawer.js
```diff
@@ -229,7 +231,8 @@ class Drawer extends React.Component {
       },
       content: {
         backgroundColor: StyleConstants.Colors.WHITE,
-        height: '100%'
+        overflow: 'auto',
+        height: `calc(100% - ${HEADER_HEIGHT})`
```
cerinman: I'm not sure how well supported `calc` is for mobile devices.  There is `?` for android, IE, and Opera.

https://developer.mozilla.org/en-US/docs/Web/CSS/calc

cerinman: Also, Firefox and Safari require prefixes it looks like.

sambev: I texted this on safari mobile, so I know it works there. According to https://caniuse.com/#search=calc It seems that this should work on everything mobile outside of opera, which I am okay with 😬 .

cerinman: Cool, the extra testing makes me feel better.  Thanks for digging.

sambev: Tested on 4.3 android, firefox, and safari and they all worked for me.

src/components/Drawer.js
```diff
@@ -96,6 +97,21 @@ class Drawer extends React.Component {
...
 
   /**
+   * Figure out the height of the header. This can come from either:
+   *
+   * this.props.headerStyle.height
+   * this.props.header.height
+   *
+   * If neither are set, default to 50px
+   */
+  _getHeaderHeight = () => {
```
sambev: @jmophoto I moved the height setting to this method. Let me know what you think

cerinman:
:shipit:

PR #647, opened 2017-08-21
Joe/theming updates

jtanner:
After merging the latest master into `joe/theme-master` I noticed there were some things that needed to be updated to use theme.

* Update `DateRangePickerTest` and `SelectionPane` to use the `theme` prop
* Finish updating `Modal` (not sure why it was only half completed)
* Other minor fixes


cerinman:
:shipit:

PR #651, opened 2017-08-23
re-added ccomponentWillReceiveProps

lukeschunk:
Re-adds in `componentWillReceiveProps` as there was a bug where the calendar wasn't updating on default range select. 

I took it out here because i thought it was no longer necessary, but the Month and Year selectors are based on state 🙁  

https://github.com/mxenabled/mx-react-components/pull/643/files#diff-b5e02cb776f0bf4505e16b8ba45bcf8dL182


cerinman:
:shipit:

jmophoto:
:shipit:

PR #652, opened 2017-08-29
Sb/date components lazy getters

sambev:
This changes the API to DateRangePicker and TimeBasedLIneChart.

Now, their default date props are functions rather than values. This allows them to pick up global config settings since defaultProps are set at import time rather than instantiation time.


sambev:
It is also worth noting that getting a moment date is a side effect, which is a bad thing to have with props: https://facebook.github.io/react/docs/components-and-props.html#props-are-read-only
>All React components must act like pure functions with respect to their props.

PR #653, opened 2017-08-29
Update DateRangePicker to use _merge and a nested object for styles prop

bsbeeks:
@lukeschunk here is what we talked about. I started with adding a "position" prop for the optionsWrapper, but ended on just letting the user pass in a styles object. This gives them more flexibility and allows them to set the position (left, center, right) of the optionsWrapper.


lukeschunk:
:shipit:

src/components/DateRangePicker.js
```diff
@@ -28,6 +28,7 @@ class DateRangePicker extends React.Component {
     minimumDate: PropTypes.number,
     onClose: PropTypes.func,
     onDateSelect: PropTypes.func,
+    optionsWrapperStyle: PropTypes.object,
```
jtanner: @bsbeeks we're moving away from using individual props for updating specific styles. Can you please change this to use the `styles` prop pattern? Here's an example https://github.com/mxenabled/mx-react-components/blob/master/src/components/SimpleSelect.js#L156

bsbeeks: Love it, I'll make the change.

bsbeeks:
@jtanner changes made. This will be a breaking change since the styles prop will now be a nested object. I took the opportunity to also rename the prop to `styles` to match the example you provided.

src/components/DateRangePicker.js
```diff
@@ -33,7 +34,7 @@ class DateRangePicker extends React.Component {
     selectedEndDate: PropTypes.number,
     selectedStartDate: PropTypes.number,
     showDefaultRanges: PropTypes.bool,
-    style: PropTypes.object
```
jtanner: @bsbeeks thank you for adding the `styles` prop 👍 
I have one more small request, can you bring back this `style` prop so this PR isn't a breaking change and to match the example that I provided?

https://github.com/mxenabled/mx-react-components/blob/master/src/components/SimpleSelect.js#L109

bsbeeks: Gotcha. It seemed redundant to have both but I understand the desire to avoid a breaking change.

bsbeeks:
@jtanner style prop has been added back. Let me know if there is anything else.

cerinman:
I just caused you some merge conflicts because of our theming changes I think @bsbeeks.  Sorry about that!

bsbeeks:
@cerinman no problem. Merged your changes and found two places I had to make adjustments:

1) Not passing a theme was breaking the component. So I set the default `theme` prop to `/constants/Style`
2) `getDateFormat` was calling `_isLargeOrMediumWindowSize` directly, even though it was being passed it's results. So I switched to used the passed value.

src/components/DateRangePicker.js
```diff
@@ -79,7 +82,8 @@ class DateRangePicker extends React.Component {
     onClose () {},
     onDateSelect () {},
     placeholderText: 'Select A Date Range',
-    showDefaultRanges: false
+    showDefaultRanges: false,
+    theme: defaultTheme
```
jtanner: @bsbeeks can you remove this `defaultTheme` please? It's not necessary because of `StyleUtils.mergeTheme` (see https://github.com/mxenabled/mx-react-components/blob/7f2d4817c8121be95c1ce7540d324c90efd325fb/src/components/DateRangePicker.js#L222). Sorry to ask you to change yet another thing 😬

bsbeeks: I'll give it another go. But when no `theme` was provided in the props I got an error when the component tried to access `theme.BreakPoints`.

bsbeeks:
@jtanner must have been related to the `getDateFormat` issue I resolved. `defaultTheme` removed!

jtanner:
@bsbeeks this was released in v5.0.1

PR #658, opened 2017-09-07
Fix references that should have moved from props to state

jtanner:
`lineColor` and `margin` were moved to state and I missed updating them in a couple places.


paniclater:
:shipit:

lukeschunk:
:shipit:

PR #659, opened 2017-09-19
added state to manage which default range the user clicked on

lukeschunk:
This fixes a bug where on months that had 30 days, two default ranges could appear to be "selected" with the Icon showing up in the UI. 

This adds a `selectedOption` state on the DefaultRanges component and displays the icon based on that `selectedOption` state in addition to the logic that was previously in place. 

No longer a stateless functional component though.


jtanner:
:shipit:

PR #661, opened 2017-09-25
Fix Drawer header size

jtanner:
The `Drawer`'s header height was 65px but the code expected it to be 50px [[see line 132]](https://github.com/mxenabled/mx-react-components/pull/661/files#diff-33a8118349b4401b3292483c9767bcdeR132) which caused issues with the content extending past the bottom of the page.

Before:

screen shot 2017-09-25 at 11 09 53 am

After:

screen shot 2017-09-25 at 11 10 22 am


paniclater:
I truly don't understand why box-sizing isn't the default behavior of block dom elements. 

Thanks for the fix: :shipit:

jtanner:
I agree, it really should be the default.

PR #662, opened 2017-09-25
Luke/accessibility date range picker

lukeschunk:
### First Cut at Accessibility - Keyboard Nav on DateRange Picker 

I was modeling much of the functionality off of this date picker: https://dequeuniversity.com/library/aria/date-pickers/sf-date-picker 

Some Notes: 
* Adds support for keyboard navigation and selection across the date range picker - specifically in the month table 
* used the ref callback and dynamic setting of tabIndex to accomplish desired behavior. In doing this, I had to make monthTable a stateful component 
* Used a large switch/case function to handle different keys and scenarios. [this is the messiest part of this PR](https://github.com/mxenabled/mx-react-components/compare/luke/accessibility-date-range-picker?expand=1#diff-b5e02cb776f0bf4505e16b8ba45bcf8dR166) any suggestions welcome 

To Do: 
* Add some sort of FocusTrap to open Date Range Picker 
* all aria attributes/screen reader support fixes 
* Fix undesired behavior where date autofocuses even after selection pane or default range selection 

### Gif (all keyboard nav)
drp_gif


src/components/DateRangePicker.js
```diff
@@ -152,12 +156,49 @@ class DateRangePicker extends React.Component {
...
+
+      this.setState({ focusedDay: moment.unix(this.state.focusedDay).subtract(7, 'days').startOf('day').unix() });
+    } else if (keycode(e) === 'down') {
+      e.preventDefault(); //stop browserscrolling
+
+      if (moment.unix(this.state.focusedDay).add(7, 'days').startOf('day').isSameOrAfter(endDate)) {
+        this.setState({ currentDate: moment.unix(this.state.focusedDay).add(7, 'days').startOf('day').unix() });
+      }
+
+      this.setState({ focusedDay: moment.unix(this.state.focusedDay).add(7, 'days').startOf('day').unix() });
```
jtanner: Is it necessary to call `.unix()` at the end?

src/components/DateRangePicker.js
```diff
@@ -152,12 +156,49 @@ class DateRangePicker extends React.Component {
...
+      this.setState({ focusedDay: moment.unix(this.state.focusedDay).subtract(1, 'days').startOf('day').unix() });
+    } else if (keycode(e) === 'enter') {
+      this._handleDateSelect(true, this.state.focusedDay);
+    } else if (keycode(e) === 'up') {
+      e.preventDefault(); //stop browserscrolling
+      if (moment.unix(this.state.focusedDay).subtract(7, 'days').startOf('day').isBefore(startDate)) {
+        this.setState({ currentDate: moment.unix(this.state.focusedDay).subtract(7, 'days').startOf('day').unix() });
+      }
+
+      this.setState({ focusedDay: moment.unix(this.state.focusedDay).subtract(7, 'days').startOf('day').unix() });
```
jtanner: Will you clean up the duplication in these `if` blocks so they're a little easier to read?
e.g.
```js
      const day = moment.unix(this.state.focusedDay).subtract(7, 'days').startOf('day');

      if (day.isBefore(startDate)) {
        this.setState({ currentDate: day.unix() });
      }

      this.setState({ focusedDay: day.unix() });
```

src/components/DateRangePicker/MonthTable.js
```diff
@@ -2,61 +2,76 @@ const React = require('react');
...
+      const savedStartDate = startDate.date();
+      const day = (
+        <div
+          key={startDate}
+          onClick={!disabledDay && handleDateSelect.bind(null, startDate.unix())}
+          onKeyDown={handleKeyDown}
+          onMouseEnter={!disabledDay && handleDateHover.bind(null, startDate.unix())}
+          ref={ref => {
+            if (moment.unix(focusedDay).date() === savedStartDate) {
+              if (ref) ref.focus();
```
jtanner: Let's combine these two `if` statements.
```js
if (ref && moment.unix(focusedDay).date() === savedStartDate) {
```

src/components/DateRangePicker.js
lukeschunk: yeah definitely. i also thought i made this a case/switch but i must have messed something up with my commits but let me put that back in as well

lukeschunk: ok i stuck with the if/else so i could block scope the `consts` which i think makes it cleaner and better than the switch/case anyway

lukeschunk:
@jtanner thanks for the review - ready for another look

src/components/DateRangePicker.js
lukeschunk: yes because i want to pass around unix time stamps instead of moment objects - i think there's less room for error/mutation this way but lmk if you have other thoughts

jtanner: That makes sense 👍

PR #663, opened 2017-09-28
adding type to HTML input tag in DisplayInput

redperk:
This MR allows us to define a type prop for the input tag in DisplayInput.


redperk:
Turns out we can used the elementProps to pass the type in.

PR #666, opened 2017-10-10
Remove DonutChart label click handler

jtanner:
The click handler of the label in the center of the `DonutChart` was setup incorrectly. It didn't pass along the required index to the handler function which caused the `onClick` prop to receive `undefined` for the `event`. This made users of the `DonutChart` have to work around the problem.

The behavior is also undocumented because the docs state that the `onClick` prop is for when a "pie slice is clicked" mentioning nothing about the label. This removes the click handler from the label.


lukeschunk:
:shipit: 👍

PR #669, opened 2017-10-13
Tabs > handling single tab use case

redperk:
This PR adds a check to handle situations where only 1 tab is passed in. If only 1 tab is passed in, it will display that tab at the top and disable the dropdown feature.

### DIFF
--- a/src/components/Tabs.js
+++ b/src/components/Tabs.js
@@ -174,6 +174,7 @@ class Tabs extends React.Component {
         alignItems: 'center',
         boxSizing: 'border-box',
         color: this.props.brandColor,
+        cursor: 'pointer',
         lineHeight: '20px',
         fontSize: StyleConstants.FontSizes.MEDIUM,
         fontStyle: StyleConstants.Fonts.SEMIBOLD,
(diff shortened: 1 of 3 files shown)


src/components/Tabs.js
```diff
@@ -26,9 +26,18 @@ class Tabs extends React.Component {
 
   state = {
     selectedTab: this.props.selectedTab || 0,
-    showMenu: false
+    showMenu: false,
+    singleTab: false
   };
 
+  componentDidMount () {
+    if (this.props.tabs.length === 1) {
```
paniclater: What made you want to use state here instead of relying on the prop.tabs.length check on its own?

redperk: That's a good point. I thought I was going to need it in more places, but I'm only using it once. I'll make the change.

paniclater:
This looks good, just has the conflicts to be taken care of. I do want to note that it looks like to me if the state had been used as is there would have been bugs because the state wouldn't have been updated on componentDidUpdate, only on componentDidMount. Issues like this lead me to recommend that wherever possible we avoid assigning props to state and managing that reconciliation process. 

Thanks for the updates

redperk:
Closing this since Joe already had a fix for this that he had forgotten to merge in.

PR #670, opened 2017-10-13
Properly set default element props on Icon.

sambev:
Previously, we set aria-hidden and focusable, but if the consumer
passed in any of their own props, those would be removed entirely.

This merges them back in with element props, ensuring the passed in
props 'win' over.


cerinman:
:shipit:

PR #671, opened 2017-10-16
removed reference to React.PropTypes

lukeschunk:
Was doing some testing when I noticed we were still referencing `React.PropTypes`. This was breaking for me when testing on React 16 Internally.

Did a project wide search for other references like this and found none.


cerinman:
Closing this.  It's way old and not needed any more.

PR #673, opened 2017-10-17
Fixing ToggleSwitch in ie11

redperk:
For whatever reason, ie11 is not obeying the the width and height properties. I've added minWidth and minHeight to enforce correct CSS.


jtanner:
@redperk you shouldn't update the changelog or package.json as part of your PR. Someone else could create another release before you are able to release this which would cause conflicts. You can update the changelog etc. as part of the release process.

Will you please remove those release related changes?

PR #674, opened 2017-10-18
Updating columns logic with showHint and hasLabel

redperk:
I was running into a strange issue where I was using DisplayInput but I wasn't using a label and the input box column wasn't using up the whole row.


lukeschunk:
makes sense to me :shipit:

PR #675, opened 2017-10-22
Fix error in IE 11 for Modal component

cerinman:
Pulls in update to `focus-trap-react` library to protect against calling `focus()` on an empty object because IE11 and iFrames suck.

https://github.com/davidtheclark/focus-trap-react/pull/19


redperk:
:shipit:

PR #676, opened 2017-10-23
Correct focus trap update

cerinman:
My last release that updated `focus-trap-react` didn't seem to work.  The version in the node modules directory what still on 3.0.3 even though the yarn.lock said 3.0.5.  Blowing out and yarn installing also didn't fix the issue.  This PR updates the version in the package.json file to 3.0.5 in an attempt to fix.


paniclater:
:shipit:

PR #681, opened 2017-11-14
Button Component - Add elementProps prop

cerinman:
We have a need to be able to pass various aria attributes to the native button element in the Button component for accessibility purposes so this adds the elementProps prop to allow for that.  

Also updated the docs and change log respectively.


jmophoto:
Would this enable us to put `onClick` in the element props instead directly? And if so... should we put a deprecation warning in?

cerinman:
It would allow for that but I've spread the elementProps last on the button so they would win in the case of passing an onClick prop and an onClick in elementProps so there wouldn't be any conflicts.  

I like the idea of leaving the onClick prop as it makes for quick use of the button component in most cases.  At least for now I feel like it is ok to have both.  I might come back through and remove the `aria-label` prop though as it no longer needs to be a top level prop. I just didn't want to have to update that everywhere internally yet.

jmophoto:
:shipit:

PR #682, opened 2017-11-28
add elementProps to buttonGroup

lukeschunk:
I ran into an issue where i needed to be able to pass element props in the ButtonGroup. 

This adds support for that prop.


lukeschunk:
because of wengert's changes here https://github.com/mxenabled/mx-react-components/pull/683 we do not need this change. closing.

PR #683, opened 2017-11-29
Wengert/button

phantomxc:
We had a mix of native props and custom props being passed to our Button element. We also had the concept of elementProps which would just expand onto the native button. This removes that confusion so we don't have to use elementProps anymore while still maintaining backwards compatibility.  

@jtanner @lukeschunk

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -58,32 +58,34 @@ class Button extends React.Component {
   };
 
   render () {
-    const theme = StyleUtils.mergeTheme(this.props.theme, this.props.primaryColor);
-    const styles = this.styles(theme);
+    // Manually consume everything that isn't going to be passed down to the button so we don't have to keep adding props one at a time.
+    // Keep elementProps for backwards compatibility.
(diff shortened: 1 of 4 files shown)


src/components/Button.js
```diff
@@ -58,32 +58,35 @@ class Button extends React.Component {
...
+    // Manually consume everything that isn't going to be passed down to the button so we don't have to keep adding props one at a time.
+    // Keep elementProps for backwards compatibility.
+    const { actionText, buttonRef, children, elementProps, icon, isActive, primaryColor, style, theme, ...rest } = this.props
+    const mergedTheme = StyleUtils.mergeTheme(theme, primaryColor);
+    const styles = this.styles(mergedTheme);
 
     return (
       <button
-        aria-label={this.props['aria-label']}
-        onClick={this.props.type === 'disabled' ? null : this.props.onClick}
```
jtanner: Does removing this mean that setting the `disabled` attribute to `true` will prevent the `onClick` handler from firing?

phantomxc: I answered this in a slack message but the answer is yes.

lukeschunk:
:shipit:

PR #685, opened 2017-12-05
Luke/Adds Prettier to OS

lukeschunk:
Adds Prettier to the OS, and made updates to files in `/src/` and `/docs/` 

Adds these deps: `"eslint-config-prettier"`, `"prettier"`, `"husky"`, and `"lint-staged"` 

I configured it the same as how we did it in raja - the `lint-staged` script in the package.json was copied exactly - if we want to configure the precommit hook differently we could change this. 

Main files to review: 
`.eslintrc` 
`.prettierrc`
`package.json`


lukeschunk:
removing Needs Review label as there were a couple things i didn't quite have right with the config...will revisit

cerinman:
This is olde enough now that I'm just going to close.  We can open a new PR when we're ready to try this again.

PR #686, opened 2017-12-05
Luke/modal accessibility fix

lukeschunk:
Once upon a time i was naiive and I thought I could focus and use aria attributes on SVGs...

However in my old age I've realized that wrapping is the way to go - here I wrapped the modal's close Icon with a button for accessibility. 

modalaccessibility


cerinman:
:shipit:

PR #688, opened 2017-12-11
SimpleSelect accessibility updates

cerinman:
Makes it easy to change accessibility on Options in the SimpleSelect by passing additional properties on the item objects in the items array prop that then over write existing properties.  

I did this because I have a use case internally where I need on option to be treated as a toggle button but the rest to remain as standard options.  I achieved this by destructering the expected properties from the item object and then spreading the rest across the Option component.  This should make it easy to over write when needed without having to add a new property any time we need to change something.

I also added the `isSelected` prop to the usage of the `Option` component in the SimpleSelect.  The API for the `Option` component already supported it, we just weren't using it as expected so I fixed it.

### DIFF
--- a/src/components/SimpleSelect.js
+++ b/src/components/SimpleSelect.js
@@ -78,6 +78,7 @@ class SimpleSelect extends React.Component {
 
               return (
                 <Option
+                  isSelected={isSelected}
                   key={i}
                   label={text}
                   onClick={e => {
--- a/src/components/SimpleSelect.js
+++ b/src/components/SimpleSelect.js
@@ -74,17 +74,28 @@ class SimpleSelect extends React.Component {
           {this.props.children ?
             this.props.children :
             (this.props.items.map((item, i) => {
+              const { icon, isSelected, onClick, text, ...rest } = item;
+
               return (
                 <Option
                   key={i}
-                  label={item.text}
-                  onClick={this._handleItemClick.bind(null, item)}
+                  label={text}
+                  onClick={e => {
+                    if (this.props.scrimClickOnSelect) {
+                      this.props.onScrimClick(e);
+                    }
+
+                    if (onClick && typeof onClick === 'function') {
+                      onClick(e, item);
+                    }
+                  }}
                   style={styles.item}
+                  {...rest}
                 >
-                  {item.icon ? (
-                    <Icon size={this.props.iconSize || 20} style={styles.icon} type={item.icon} />
+                  {icon ? (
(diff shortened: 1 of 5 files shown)


src/components/SimpleSelect.js
```diff
@@ -82,17 +74,29 @@ class SimpleSelect extends React.Component {
           {this.props.children ?
             this.props.children :
             (this.props.items.map((item, i) => {
+              const { icon, isSelected, onClick, text, ...rest } = item;
+
               return (
                 <Option
+                  isSelected={isSelected}
```
phantomxc: Is this needed if you're just passing it straight down? Seems like it could pass along with `...rest`.

src/components/SimpleSelect.js
```diff
@@ -82,17 +74,29 @@ class SimpleSelect extends React.Component {
...
                   key={i}
-                  label={item.text}
-                  onClick={this._handleItemClick.bind(null, item)}
+                  label={text}
+                  onClick={e => {
+                    if (this.props.scrimClickOnSelect) {
+                      this.props.onScrimClick(e);
+                    }
+
+                    if (onClick && typeof onClick === 'function') {
```
phantomxc: We could set a defaultProp instead of the if?

src/components/SimpleSelect.js
cerinman: Technically it doesn't have to be but I left it this way because it seemed more like a public api prop where as the over writing I need to do the rest I was going to leave undocumented.

src/components/SimpleSelect.js
cerinman: good call.  I'll change it.

lukeschunk:
is there any worry for someone who isn't familiar with the API and then accidentally passes down a prop. for example if someone passed an `onClick` prop, it could cause weird behavior, as there is no `onClick` at the parent level, and then it will override the default `onClick` of the Option

src/components/SimpleSelect.js
paniclater: yeah good call

cerinman: oh, no I can't because the onClick is a property in the item object in the items array prop

cerinman:
@lukeschunk no, there shouldn't be any worry because I only do rest on the properties in the item objects of the items array prop.  onClick and other props for the SimpleSelect do not come into play.

phantomxc:
:shipit:

PR #690, opened 2018-01-04
Allow click events on elements outside of FocusTrap components

cerinman:
The `FocusTrap` component that we use on the `Drawer` and `Modal` components was preventing clicks from happening on elements outside the `FocusTrap`.  This was causing issues for things such as are notifications toast that are rendered above drawer and modal components.  

This fixes the issue by passing the correct config option `clickOutsideDeactivates`  to the FocusTrap component for the `Drawer` and `Modal` components.

See documentation for FocusTrap Libraries for details.

https://github.com/davidtheclark/focus-trap-react
https://github.com/davidtheclark/focus-trap

Specific line in focus-trap library that was causing the issue.
https://github.com/davidtheclark/focus-trap/blob/master/index.js#L177

### DIFF
--- a/docs/components/Changelog.js
+++ b/docs/components/Changelog.js
@@ -8,6 +8,11 @@ class Changelog extends React.Component {
 
         <h2>MX React Components V 5.0</h2>
 
+        <h3>5.1.11</h3>
+        <ul>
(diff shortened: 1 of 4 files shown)


sambev:
:shipit:

lukeschunk:
🏀

PR #691, opened 2018-01-09
Calendar accessible

lukeschunk:
There was a UI internally where we use the `<Calendar />` component - which was not accessible. 

This makes the calendar accessible by making it navigable by keyboard. 

calendar_accessible


jtanner:
Any chance we can get some tests around this new logic?

lukeschunk:
@jtanner yeah I think that would be good, let me see what I can do.

src/enhancers/calendar-enhancers.js
```diff
@@ -0,0 +1,17 @@
+const moment = require('moment');
+
+export const __calculateDayByKey = (code, focusedDay) => {
```
paniclater: I think it is a good idea to pull this out to make things testable. Would it make more sense to pass in focusedDay as a moment object so it can be pure, or will that make the component harder to test?

paniclater: also, what's the thinking behind the double underscore function name prefix?

src/components/Calendar.js
```diff
@@ -161,16 +233,19 @@ class Calendar extends React.Component {
...
     return {
-      component: Object.assign({
-        backgroundColor: theme.Colors.WHITE,
-        border: '1px solid ' + theme.Colors.GRAY_300,
-        borderRadius: 3,
-        boxSizing: 'border-box',
-        marginTop: 10,
-        padding: 20
-      }, this.props.style),
+      component: Object.assign(
```
paniclater: so long as this is getting updated stylistically, perhaps add object spread?

src/components/Calendar.js
```diff
@@ -47,24 +56,62 @@ class Calendar extends React.Component {
...
+
+    if (keycode(e) === 'enter') {
+      this._handleDateSelect(this.state.focusedDay, e);
+    }
+
+    if (['left', 'right', 'up', 'down'].includes(keycode(e))) {
+      //prevent browser scrolling with arrow key press
+      e.preventDefault();
+
+      const day = __calculateDayByKey(keycode(e), this.state.focusedDay);
```
paniclater: yeah, seems like passing in the moment obj may be a win since you've already got to take into consideration moment here. Better to isolate the places that need moment.

src/components/Calendar.js
```diff
@@ -2,11 +2,14 @@ const React = require('react');
 const PropTypes = require('prop-types');
 const Radium = require('radium');
 const moment = require('moment');
+const keycode = require('keycode');
 
 const Icon = require('./Icon');
 
 const { themeShape } = require('../constants/App');
 
+import { __calculateDayByKey } from '../enhancers/calendar-enhancers';
```
paniclater: since we call these `utils` in our main codebase, maybe stick to that naming convention in the open source?

lukeschunk: sounds good - i thought of this idea of trying to keep our actual components very minimal (rendering JSX only, if possible) and moving a lot of these functions into utils. 

i guess what I'm saying is I think we under-use utils for this sort of thing. writing code like this will help keep our functions more pure and easier to test

src/components/Calendar.js
lukeschunk: so have it take a third argument, `moment`?

src/enhancers/calendar-enhancers.js
lukeschunk: it doesn't have to be double underscore, but I think something to show that it's coming from a utils file? this way when you see it in the code base you know it's not a class method (you also will probably know because there will be no `this.` when invoking it... 

could be overkill

lukeschunk:
alright @paniclater ready for another look

src/components/Calendar.js
```diff
@@ -87,29 +121,51 @@ class Calendar extends React.Component {
...
 
             return (
               <div
+                className='calendar-day'
+                id={day.isSame(moment.unix(this.state.focusedDay), 'day') ? 'focused-day' : null}
                 key={day}
-                onClick={disabledDay ? null : this._handleDateSelect.bind(null, day.unix())}
-                style={Object.assign({},
+                onClick={() => {
+                  if (!disabledDay) this.props.onDateSelect(day.unix());
```
paniclater: Can disabledDay be removed as an outside dependency for this function?

src/components/__tests__/Calendar-test.js
```diff
@@ -0,0 +1,100 @@
+import React from 'react'; // eslint-disable-line no-unused-vars
+import { shallow } from 'enzyme';
+import moment from 'moment';
+
+import { getNewDateStateChange } from '../../utils/Calendar';
+
+import Calendar from '../Calendar';
```
paniclater: great job testing!

src/utils/Calendar.js
```diff
@@ -0,0 +1,23 @@
+export const getNewDateStateChange = ({ code, focusedDay, startDate, endDate }) => {
```
paniclater: now that this is more specific to state changes within the component, I think it would be valuable to keep it in the same file and export it from there for testing.

src/components/Calendar.js
```diff
@@ -87,29 +121,51 @@ class Calendar extends React.Component {
...
+  _renderMonthTable = styles => {
     const weeks = this._getWeeks();
 
     return weeks.map(week => {
       return (
         <div key={week} style={styles.calendarWeek}>
           {week.map(day => {
-            const isCurrentMonth = day.isSame(moment.unix(this.state.currentDate), 'month');
-            const isSelectedDay = day.isSame(moment.unix(this.props.selectedDate), 'day');
+            const isCurrentMonth = day.isSame(
```
paniclater: is it possible to remove some of these assignments?

src/components/Calendar.js
```diff
@@ -87,29 +121,51 @@ class Calendar extends React.Component {
     return weeks;
   };
 
-  _renderMonthTable = (styles) => {
+  _renderMonthTable = styles => {
```
paniclater: A nice follow up here next would be to refactor these into components.

src/utils/Calendar.js
paniclater: I think this would also be a good place for jsdoc comments to indicate it returns an object that contains one of or both of currentDate and focusedDate as unix moment values or null.

src/components/Calendar.js
```diff
@@ -27,44 +30,75 @@ class Calendar extends React.Component {
...
 
     this.setState({
       currentDate
     });
   };
 
   _getWeeks = () => {
-    const startDate = moment.unix(this.state.currentDate).startOf('month').startOf('week');
-    const endDate = moment.unix(this.state.currentDate).endOf('month').endOf('week');
+    const startDate = moment
```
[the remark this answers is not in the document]
paniclater: Check that, I guess this one is calculating week data. So definitely would be better to pass it in only what it needs and have it return a value, but wouldn't make a good candidate for component conversion.

src/components/Calendar.js
lukeschunk: I totally can - and I started to do this, but I actually think that the variable names make it clearer for the developer. For me when I read `isCurrentMonth` it's immediately obvious to me what the variable represents whereas parsing through
```
            day.isSame(
              moment.unix(this.state.currentDate),
              'month'
            );
```
is a little bit more difficult? what do you think

src/utils/Calendar.js
lukeschunk: okay so check out what i pushed up - now instead of exporting the calendar like this: 
`module.exports = Radium(Calendar);` I've exported it like this: `export default Radium(Calendar)` and then I also export the func at the top for testing. Sam helped me make this work by importing using ES6 instead of commonjs

src/components/Calendar.js
lukeschunk: 👍

src/components/Calendar.js
lukeschunk: so something like this, perhaps: 
```
                onClick={
                  disabledDay ?
                    () => {} :
                    () => this.props.onDateSelect(day.unix())
                }
```

paniclater: It's probably ok as is, just trying to find ways t make these functions more self contained.

src/components/__tests__/Calendar-test.js
lukeschunk: thanks!

lukeschunk:
@paniclater i believe i addressed all feedback

paniclater:
:shipit:

PR #693, opened 2018-01-10
Add the side widths back to the Drawer.

sambev:
These are required in the scenario in which we have no 'right' item
in the header.

Before:
screen shot 2018-01-10 at 10 13 02 am

After:
screen shot 2018-01-10 at 10 12 24 am


cerinman:
:shipit:

PR #695, opened 2018-01-24
.default for Calendar import

lukeschunk:
Since we're now exporting two things from the `Calendar.js` file, the component, and a function for testing, we have to add this `.default`. Unlike ES6 `import` commonJS won't grab the default automatically, so I have to specify which key to grab.


sambev:
:shipit:

jtanner:
:shipit:

PR #696, opened 2018-01-25
Calendar - day changed to a tag

lukeschunk:
I"m not sure why - but when I went to use the `Calendar` internally, the divs were not getting a focused state. I feel like I ran into a similar issue with Icon buttons, and had to wrap them in `a` tags instead.

### DIFF
--- a/src/components/Calendar.js
+++ b/src/components/Calendar.js
@@ -179,7 +179,7 @@ class Calendar extends React.Component {
             const savedStartDate = day.date();
 
             return (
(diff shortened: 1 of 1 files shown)


phantomxc:
No style changes needed?

lukeschunk:
@phantomxc seems to be fine: 

screen shot 2018-01-25 at 3 12 14 pm

phantomxc:
:shipit:

PR #697, opened 2018-01-25
Calendar - only prevent default if its up or down arrow

lukeschunk:
@paniclater originally we talked about always preventing default, however this was causing issues - for example when pressing the tab key, i _do_ want the default browser behavior. Other keys as well we probably want to allow defaults. 

Really the only place I saw the default causing the issue was the up/down arrow with browser scrolling. So I only prevent default if it's an up/down key. 

I thought about trying to do this check in the `getNewDateStateChange` function, but then I'd have to pass in the event, and that function which had become fairly pure would be less so... 

#### edit 

Also adds tabIndex/keyboard nav support to the next and previous month icon buttons.

### DIFF
--- a/src/components/Calendar.js
+++ b/src/components/Calendar.js
@@ -93,7 +93,7 @@ class Calendar extends React.Component {
   };
 
   _handleDayKeyDown = e => {
-    e.preventDefault();
+    if (keycode(e) === 'up' || keycode(e) === 'down') e.preventDefault();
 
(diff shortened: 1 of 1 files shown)


paniclater:
my bad! :shipit:

lukeschunk:
@paniclater i actually added a couple more things to this MR, updated the description to reflect the changes as well

jmophoto:
:shipit:

PR #698, opened 2018-01-30
Calendar - pass e, prop expects an e in DateTimePicker

lukeschunk:
I was using the `<DateTimePicker />` and I noticed it was throwing a `cannot read stopPropagation of undefined`. This is because the prop in the DateTimePicker [expects an event](https://github.com/mxenabled/mx-react-components/blob/master/src/components/DateTimePicker.js#L87), and we weren't passing it in the Calendar's onClick.


jmophoto:
:shipit:

Issue #616, opened 2017-06-12
SimpleInput is missing docs for most of its props


jtanner:
Let's update the docs for `SimpleInput` because it only has info for 2 of 12 props.

markstewy:
@jtanner I think a few other component docs could use some updating as well. Would that be helpful? or is there something else you would rather have me work on?

cerinman:
This has been resolved.  Closing.

Issue #582, opened 2017-05-25
Keep documentation example code in sync with the live demo


jtanner:
I had a chat with @jmophoto recently about how our documentation example code has a tendency to get out of sync with the code used for the demo. Using the same code for both the demo and the example I think would help. Here's an idea for how we could accomplish sharing the code:

1. Move the demo code in the docs files to a separate file. For example, documentation markup for `Tabs` would be in `TabsDocs.js` and it would require the demo component `TabsDemo.js`.
2. Add [webpack raw-loader](https://github.com/webpack-contrib/raw-loader) so we can use to load a file as a string.
3. Require the demo code as a string using `'raw-loader!./TabsDemo.js'` in `TabsDocs.js`
4. Add a `Show source` button below the demo to show the code (instead of showing it at the bottom of the file)

#### Picture example:
![screen shot 2017-05-25 at 10 23 41 am](https://cloud.githubusercontent.com/assets/4348/26460439/da5281a0-4136-11e7-9abc-cadbe8e33f06.png)

#### Then when clicked:
![screen shot 2017-05-25 at 10 23 41 am](https://cloud.githubusercontent.com/assets/4348/26460439/da5281a0-4136-11e7-9abc-cadbe8e33f06.png)
![screen shot 2017-05-25 at 10 24 17 am](https://cloud.githubusercontent.com/assets/4348/26460445/dd11829c-4136-11e7-830f-13346c670dbe.png)

paniclater:
I think this would be super cool.

PR #702, opened 2018-02-03
Simple input docs

markstewy:
issue #616 (SimpleInput is missing docs for most of its props) 

I updated/added prop descriptions in the docs and added them to the example snippet for syntax clarity. I removed placeholder prop and excluded style prop from the docs (placeholder was nonexistent and style was redundant to styles - left in component for backwards compatibility).

I was thinking about overall usability of the docs. Do we want to show multiple variations in the top Demo section so people can see at a glance what customization is possible? Or do we want to keep it clean with a single basic demo?  

Maybe we keep the top demo demo section simple with one basic/default component displayed but expand bottom Example section to show multiple variations displayed along with their code snippets? Thoughts?

### DIFF
--- a/docs/components/SimpleInputDocs.js
+++ b/docs/components/SimpleInputDocs.js
@@ -22,17 +22,48 @@ class SimpleInputDocs extends React.Component {
           elementProps={{
             placeholder: 'Type something'
           }}
+          handleResetClick={() => {
+            /* eslint-disable no-console */
+            console.log('clicked me');
+            /* eslint-enable no-console */
+          }}
+          // icon={'bike'}
+          // rightIcon={'close-solid'}
         />
 
         <h3>Usage</h3>
+        <h5>baseColor <label>String</label></h5>
+        <p>The color of the <Code>input</Code> border on focus.</p>
+
         <h5>elementProps <label>Object</label></h5>
         <p>Properties to pass directly to the <Code>input</Code> element.</p>
+
+        <h5>focusOnLoad <label>Boolean</label></h5>
+        <p>Focus <Code>input</Code> on load, default of false.</p>
+
+        <h5>handleResetClick <label>Function</label></h5>
+        <p>The function that will execute when <Code>rightIcon</Code> is clicked. This prop will not work properly unless <Code>rightIcon</Code> is also declared. </p>
+
+        <h5>icon <label>String</label></h5>
+        <p>The name of icon to display in the left side of the <Code>input</Code>.</p>
 
         <h5>placeholder <label>String</label></h5>
         <p>The text to show before the user starts typing or when the <Code>input</Code> is empty.</p>
 
+        <h5>rightIcon <label>String</label></h5>
+        <p>The name of icon to display in the right side of the <Code>input</Code>. This icon is clickable and will execute the function specified in the <Code>handleResetClick</Code> prop.  This prop will not work properly unless <Code>handleResetClick</Code> is also declared.</p>
+
+        <h5>style <label>Object or Array</label></h5>
+        <p>blah blah</p>
+
+        <h5>styles <label>Object</label></h5>
+        <p>blah blah</p>
+
         <h5>theme <label>Object</label></h5>
         <p>Customize the component&apos;s look. See <Link to='/components/theme'>Theme</Link> for more information.</p>
+
+        <h5>type <label>String</label></h5>
+        <p>blah blah</p>
 
         <h5>valid <label>Boolean</label></h5>
         <p>Indicates whether the value of Input field is valid. If it is not valid, the input field will have a red border.</p>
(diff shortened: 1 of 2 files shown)


markstewy:
just saw #582 maybe i'll take that one

docs/components/SimpleInputDocs.js
```diff
@@ -25,14 +25,40 @@ class SimpleInputDocs extends React.Component {
         />
 
         <h3>Usage</h3>
+        <h5>baseColor <label>String</label></h5>
+        <p>The color of the <Code>input</Code> border on focus.</p>
```
lukeschunk: I would make a note that this prop is deprecated and to use `theme` instead. here's the warning we throw when someone uses that prop":https://github.com/mxenabled/mx-react-components/blob/master/src/utils/Deprecation.js

docs/components/SimpleInputDocs.js
```diff
@@ -25,14 +25,40 @@ class SimpleInputDocs extends React.Component {
...
+
         <h5>elementProps <label>Object</label></h5>
-        <p>Properties to pass directly to the <Code>input</Code> element.</p>
+        <p>Attributes to pass directly to the <Code>input</Code> element. ie. placeholder, value, onchange, etc.</p>
```
lukeschunk: I think I'd prefer the term props, or properties here rather than attributes - since we're still in React, these are technically props

docs/components/SimpleInputDocs.js
```diff
@@ -25,14 +25,40 @@ class SimpleInputDocs extends React.Component {
...
+        <h5>focusOnLoad <label>Boolean</label></h5>
+        <p>Focus <Code>input</Code> on load, default of false.</p>
+
+        <h5>handleResetClick <label>Function</label></h5>
+        <p>The function that will execute when <Code>rightIcon</Code> is clicked. This prop will not work properly unless <Code>rightIcon</Code> is also declared. </p>
 
-        <h5>placeholder <label>String</label></h5>
-        <p>The text to show before the user starts typing or when the <Code>input</Code> is empty.</p>
+        <h5>icon <label>String</label></h5>
+        <p>The name of icon to display in the left side of the <Code>input</Code>.</p>
```
lukeschunk: I would say the "type" of icon instead of "name"

docs/components/SimpleInputDocs.js
```diff
@@ -25,14 +25,40 @@ class SimpleInputDocs extends React.Component {
...
+
+        <h5>rightIcon <label>String</label></h5>
+        <p>The name of icon to display in the right side of the <Code>input</Code>. This icon is clickable and will execute the function specified in the <Code>handleResetClick</Code> prop.  This prop will not work properly unless <Code>handleResetClick</Code> is also declared.</p>
```
lukeschunk: also use "type of icon"

docs/components/SimpleInputDocs.js
```diff
@@ -25,14 +25,40 @@ class SimpleInputDocs extends React.Component {
...
+        <h5>styles <label>Object</label></h5>
+        <p>Allows style additions or overrides to specific component elements including:</p>
+          <ul >
+          <li><Code>wrapper</Code>: the outermost element of the component.</li>
+          <li><Code>activeWrapper</Code>: the outermost element of the component when active.</li>
+          <li><Code>input</Code>: the component's <Code>input</Code> element.</li>
+          <li><Code>icon</Code>: the icon located in the left side of the <Code>input</Code> element.</li>
+          <li><Code>rightIcon</Code>: the icon located in the right side of the <Code>input</Code> element.</li>
+          </ul>
+          <p>*<Code>styles</Code> will override defaults and conflicting <Code>themes</Code></p>
```
lukeschunk: i have mixed feelings about being so specific about the different styles objects you can override. On one hand I feel like if someone wants to get that specific they could look at the source, but I do like having the components be completely useable and everything explained from the docs. 

Maybe there's another clear & concise way to get the info across, but i think your change is a good first cut.

markstewy:
simpleinputscreenshot

docs/components/SimpleInputDocs.js
markstewy: should only list the key values (wrapper, input, etc) but not the descriptions of the elements they represent? Let me know and I can make the edits and push today.

markstewy: also, should i just omit baseColor from the docs like I did with style?

lukeschunk:
This looks good to me - we'll get it in on the next release :shipit:

markstewy:
thanks @lukeschunk ! should I just leave deprecated baseColor in the docs? or should I remove it like I did with style?

lukeschunk:
I think it's good as is - merging these in and i'll cut a release with your changes!

PR #703, opened 2018-02-05
Luke/Simple Select pass e 

lukeschunk:
`this.props.onScrimClick` can take the event as an argument so you can act on the event in the parent if need be. We weren't passing the event when `this.props.onScrimClick()` was called when the escape key was pressed.

### DIFF
--- a/docs/components/SimpleSelectDocs.js
+++ b/docs/components/SimpleSelectDocs.js
@@ -63,7 +63,7 @@ class SimpleSelectDocs extends React.Component {
         <p>An array of objects that specify <em>icon</em>, <em>isSelected</em>, <em>text</em>, and/or <em>onClick</em> of the item..</p>
 
         <h5>onScrimClick <label>function</label></h5>
-        <p><em>onClick</em> handler for the menu scrim. Function will be passed the event as its first argument.</p>
+        <p><em>onClick</em> handler for the menu scrim.</p>
 
         <h5>scrimClickOnSelect <label>boolean</label></h5>
         <p>Boolean that triggers a scrim click on select. Can be used to auto-close.</p>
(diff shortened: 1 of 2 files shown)


jtanner:
Can you update the documentation for `onScrimClick` too?

src/components/SimpleSelect.js
```diff
@@ -56,7 +56,7 @@ class SimpleSelect extends React.Component {
   _handleKeyDown = (e) => {
     if (keycode(e) === 'esc') {
       e.preventDefault();
-      this.props.onScrimClick();
+      this.props.onScrimClick(e);
```
jtanner: I don't think `onScrimClick` should know about the event. You should instead call `stopPropagation` in the click handler on line 105.

docs/components/SimpleSelectDocs.js
```diff
@@ -63,7 +63,7 @@ class SimpleSelectDocs extends React.Component {
         <p>An array of objects that specify <em>icon</em>, <em>isSelected</em>, <em>text</em>, and/or <em>onClick</em> of the item..</p>
 
         <h5>onScrimClick <label>function</label></h5>
-        <p><em>onClick</em> handler for the menu scrim.</p>
+        <p><em>onClick</em> handler for the menu scrim. Function will be passed the event as its first argument.</p>
```
jtanner: And remove this.

jtanner: Sorry, I should have asked you more about it before asking you to update the docs 😬

lukeschunk:
@jtanner addressed feedback - what do you think - could this be considered a breaking change? If people are relying on the event in the callback func that they're passing down, things will break...

jtanner:
@lukeschunk about this being a breaking change, I would consider it a bug fix because we were inconsistently passing the event in some cases but not in another. So users relying on this passing the event would have had to handle the case where there wasn't an event passed anyway.

Definitely mention it in the release notes though.

Issue #642, opened 2017-08-11
Date Range Picker (test/new) - Last Year should show last calendar year


lukeschunk:
The `Last Year` option should show the previous calendar year rather then the last 365 days like it is right now. 

Ie, right now, last year should show `January 1, 2016 - December 31, 2016`

cerinman:
This has been resolved.  Closing.

PR #704, opened 2018-02-10
Fix - Issue #642 : select last year

markstewy:
Fix for issue #642 "Date Range Picker (test/new) - Last Year should show last calendar year"

I will now display the last calendar year instead of the last 365 days to the current day.


lukeschunk:
@markstewy this looks great! Thanks so much 🎉

lukeschunk:
:shipit:

PR #706, opened 2018-02-22
Remove getComputedStyle

jmophoto:
This removes the `window.getComputedStyle` from the RangeSelector and replaces it with component.offsetWidth.

The `getComputedStyle` was triggering a honeybadger error in the segment builder.

`Failed to execute 'getComputedStyle' on 'Window': parameter 1 is not of type 'Element'.`


lukeschunk:
:shipit:

PR #707, opened 2018-02-28
Fix Calendar Component accessibility

cerinman:
#### Fixes accessibility issue with Calendar component. 

We were running into issues internally not being able to tab to anything on the Calendar component because of the focus call we were doing in the ref callback on the day anchor tag interfering with how the Focus Trap library manages tabbing in the Drawer component works.  This MR fixes that issue and a few other minor things.

- Removes weird focus call in ref call back
- Moves focusing of day to componentDidUpdate
- Fixes moment object mutation bug on focusedDay
- Fixes typo for onClick handler on go back a month calendar button
- Groups click handlers for month nav buttons together

### DIFF
--- a/docs/components/Changelog.js
+++ b/docs/components/Changelog.js
@@ -7,6 +7,11 @@ class Changelog extends React.Component {
         <h1>Change Log</h1>
 
         <h2>MX React Components V 5.0</h2>
+        <h3>5.1.21</h3>
+        <ul>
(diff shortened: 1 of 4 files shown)


lukeschunk:
nice - way to use the lifecycle methods :shipit:

jtanner:
Looks good 👍

PR #709, opened 2018-03-02
Fixed console log statement when onscrimclick is false

joshwilcken:
When closOnScrimClick is false the console statement no longer appears


lukeschunk:
Hey @joshwilcken thanks so much for the contribution. I'm actually going to close this PR in favor of https://github.com/mxenabled/mx-react-components/pull/710. I think it makes more sense to always pass a function to that prop, rather than sometimes null. 

Your fix would totally have worked though; if you're interested in contributing more in the future, feel free. Also if you have any questions on issues or the repo ask away!

Issue #708, opened 2018-03-02
Drawer Component > closeOnScrimClick


redperk:
So we have a boolean check to decide whether or not clicking the scrim will close the drawer.
`<div onClick={this.props.closeOnScrimClick && this.close} style={styles.scrim} />`

`onClick` is expecting a function, which is all good and fine if `closeOnScrimClick === true`, but if it's false `onClick` receives a boolean instead of a function and we get an annoying console warning:

![image](https://user-images.githubusercontent.com/6424668/36876884-839f49f4-1d74-11e8-8531-0f38d3d2ab31.png)

cerinman:
This has been resolved.  Closing.

PR #710, opened 2018-03-06
MBD - update onclick

marabd:
update drawer component to make `onclick` a function


lukeschunk:
:shipit:

lukeschunk:
looks like this other PR also accomplishes the same thing https://github.com/mxenabled/mx-react-components/pull/709 (gets rid of the console error mentioned here https://github.com/mxenabled/mx-react-components/issues/708) 

Do we have a preference? I lean towards this PR.

PR #711, opened 2018-03-13
Drawer Tweaks

cerinman:
- Fixes accessibility issue with title of drawer by wrapping it in an H1 tag and tweaking it styling so to continues to look the way it did before.
- Allows children of Drawer to be a function so that we can pass the children the drawer's close function
- Allows headerMenu prop to be a function so that we can pass the headerMenu the drawer's close function
- Updates the Drawers `_renderNav` function to take the navConfig as an argument rather than accessing the prop directly.

I've added the function as children functionality here because we have gotten into the habit internally of adding `refs` to drawers to get at its `close` function to ensure that the drawer animation fires before unmount.  

This is fragile because the ref can change due to things like wrapping it with redux connect.  I feel that getting at the drawers close method via a function call is a way more stable code path so I've put together this PR. 

### Demo of Drawer with function as children

drawerdemo

### DIFF
--- a/src/components/Drawer.js
+++ b/src/components/Drawer.js
@@ -33,7 +33,10 @@ class Drawer extends React.Component {
     duration: PropTypes.number,
     easing: PropTypes.array,
     focusOnLoad: PropTypes.bool,
-    headerMenu: PropTypes.element,
+    headerMenu: PropTypes.oneOfType([
+      PropTypes.element,
+      PropTypes.func
+    ]),
     headerStyle: PropTypes.oneOfType([
       PropTypes.array,
       PropTypes.object
--- a/src/components/Drawer.js
+++ b/src/components/Drawer.js
@@ -222,9 +222,9 @@ class Drawer extends React.Component {
                     </Button>
                   }
                 </span>
-                <span style={styles.title}>
+                <h1 style={styles.title}>
                   {this.props.title}
-                </span>
+                </h1>
                 <div style={styles.headerMenu}>
                   {this.props.headerMenu ? this.props.headerMenu : this.props.navConfig && this._renderNav(styles, theme)}
                 </div>
--- a/src/components/Drawer.js
+++ b/src/components/Drawer.js
@@ -310,7 +310,13 @@ class Drawer extends React.Component {
         }
       },
       title: {
+        alignItems: 'center',
+        display: 'flex',
         flex: '1 0 auto',
+        fontSize: theme.FontSizes.MEDIUM,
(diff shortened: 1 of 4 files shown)


src/components/Drawer.js
```diff
@@ -222,15 +240,15 @@ class Drawer extends React.Component {
                     </Button>
                   }
                 </span>
-                <span style={styles.title}>
+                <h1 style={styles.title}>
                   {this.props.title}
-                </span>
+                </h1>
```
lukeschunk: the `<h1>` here is tough - I can see us running into potential issues in the future where the title should not be an `<h1>`... I think what could be a good solution would be to make the title a prop that takes a tiny functional component, then the user could do something like this: 

```
<Drawer title={() => <h1>If I want an h1 <h1>} /> 
or 
<Drawer title={() => <h2>If I want an h2 <h2>} /> 
```
I think you would just have to do something with the styling to normalize the styles no matter what tag the user used here...

lukeschunk: ooooh what about this: 
```
<Drawer title={({ style }) => <h1 style={style}>Here we go</h1>
```

and then within the Drawer the code would look like this: 
```
<this.props.title style={{ ...here youd put the styles you need to normalize }} /> 
```

cerinman: I considered what you suggested when I first started working on this but settled on the h1 directly because I couldn't think of a single use case where a drawer title shouldn't be an h1.  It describes the drawers contents.  It's also easier to remember that children of the drawer should use h2 or less when I know that the title is always an h1.

lukeschunk: alright - also a lot of the research i've done lately says multiple H1 tags on a page is ok, so I suppose we could be fine even if there was <h1> content within the drawer

cerinman: If you want to chat in person about it I'm more than happy to.

jtanner:
On second thought, I wonder if we would want to expose more than just the close function in the future? 🤔 

If so then this could pass an object to the `children` function like this:

```js
<Drawer>{ ({ close }) => ... }</Drawer>
```

What do you think about that future proofing idea @cerinman?

cerinman:
Yeah I like that @jtanner.  I'll make the change.

src/components/Drawer.js
lukeschunk: no we're on the same page 👍 i was just thinking out loud with my last comment

cerinman:
Updated @jtanner

PR #713, opened 2018-03-13
Luke/use glamor in button

lukeschunk:
This PR 
* Adds glamor as a dependency 
* Uses Glamor in the Button component ( I think this will fix a bug internally )

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -1,6 +1,7 @@
 const React = require('react');
 const PropTypes = require('prop-types');
-import { css } from 'glamor'
+
+import { css } from 'glamor';
 
 const Icon = require('./Icon');
 const Spin = require('./Spin');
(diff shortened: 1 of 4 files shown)


src/components/Button.js
```diff
@@ -66,9 +67,9 @@ class Button extends React.Component {
 
     return (
       <button
+        className={css(Object.assign({}, styles.component, styles[this.props.type], style))}
```
cerinman: I know the object assign was already there but could you switch it object spread while you're here?

src/components/Button.js
```diff
@@ -248,4 +249,4 @@ class Button extends React.Component {
   };
 }
 
-module.exports = Radium(Button);
+module.exports = Button;
```
cerinman: Do you think that removing `Radium` will fix the bug you are experiencing?

lukeschunk: i do

src/components/Button.js
lukeschunk: 👍

cerinman:
:shipit:

PR #714, opened 2018-03-14
Add aria label to Drawer's close button

cerinman:
Before we had visually hidden text in the close button but it wasn't very descriptive.  This PR removes the visually hidden text and replaces it with a more descriptive aria label on the button it self.  I also updated the docs accordingly.

cc @lukeschunk


phantomxc:
:shipit:

PR #715, opened 2018-03-16
Close Drawer on press of esc key

cerinman:
I noticed that we had a few places internally where we added key board listeners to close drawers when the `esc` key is pressed.  It's weird to handle it outside the component because you then have to reach into the drawer to call it's close method.  This PR adds the listener to the Drawer's wrapping component div.  See gif for drawer closing on esc.

### gif

esc-close

### DIFF
--- a/docs/components/Changelog.js
+++ b/docs/components/Changelog.js
@@ -7,10 +7,16 @@ class Changelog extends React.Component {
         <h1>Change Log</h1>
 
         <h2>MX React Components V 5.0</h2>
+        <h3>5.1.25</h3>
+        <ul>
+          <li>Adds onKeyUp prop to Drawer component(<a href='https://github.com/mxenabled/mx-react-components/pull/715'>#715</a>)</li>
+        </ul>
+
         <h3>5.1.24</h3>
         <ul>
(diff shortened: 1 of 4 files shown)


lukeschunk:
i'm sort of surprised this works even though the div isn't focusable

jtanner:
Do we always want to close the `Drawer` on `esc`? Should it be optional?

cerinman:
A DOM element does not have to be focusable to have an event listener.  As long as the element is high enough up the chain then it works.

cerinman:
That is a good question.  I think we do our internal usage but I can see others possible not wanting it?  I can add a prop that defaults to true maybe?

sambev:
I was going to ask the same question as Joe, but if we go the optional route, I would go a step further and say it should just be allowed to be overridden instead of there or not there.

lukeschunk:
right, event bubbling

cerinman:
Ok, I like @sambev 's idea of passing the entire onKeyUp event handler function as a prop and we can default to ours if the prop is null/undefined.  I'll update.

lukeschunk:
:shipit:

PR #716, opened 2018-03-20
Drawer focus trap fix

cerinman:
Adds new prop `focusTrapProps` to be able to pause the focus trap for the drawer if another drawer or modal needs to be mounted/rendered as a child of the drawer.

From the `Focus Trap` library documentation

```
focusTrap.pause()
Pause an active focus trap's event listening without deactivating the trap.

If the focus trap has not been activated, nothing happens.

Returns the focusTrap.

Any onDeactivate callback will not be called, and focus will not return to the element that was focused before the trap's activation. But the trap's behavior will be paused.

This is useful in various cases, one of which is when you want one focus trap within another. demo-six exemplifies how you can implement this.
```

The traps pause function is called when the Focus Trap React component is passed true for it's paused prop. See links below for the details in the Libraries docs.

Focus Trap React Library
https://github.com/davidtheclark/focus-trap-react

Focus Trap Library
https://github.com/davidtheclark/focus-trap

### DIFF
--- a/docs/components/Changelog.js
+++ b/docs/components/Changelog.js
@@ -9,7 +9,7 @@ class Changelog extends React.Component {
         <h2>MX React Components V 5.0</h2>
         <h3>5.1.26</h3>
         <ul>
-          <li>Adds focusTrapPaused prop to Drawer component(<a href='https://github.com/mxenabled/mx-react-components/pull/716'>#716</a>)</li>
+          <li>Adds focusTrapProps prop to Drawer component(<a href='https://github.com/mxenabled/mx-react-components/pull/716'>#716</a>)</li>
         </ul>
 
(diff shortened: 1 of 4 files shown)


cerinman:
ugh, Travis is having issues.

lukeschunk:
code looks good :shipit:

lukeschunk:
did we ever explore adding a blanket prop something like `{...focusTrapProps}` that way if we need to use more of these in the future we can just pass it in that object rather than having to go in and update our OS component every time?

cerinman:
No, I haven't explored that but it is a good idea.  I'll update this PR while I'm waiting on Travis/yarn to start working again. 👍

lukeschunk:
nice - :shipit:

PR #718, opened 2018-03-23
Auto doc

markstewy:
This was an idea i had in response to issue #582 

This approach uses tags ie:
@autoDocSrc(myAssignedName1) and @autoDocsTarget(myAssignedName1)

to automate the markdown for demos. Multiple tags and combinations of tags allow JSX, supporting methods, props or anything else to be organized and displayed in the markdown. The autoDocsLoader parses everything in webpack. There is still a lot of work that needs to be done, but I wanted to get feedback from the team about this overall approach before I do any more work. 

@jtanner this doesn't have any of the tab formatting in place yet, but I would like the end result to look like your screenshots.

### DIFF
--- a/docs/autoDocsLoader.js
+++ b/docs/autoDocsLoader.js
@@ -1,19 +1,43 @@
 module.exports = function(content) {
+  //Tag Aggregation
+  let tagRegex = /(@autoDocsSrc\(?.+\S\)?)\S/g;
+  let tagMatch = [];
+  let tagName = '';
+  let tagList = [];
 
-  let srcRegex = /(@autoDocsSrc\(?.+\S\)?)\S/g;
-  let targetRegex = /(@autoDocsTarget\(?.+\S\)?)\S/g;
-  let targetMatch, targetMatches, srcMatch, srcMatches = [];
-
-  while ((targetMatch = targetRegex.exec(content)) != null) {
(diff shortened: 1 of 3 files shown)


markstewy:
screen shot 2018-03-23 at 3 27 31 pm
screen shot 2018-03-23 at 3 27 48 pm

jtanner:
Thanks @markstewy! I will review this soon.

jtanner:
@markstewy sorry for the delayed response. I think this is an interesting approach, I'd like to see you add some tests which I think would also help you with your remaining TODO tasks.

cerinman:
@markstewy I apologize for the delay. 

jtanner has moved on to another job and I've totally over looked this PR.  We were just discussing internally this week that we need to do something like this to improve documentation and the release process.  

I'm pinging @sambev who is currently on the research task for figuring this out.  Perhaps we can just use what you have here but I'll keep you posted.

paniclater:
Apologies that this has taken so long. We're currently evaluating some new design systems internally and we'll probably look at this or something similar as we input that. Thank you so much for your input and sorry this didn't pan out!

cerinman:
Sorry again for ignoring you here Mark. I wish I had a good excuse but I don't. I apologize for waisting your time here but I'm going to close this PR as this type of thing just isn't a priority for us at the moment.

Issue #585, opened 2017-05-30
Manage FocusTrap for when multiple Drawers are open


jtanner:
@lukeschunk pointed out that when a drawer is opened on top of another drawer and then closed, the focus is no longer trapped on the first drawer.

We need a way to manage what on the page is currently focus trapped and [pause/unpause](https://github.com/davidtheclark/focus-trap#focustrappause) the other traps.

[Demo 6](https://github.com/davidtheclark/focus-trap/blob/master/demo/js/demo-six.js) is an example of managing the traps.

cerinman:
Haha nice, I hadn't even seen this issue when I started working on a fix.  Thanks for pointing it out @jtanner

PR #719, opened 2018-03-26
Issue 585: Fix Nested FocusTraps

cerinman:
Fixes https://github.com/mxenabled/mx-react-components/issues/585

The focus trap library that we use on the `Drawer` and `Modal` components does NOT manage un-pausing a trap when a child trap is un-mounted.  

This MR fixes the issue by wrapping the library in a component and introducing a scoped array of traps outside the component to manage pausing and un-pausing.  

Thanks @jtanner for pairing with me on this one.

We've also added some simple tests and updated the `Drawer` docs for future testing purposes.

Here is a gif of the update docs to show that things are still working as expected.

focus-trap

### DIFF
--- a/src/components/MXFocusTrap.js
+++ b/src/components/MXFocusTrap.js
@@ -38,32 +38,4 @@ class MXFocusTrap extends React.Component {
   }
 }
 
-/**
- * For testing purposes.  This provides an easy way to
- * unmount a child focus trap to ensure the parent is
- * unpaused. This component is only used in MXFocusTrap-test.js
- **/
-class WrappedMXFocusTrap extends React.Component {
-  state = {
-    renderChildTrap: true
-  }
-
-  render () {
-    return (
-      <MXFocusTrap>
-        <button>"Outer Focusable Button"</button>
-        {this.state.renderChildTrap ? (
-          <MXFocusTrap>
-            <button>"Inner Focusable Button"</button>
-          </MXFocusTrap>
-        ) : null}
-      </MXFocusTrap>
-    );
-  }
-}
-
-
-module.exports = {
-  MXFocusTrap,
-  WrappedMXFocusTrap
-};
+module.exports = MXFocusTrap;
--- a/src/components/MXFocusTrap.js
+++ b/src/components/MXFocusTrap.js
@@ -1,0 +1,41 @@
+let traps = [];
+
+const React = require('react');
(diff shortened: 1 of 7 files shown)


jmophoto:
:shipit: 🍪

src/components/MXFocusTrap.js
```diff
@@ -0,0 +1,69 @@
...
+    );
+  }
+}
+
+/**
+ * For testing purposes.  This provides an easy way to
+ * unmount a child focus trap to ensure the parent is
+ * unpaused. This component is only used in MXFocusTrap-test.js
+ **/
+class WrappedMXFocusTrap extends React.Component {
```
jtanner: Because this is only for testing I think it should live in the test file. Can you please move it there?

cerinman: Sure, I actually thought about that but then second guessed myself.  I'll move it. 👍

jmophoto: Good call @jtanner. I read that comment wrong and was thinking that would be used for external testing when implementing the module.

src/components/__tests__/MXFocusTrap-test.js
```diff
@@ -0,0 +1,26 @@
...
+    /**
+     * WrappedMXFocusTrap renders a MXFocusTrap within a MXFocusTrap
+     * and provides a way to unmount the child trap by setting
+     * state on the wrapper.
+     **/
+    const wrapper = mount(<WrappedMXFocusTrap />);
+    const parentTrap = wrapper.find(MXFocusTrap).first().instance();
+
+    // Parent should be un-paused when child is unmounted.
+    wrapper.setState({ renderChildTrap: false });
```
jtanner: I like how you setup `WrappedMXFocusTrap` to easily test mounting and unmounting a child. Nice work on getting this tested 👍

src/components/MXFocusTrap.js
jtanner: It will also make it so you can change the imports to:

```
const MXFocusTrap = require('../components/MXFocusTrap');
```

cerinman:
Moves `WrappedMXFocusTrap` to the test file @jtanner.  This is ready for another review.

jtanner:
Thanks @cerinman, it looks great 🍪

jmophoto:
:shipit:

Issue #717, opened 2018-03-23
Create a ThemeProvider so passing a theme prop is unnecessary


jtanner:
I think we should add `ThemeProvider` and `ThemeConsumer` components to make passing the `theme` prop optional for all components that have that prop.

"A new, safe version of context is under development" for the upcoming React 16.3 release [[1]](https://reactjs.org/docs/context.html) and even though the new release is still in alpha I think we can design these components as if it were already available. While reading an article on the new context API [[2]](https://css-tricks.com/putting-things-in-context-with-react/) I realized that we can use "Render Props" and the current context API *now* and then when 16.3 is ready the implementation can be easily updated to the new API.

----

Here's an example of how it could work:

```js
<ThemeProvider theme={{ Colors: { PRIMARY: '#C0FFEE' } }}>
  <Button>Click me!</Button>
</ThemeProvider>
```

*Button.js*
```js
...
render() {
  return (
    <ThemeConsumer>
      {theme =>
        ...
      }
    </ThemeConsumer>
  )
}
```

PR #721, opened 2018-03-27
Joe/theme provider

jtanner:
Issue: #717

This is the first iteration of adding `<ThemeProvider>` to make theming easier. `<ThemeProvider>` should be instantiated at the root of an application, similar to Redux's `<Provider>`.

To use the theme provided by `<ThemeProvider>` each theme-able component must use `<ThemeContext>` in its `render()` function. I added `<ThemeContext>` only to `<Button>` for now, once I get a 👍 for the approach then I will work on updating the other components.

What do you all think of how the changes in `<Button>` turned out?

UPDATE:
I followed the "Render Props" idea suggested [[here]](https://css-tricks.com/putting-things-in-context-with-react/#article-header-id-5) for the `<ThemeContext>` API. I chose to use `children` as a function instead of a `render` prop because it more closely matches the new React context API. That way, once React 16.3 is released only `<ThemeProvider>` and `<ThemeContext>` need minor updates and the other components will require no changes.

----

Here's what the updated docs looks like:

screen shot 2018-03-27 at 4 15 31 pm

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -3,7 +3,7 @@ const PropTypes = require('prop-types');
 
 import { css } from 'glamor';
 
-import { ThemeContext } from './Theme';
+import { withTheme } from './Theme';
 
 const Icon = require('./Icon');
 const Spin = require('./Spin');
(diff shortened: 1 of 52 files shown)


phantomxc:
I'm not sure you should update each component to automatically use it. I think I would just make ThemeProvider a component you could put at the top level of your application and forget about it.

jtanner:
@phantomxc I think what I've proposed here is what you're suggesting, i.e. use `ThemeProvider` at the top-level of an app. If each component is not updated then how else would they get access to the theme set in the `ThemeProvider`? Can you give me an example of how you're thinking it would work instead?

jtanner:
I forgot to mention in the description that I followed the "render props" context recipe described in https://css-tricks.com/putting-things-in-context-with-react/#article-header-id-5 which will make switching to the new React context API super simple.

I will update the description...

cerinman:
Let me make sure I understand this correctly.

We would import the `ThemeProvider` component internally and user it at the top level of the app there.  That would then expose the theme in context to all the children of the app including the open source components here. 

 Is my understanding correct?

src/components/Button.js
```diff
@@ -59,38 +61,42 @@ class Button extends React.Component {
...
-          <div style={styles.buttonText}>
-            {isActive ? actionText : children}
-          </div>
-        </div>
-      </button>
+      <ThemeContext>{contextTheme => {
+        // Manually consume everything that isn't going to be passed down to the button so we don't have to keep adding props one at a time.
+        // Keep elementProps for backwards compatibility.
+        const { actionText, buttonRef, children, elementProps, icon, isActive, primaryColor, style, theme: propTheme, ...rest } = this.props;
+        const theme = StyleUtils.mergeTheme(contextTheme || propTheme, primaryColor);
```
cerinman: And the merge here would handle if contextTheme was undefined?

cerinman: I just reread this.  Somehow I missed the `||` check you are doing the first time around.  Never mind on this question...

jtanner:
@cerinman yes that is correct.

Your question made me think that `theme` in the context should probably be scoped. That way it won't conflict with other apps that could possibly also be using the same key in the context. Maybe `mxTheme`?

cerinman:
`mxTheme` works for me.

phantomxc:
Ignore my comment.

jtanner:
Looks like React 16.3.0 just landed: https://reactjs.org/blog/2018/03/29/react-v-16-3.html

I'm going to update this to include some of the ideas presented in that blog post. I think `withTheme()` will make updating all of the components a breeze.

jtanner:
Updating the components using `withTheme(...)` was a breeze. However, I ran into a snag, it turns out the enzyme testing library we use doesn't support the new Context API in React 16.3.0 😢 

https://github.com/airbnb/enzyme/issues/1509

I'm not sure I want to include the suggested fix [[see here]](https://github.com/airbnb/enzyme/issues/1509#issuecomment-377726374) in this PR.

jtanner:
@cerinman I reverted the use of the new React 16.3 `createContext` and `forwardRef` functions which fixed the issue with the tests. Once enzyme is updated we can revert https://github.com/mxenabled/mx-react-components/pull/721/commits/a322aa1c7cf3ff0894f864c4b44bd9d25e192167 in another PR to start using the new API.

This is ready for another look 👀

jtanner:
@cerinman this is ready for review again 👀 

Besides the 16.3 issue I also ran into a bug with `Tabs` that I fixed in https://github.com/mxenabled/mx-react-components/pull/721/commits/746ca3d0d280bc4f0963af2395f0a2374e6a8473

redperk:
:shipit:

PR #723, opened 2018-04-02
Upgrade react and react-router

jtanner:
This upgrades `react` to 16.3.0 and `react-router` to 4.2.0. These changes are needed so I can update https://github.com/mxenabled/mx-react-components/pull/721 to the new React Context API.

~~Also, I noticed that the docs `package.json` file was not depending on `mx-react-components` and hence was duplicating some dependencies. `react` was outdated and didn't match the dependency in the main `package.json` file, **it also caused an error that was difficult to debug when trying to use the new Context API**. They should now stay in sync better with `mx-react-components` as an
explicit dependency.~~

UPDATE:
Depending on `mx-react-components` turned out to be problematic because it pointed to `dist/Index.js` which was not updated during hot reloading. Rather than try to update the `yarn dev` script I undid the change.

### DIFF
--- a/docs/app.js
+++ b/docs/app.js
@@ -1,51 +1,11 @@
 const React = require('react');
 const ReactDOM = require('react-dom');
-const { hashHistory, Route, IndexRoute, Router } = require('react-router');
+const { Route, Switch } = require('react-router');
+const { HashRouter } = require('react-router-dom');
 
-const BarChart = require('components/BarChartDocs');
-const Button = require('components/ButtonDocs');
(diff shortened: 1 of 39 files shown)


jtanner:
Thanks for testing it out @sambev 🙌

cerinman:
Ditto to what @sambev said.  Looks good.  :shipit:

cerinman:
This has merge conflicts and Joe is no longer at MX.  Closing this for now.  We can revisit in the near future.

PR #724, opened 2018-04-02
Luke/Date Range Picker Design Updates 

lukeschunk:
Updates the `DateRangePicker` as per these design mocks: https://gitlab.mx.com/mx/widgets/issues/13#note_327298 (internal). 

The general new paradigm is, now a user can "cancel" their change. I did this by utilizing more component state. When the Modal is open, the dates are managed using component state, and when the Modal is closed, the state is either reset back to what the props were (if the user cancelled), or a callback is fired to update the parent (if the user clicked apply). 

I also added a `FocusTrap` for the `DateRangePicker` modal. 

### Tweaks from Tommy

- Removes `closeCalendarOnRangeSelect` prop.  I did this for two reasons.  1: it was no longer needed after adding the apply and cancel buttons. 2: It's current implementation was broken anyway so I doubt anyone was ever setting it to true.  I don't feel like this is a breaking change for both these reasons as well.
- Adds new prop `onDateRangeSelect`. Now that we have an apply button the naming of the prop `onDateSelect` was weird and miss leading.
- Switch to use MXFocusTrap and add new prop for passing options to FocusTrap
- Deprecates the `onDateSelect` prop
- Updated the docs for the DateRangePicker
- Resolved the issues I noted for Luke before taking this over.
- Reverted prettier and yarn.lock changes to make review easier.

### Canceling/Apply On Desktop 
apply_gif

### Mobile 
mobile drp gif


src/components/DateRangePicker.js
```diff
@@ -269,35 +365,46 @@ class DateRangePicker extends React.Component {
...
           {shouldShowCalendarIcon ? (
-            <Icon
-              size={20}
-              style={styles.selectedDateIcon}
-              type='calendar'
-            />
+            <Icon size={20} style={styles.selectedDateIcon} type='calendar' />
           ) : null}
           <div style={styles.selectedDateText}>
+            {/*These below are based on props, not state, as it displays when the date picker is closed - should match the parent*/}
```
cerinman: I don't think this comment is needed.  It reads weird anyway and I'm not even sure what you mean by it.

src/components/DateRangePicker.js
```diff
@@ -307,256 +414,359 @@ class DateRangePicker extends React.Component {
...
-                        { label: 'T', value: 'Tuesday' },
-                        { label: 'W', value: 'Wednesday' },
-                        { label: 'T', value: 'Thursday' },
-                        { label: 'F', value: 'Friday' },
-                        { label: 'S', value: 'Saturday' }].map((day) => {
-                          return (
-                            <div key={day.value} style={styles.calendarWeekDay}>
-                              {day.label}
+            {this.state.showSelectionPane && (
+              <FocusTrap>
```
cerinman: If we're going to use `FocusTrap` here then we need to add a prop that can be passed to it to control its options like we do in the drawer and the modal

src/components/DateRangePicker.js
```diff
@@ -307,256 +414,359 @@ class DateRangePicker extends React.Component {
...
+                              selectedStartDate={this.state.selectedStartDate}
+                              styles={styles}
+                              theme={theme}
+                            />
+                          )}
+                        </div>
+                      )}
+
+                      {(this.state.showCalendar ||
+                        isLargeOrMediumWindowSize) && (
```
cerinman: Should this conditional be `&&`?  It seems weird that you are saying showCalendar or is large or medium.

src/components/DateRangePicker.js
```diff
@@ -307,256 +414,359 @@ class DateRangePicker extends React.Component {
...
+                            this.setState({
+                              focusedDay: this.props.selectedStartDate,
+                              selectedStartDate: this.props.selectedStartDate,
+                              selectedEndDate: this.props.selectedEndDate
+                            });
+                            this._handleScrimClick();
+                          }}
+                          type='secondary'
+                        >
+                          Cancel
```
cerinman: I can already tell that this button is not descriptive enough for accessibility.  We should probably add an aria label with a better description than just `Cancel`

src/components/DateRangePicker.js
```diff
@@ -307,256 +414,359 @@ class DateRangePicker extends React.Component {
...
-                          onClick={() => this.setState({ showCalendar: false })}
-                          theme={theme}
+                          onClick={() => {
+                            this.props.onDateSelect(
+                              this.state.selectedStartDate,
+                              this.state.selectedEndDate
+                            );
+                            this._handleScrimClick();
+                          }}
+                          style={styles.applyButton}
```
cerinman: Same thing here.  Lets add a more descriptive aria label for this apply button

cerinman:
This is s significant change to the DateRangePicker but I don't see anything here in this PR that updates the docs.

cerinman:
Also, it looks like you're not handling the case where the `closeCalendarOnRangeSelect` prop is true.  In that case we need to decide if we even want to show the cancel and apply buttons at all since selecting the range would also apply it.

src/components/DateRangePicker.js
cerinman: This also needs to be the MXFocusTrap as well.

lukeschunk:
@cerinman 🆗

cerinman:
@jtanner this can use a review now!

cerinman:
No worries.  I reverted them mostly for myself.  I wanted to make sure I didn't miss anything you changed while reviewing.  Took less than 30 minutes.

cerinman:
@jtanner I didn't add tests for this on this go around because of the high priority.  I plan on adding them in a follow up PR.

src/components/DateRangePicker.js
```diff
@@ -273,12 +285,22 @@ class DateRangePicker extends React.Component {
...
+      paused: false,
+      ...this.props.focusTrapProps
+    };
 
     return (
       <div style={styles.component}>
         <a
           onClick={this._toggleSelectionPane}
-          onKeyUp={(e) => keycode(e) === 'enter' && this._toggleSelectionPane()}
+          onKeyDown={(e) => keycode(e) === 'enter' && this._toggleSelectionPane()}
```
jtanner: Why the change from keyup to keydown?

docs/components/DateRangePickerDocs.js
```diff
@@ -55,6 +52,12 @@ class DateRangePickerDocs extends React.Component {
         `}
         </Markdown>
 
+        <h5>focusTrapProps<label>Object</label></h5>
+        <p>Default: Empty Object</p>
+        <p>The DateRangePicker component uses the <a href='https://github.com/davidtheclark/focus-trap-react'>Focus Trap React</a> library to prevent a user from tabing outside the date range picker for accessibility reasons.</p>
+        <p>The focusTrapProps object provides a mechanism for passing the focus trap component props.</p>
```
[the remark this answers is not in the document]
cerinman: Yep

src/components/DateRangePicker.js
cerinman: I'm not 100% sure why Luke did this but I would guess he was fixing an event listener over lap with the cancel/apply buttons.  They have key down listeners so I'm assuming the key up here was firing immediatley upon using one of those buttons. 

If you're opposed to this I could switch it back and then preventDefault/stop propagation on the buttons.

cerinman:
Tested in IE/Edge and found a minor styling bug.  That is now fixed and here are some screen of IE at desktop and mobile resolutions.

screen shot 2018-04-10 at 10 40 16 am
screen shot 2018-04-10 at 10 54 31 am
screen shot 2018-04-10 at 10 54 43 am

src/components/DateRangePicker.js
jtanner: It's fine how it is, I was just curious.

jtanner:
:shipit:

PR #726, opened 2018-04-10
Button hover fix

jtanner:
I noticed while testing some other changes that the `Button` component no longer changed color on mouse hover. It turned out that there was a regression in the logic in `_windowSizeIsSmall`.


cerinman:
:shipit:

PR #727, opened 2018-04-10
Theme the cancel and apply buttons in DateRangePicker

cerinman:
🤦‍♂️ I forgot to add theme to the cancel and apply buttons just added in the `DateRangePicker` component.  This fixes that glaring mistake...

screen shot 2018-04-10 at 1 54 53 pm


jmophoto:
:shipit:

PR #728, opened 2018-04-10
DateRangePicker - fix state issues

cerinman:
### TLDR

- Consolidate reset logic for closing the range selector
- Fix `componentWillReceiveProps` logic

### Full Story

After more testing internally, I discovered that we had state issues as it pertained to Transactions and Spending both using the `DateRangePicker`.  If you set a range in one and then changed to the other, the date picker there retained the date range from the previous widget. This was caused by faulty logic in two locations. 1: Resetting on close and 2: handling new props in `componentWillReceiveProps`

Ugh!  I hate passing props off to state for components like this.  There are so many edge cases to get wrong but I can't think of a better way at the moment to handle this stand alone `DateRangePicker` component.

I've spent the entire afternoon testing internally and I'm pretty sure I have this right now.

### DIFF
--- a/src/components/DateRangePicker.js
+++ b/src/components/DateRangePicker.js
@@ -159,9 +159,11 @@ class DateRangePicker extends React.Component {
       if (isLargeOrMediumWindowSize) this.setState({ selectedBox: SelectedBox.FROM });
     }
 
+    const modifiedRangeCompleteButDatesInversed = startDate && endDate && this._endDateIsBeforeStartDate(startDate, endDate);
+
     this.setState({
-      selectedStartDate: startDate,
-      selectedEndDate: endDate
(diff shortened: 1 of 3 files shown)


src/components/DateRangePicker.js
```diff
@@ -230,10 +226,7 @@ class DateRangePicker extends React.Component {
 
   _handleScrimClick = () => {
     this.props.onClose();
```
jtanner: `_handleScrimClick` is now used for more than just the scrim element so the name doesn't fit anymore. Also, `_resetToPropValuesAndClose` is only used in that one function so I think the two functions could be combined. 

What do you think about moving this line in to `_resetToPropValuesAndClose` and getting rid of `_handleScrimClick`?

cerinman: I agree with that.  I'll update.

PR #729, opened 2018-04-12
Don't clobber the className when it's passed as a prop

jtanner:
The footer buttons in the Modal component set a `className` prop which was clobbering the `className` in Button, which is needed for glamor css styling. This fixes the issue and adds a test to prevent regressions.

Here's how the Modal footer looked while broken (taken from the docs page):

screen shot 2018-04-12 at 2 13 59 pm

Here is it fixed:

screen shot 2018-04-12 at 2 14 17 pm


redperk:
:shipit:

PR #732, opened 2018-04-14
Ensure withTheme wrapped components work with ref

jtanner:
Components themed using `withTheme` were not working with `ref` props. This converts the wrapping component in `withTheme` from a functional component to a `React.Component` which fixes the problem.

### DIFF
--- a/src/components/SimpleInput.js
+++ b/src/components/SimpleInput.js
@@ -13,7 +13,7 @@ const {
   deprecateProp
 } = require('../utils/Deprecation');
 
-class Input extends React.Component {
+class SimpleInput extends React.Component {
   static propTypes = {
     baseColor: PropTypes.string,
     elementProps: PropTypes.object,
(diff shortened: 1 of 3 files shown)


src/components/SimpleInput.js
```diff
@@ -13,7 +13,7 @@ const {
   deprecateProp
 } = require('../utils/Deprecation');
 
-class Input extends React.Component {
+class SimpleInput extends React.Component {
```
paniclater: A+ thanks for cleaning this up.

paniclater:
:shipit: I would definitely be interested to learn more what the diffs are between the functional components and class components are that cause stuff like this. I wouldn't have thought converting the func to a class would do it, but glad it is working!

jtanner:
@paniclater in https://reactjs.org/docs/refs-and-the-dom.html the docs say it's because of the following:
> You may not use the ref attribute on functional components because they don’t have instances.

Had I not needed to rollback the React 16.3 changes in my other PR this wouldn't have been necessary because `withTheme` was using the new `React.forwardRef` (https://reactjs.org/docs/forwarding-refs.html).

Alas, sometimes it's painful to use the bleeding edge releases.

PR #733, opened 2018-04-16
Fix modal focus

cerinman:
- Adds new prop focusOnLoad to control focusing the Modal on mount
- Adds new prop focusTrapProps to control focus trap
- Cleans up deprecation warning for the `isOpen` prop
- Updates the Docs with the changes

This will be included in the 5.2.3 release in https://github.com/mxenabled/mx-react-components/pull/731

### DIFF
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -191,7 +191,6 @@ class Modal extends React.Component {
       focusTrapOptions: {
         clickOutsideDeactivates: true
       },
-      paused: false,
       ...this.props.focusTrapProps
     };
 
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -11,7 +11,7 @@ const _merge = require('lodash/merge');
 const { themeShape } = require('../constants/App');
 
 const StyleUtils = require('../utils/Style');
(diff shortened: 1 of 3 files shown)


src/components/Modal.js
```diff
@@ -182,9 +187,16 @@ class Modal extends React.Component {
   render () {
     const theme = StyleUtils.mergeTheme(this.props.theme, this.props.color);
     const styles = this.styles(theme);
+    const mergedFocusTrapProps = {
+      focusTrapOptions: {
+        clickOutsideDeactivates: true
+      },
+      paused: false,
```
jtanner: Why did you include this line?

jtanner: It looks like MXFocusTrap will override this prop with it's state: https://github.com/mxenabled/mx-react-components/blob/66a588289d0bcd656b81975b9866a4398b399615/src/components/MXFocusTrap.js#L34

cerinman: I was thinking that it was required because of the the `MXFocusTrap` stuff we did but it looks like `MxFocusTrap` defaults to false on its own.  I'll remove it.

PR #734, opened 2018-04-16
Select component children function handler and elementRef fix

jtanner:
When custom `children` are supplied to `Select` there is no way to let `Select` know that an option has been selected. This updates `Select` to pass `onOptionClick` handler to `children` when it's a function, the handler sets the selected option and also calls the `onChange` prop.

I also made it so the `elementRef` is actually called when provided.


cerinman:
:shipit:

PR #736, opened 2018-04-19
Drawer Accessibility via role and aria attributes

cerinman:
### Drawer Component

- Adds role prop which defaults to `dialog`
- Adds aria-describedby prop
- Adds aria-labelledby prop
- Updates the docs for Drawer

Because of the above we can now use elements that our children of the drawer to describe it's content for screen readers.  This is a vast improvement in accessibility for the Drawer.  Here is the docs on the `dialog` role for clarity around the aria attributes added.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_dialog_role

### DIFF
--- a/docs/components/DrawerDocs.js
+++ b/docs/components/DrawerDocs.js
@@ -237,7 +237,6 @@ class DrawerDocs extends React.Component {
       title='Demo'
     >
       <p id='description'>This is a demo drawer</p>
-      <div>// Content Here</div>
     </Drawer>
   `}
         </Markdown>
--- a/docs/components/DrawerDocs.js
+++ b/docs/components/DrawerDocs.js
@@ -42,6 +42,7 @@ class DrawerDocs extends React.Component {
 
     return (
       <Drawer
+        aria-describedby='description'
         breakPoints={{ large: 1200, medium: 1100 }}
         contentStyle={styles.content}
         headerMenu={({ close }) => (
@@ -62,7 +63,7 @@ class DrawerDocs extends React.Component {
           return (
             <div>
               {this.state.clickedMenu && <code>You clicked: {this.state.clickedMenu.text}</code>}
-              <p>
+              <p id='description'>
                 Drawer Component
               </p>
               <p>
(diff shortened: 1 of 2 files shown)


src/components/Drawer.js
```diff
@@ -268,14 +273,14 @@ class Drawer extends React.Component {
                     />
                   }
                 </span>
-                <h1 style={styles.title}>
+                <h1 id='mx-drawer-title' style={styles.title}>
```
jtanner: I think the `id` here is too general. If there are multiple `Drawer` components open won't the `id`s conflict?

You could use lodash's `uniqueId` to make the id more unique.

src/components/Drawer.js
```diff
@@ -268,14 +273,14 @@ class Drawer extends React.Component {
...
                   {this.props.title}
                 </h1>
                 <div style={styles.headerMenu}>
                   {menu}
                 </div>
               </header>
-              <div style={Object.assign({}, styles.content, this.props.contentStyle)}>
+              <div style={{ ...styles.content, ...this.props.contentStyle }}>
```
jtanner: Thanks for updating these 🙌

docs/components/DrawerDocs.js
```diff
@@ -42,6 +42,7 @@ class DrawerDocs extends React.Component {
 
     return (
       <Drawer
+        aria-describedby='description'
```
jtanner: Do you want to also update the `Drawer` instance on line 225 to match these changes?

src/components/Drawer.js
cerinman: yep, good idea.  I'll update

docs/components/DrawerDocs.js
cerinman: yes, yes I do.  I forgot about that. 👍

src/components/Drawer.js
```diff
@@ -238,6 +243,7 @@ class Drawer extends React.Component {
     } else if (navConfig) {
       menu = this._renderNav(navConfig, styles, theme);
     }
+    const titleUniqueId = _uniqueId('mx-drawer-titile-');
```
jtanner: Nitpick: `titile` -> `title`

docs/components/DrawerDocs.js
```diff
@@ -223,7 +236,8 @@ class DrawerDocs extends React.Component {
       onClose={this._handleDrawerClose}
       title='Demo'
     >
-      // Content Here
+      <p id='description'>This is a demo drawer</p>
+      <div>// Content Here</div>
```
jtanner: Can we just get rid of this line? I don't think it adds much.

PR #737, opened 2018-04-19
The primaryColor prop should take precedence

jtanner:
Even though the `primaryColor` prop is deprecated it should still take precedence. This fixes a bug with f6eff0179d6fda8e7c885a3c81c839d76a26d15e that was not noticed until the `ThemeProvider` was introduced.


cerinman:
💯 for the tests!

cerinman:
:shipit:

PR #738, opened 2018-04-25
SimpleSelect - Wrap in Focus Trap

cerinman:
- Wraps the SimpleSelect component in the MXFocusTrap to prevent tabbing off of the options.
- Adds focusTrapOptions prop to SimpleSelect to allow managing the focus trap.
- Updated the SimpleSelect docs.
- Cleaned up deprecation warnings in componentDidMount.

Tested internally and found no issue in the areas where we use the SimpleSelect.  I also verified that we were not doing anything crazy with refs now that we're wrapping in the focus trap.  All tests are also passing internally. 👍 

screen shot 2018-04-25 at 2 44 05 pm

### DIFF
--- a/docs/components/SimpleSelectDocs.js
+++ b/docs/components/SimpleSelectDocs.js
@@ -55,6 +55,12 @@ class SimpleSelectDocs extends React.Component {
         </div>
 
         <h3>Usage</h3>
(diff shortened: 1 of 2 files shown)


redperk:
:shipit:

PR #739, opened 2018-05-01
Accessibility Tweaks for Calendar, DateRangePicker, and Drawer

cerinman:
Turns out that iOS and Android voice over can care less about aria labels unless you use appropriate roles and/or appropriate semantic HTML like an h tag.

This fixes that issue by adding roles to the various items in the `Calendar`, `DateRangePicker`, and `Selector` components. 

I also... 

- tweaked a couple of aria labels
- hid the calendar day of week column headers from screen readers so that the mobile voice over apps wouldn't force a user to cycle through them before getting to the dates in the calendar. Don't worry, the dates call out their day of week.
- Switched to focusing the close button on mount for the Drawer. Focusing the entire content was problematic as it would look like we were trying to focus something off screen and/or would leave the thing that opened the drawer focused.
- Adds a new `AccessibilityStyles` object to the `Style` constants with an `VISUALLY_HIDDEN` style that we can use via theme to hide things from visual users but still allow the screen reader to read them.

### DIFF
--- a/src/components/DateRangePicker.js
+++ b/src/components/DateRangePicker.js
@@ -311,6 +311,7 @@ class DateRangePicker extends React.Component {
           onClick={this._toggleSelectionPane}
           onKeyDown={(e) => keycode(e) === 'enter' && this._toggleSelectionPane()}
           ref={this.props.elementRef}
+          role='button'
           style={styles.selectedDateWrapper}
           tabIndex={0}
         >
--- a/src/components/DateRangePicker.js
+++ b/src/components/DateRangePicker.js
@@ -402,7 +402,7 @@ class DateRangePicker extends React.Component {
                               ].map(day => {
                                 return (
                                   <div
-                                    aria-label={`Column Header for ${day.value}`}
+                                    aria-hidden={true}
                                     key={day.value}
                                     role='option'
                                     style={styles.calendarWeekDay}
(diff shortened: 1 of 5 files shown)


lukeschunk:
huh interesting! TIL

src/constants/Style.js
```diff
@@ -1,4 +1,32 @@
 const Style = {
+  AccessibilityStyles: {
+
+    /**
+     * Recommended way to hide elements visually but
+     * still allow screen readers to read them.
+     * Verified working in Chrome, Firefox, Safari, IE 11, and Edge
+     *
+     * Used for H1 Budgets title
```
jtanner: Did you mean to keep this line in here?

src/components/DateRangePicker.js
```diff
@@ -401,8 +402,9 @@ class DateRangePicker extends React.Component {
                               ].map(day => {
                                 return (
                                   <div
-                                    aria-label={day.value}
+                                    aria-hidden={true}
                                     key={day.value}
+                                    role='option'
```
jtanner: Does this need a role if `aria-hidden` is true?

src/constants/Style.js
cerinman: haha nope, let me remove that quick

PR #740, opened 2018-05-01
Fix focus issue caused by close button not being present

cerinman:
I introduced a bug in https://github.com/mxenabled/mx-react-components/pull/739 where trying to focus the drawer's close button when the prop `showCloseButton` was set to false would cause an exception 💥

This fixes that issue....


phantomxc:
:shipit:

PR #741, opened 2018-05-08
Reduces Drawer title from an h1 to an h2

cerinman:
I wrongly assumed that the drawer title should be an h1.  For accessibility reasons it actually needs to be an h2.  This PR corrects that assumption.


redperk:
:shipit:

PR #744, opened 2018-05-23
Replace FocusTrap library with internal component and utils

cerinman:
### Issue

The `FocusTrap` library that we currently use to restrict focus to modal and drawer content does not work as expected on mobile devices with mobile voice over apps.  The main issue being that mobile voice over apps don't seem to fire touch events when navigating between focusable elements so `FocusTraps` javascript logic fails allowing navigation to items behind popover content.

### Resolution

- Adds a new RestrictFocusToChildren component with utils. This component changes tabindex and aria-hidden attribute values so that mobile voice over apps ignore content behind popover content.
- Handles scenarios where popover content can be nested. eg: drawer within a drawer
- Handles preserving tabindex values on elements that already have a tabindex value
- Handles scenario where DOM nodes are added after mount outside the restricted children via the `focusin` event
- Adds tests for the new component and utils file
- Adds documentation for new component
- Exports the new component and the utils for use internally.
- Removes `FocusTrap` library, all usages of it, and updates docs accordingly.

### Speed

I've done pretty exhaustive testing internally and here in the open source to compare speed differences between `FocusTrap` and this new component.  I haven't benchmarked anything but the extra cost of looking up additional DOM elements does not appear to have any cost that cause a negative impact on components that it is used in.  

I even tested a scenario where number of focusable items on a page was 50 and saw no noticeable hick up in speed and 50 focusable elements I feel is extreme for most scenarios. 

### Screen Shots

There are none as this component renders nothing but a wrapping div and doing a video to show focus being restricted on all components that it is used in seemed a bit much.  Please do feel free to pull this branch down and run locally. 

#### NOTE

This is not a breaking change as the `Drawer`, `Modal`, `SimpleSelect`, and `DateRangePicker` continue to operate as they did before the change.  The only thing that was removed was the `focusTrapProps` prop on each component that helped configure it.


cerinman:
Offline Chat with Joe.  Documenting here for history sake.

This code naively expects content behind a drawer/modal to not change.  This would fail in scenarios where focusable content is added after mount such as a new transaction coming in from a web socket.

We have a couple of ideas for a solution around that scenario so I'll make some changes and then request another review.  Removing the `Needs Review` label for now.

cerinman:
This is now ready for another review @jtanner

src/Index.js
```diff
@@ -67,6 +68,7 @@ module.exports = {
   Styles: require('./constants/Style'),
 
   // App Utils
+  FocusManagement: require('./utils/FocusManagement'),
```
jtanner: Is there a specific reason why you're exposing `FocusManagement`? If not I'd say let's remove this line.

src/components/RestrictFocusToChildren.js
```diff
@@ -0,0 +1,116 @@
...
+
+    this.focusableNodesInParent = [];
+    this.focusableNodesInWrapper = [];
+    this.preservedTabIndexValues = [];
+    this.previousFocusedNode = document.activeElement;
+  }
+
+  componentDidMount () {
+    this.focusableNodesInWrapper = getFocusableNodesInElement(this._wrapper);
+    this.focusableNodesInParent = reconcileNodeArrays(
```
jtanner: This isn't needed anymore since you added the `focusin` event handler, right?

src/Index.js
cerinman: I had thought I would need it internally but now that we have changed things I don't think I will.  I'll remove it.

src/components/RestrictFocusToChildren.js
cerinman: I waffled back and forth on weather or not to leave this now that we have the focusin listener.  Now that I've thought about it over night I think I will remove it as it feels redundant to aria hide and change tab indexes now.

PR #749, opened 2018-06-08
Changed className in drawer component for use in Cypress

daveguymon:
Made this change for Cypress testing. When trying to navigate out of the drawers that open when creating a sub budget, there was no easy way to distinguish between the elements in each drawer. As far as I can tell, the interpolation of this className is already mx-prefixed.


phantomxc:
This is going to create a dynamically generated class name every time.

daveguymon:
I believe I no longer need this particular change to solve the Cypress issue I ran into. So I am closing this MR.

PR #750, opened 2018-06-13
Drawer Component - Close Animation and Close Button Aria Label

cerinman:
We have a need internally to prevent the close animation from happening where we want to use the back button to update content in the drawer.  This PR does that by adding two new props.

- `animateOnClose` (Default `true`) `Bool`
- `closeButtonAriaLabel` `String`

### DIFF
--- a/src/components/Drawer.js
+++ b/src/components/Drawer.js
@@ -67,6 +67,7 @@ class Drawer extends React.Component {
   };
 
   static defaultProps = {
+    animateOnClose: true,
     beforeClose: () => {},
     closeOnScrimClick: true,
     duration: 500,
--- a/src/components/Drawer.js
+++ b/src/components/Drawer.js
@@ -23,6 +23,7 @@ class Drawer extends React.Component {
   static propTypes = {
     'aria-describedby': PropTypes.string,
     'aria-labelledby': PropTypes.string,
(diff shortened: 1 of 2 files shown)


src/components/Drawer.js
```diff
@@ -167,10 +170,20 @@ class Drawer extends React.Component {
...
-      });
+
+    if (this.props.animateOnClose) {
+      return this._animateComponent({ left: '100%' })
+        .then(() => {
+          this.props.onClose();
+        });
+    } else {
+      // To keep close's api normalized we return a promise just
+      // as the _animateComponent function does above.
```
sambev: Curious. Who consumes the `close` method's api? Are consumers supposed to call this method directly?

cerinman: I can't say if anyone outside MX does but we do internally some times access the close method directly with a ref so I wanted to make sure it behaved the same for us in those cases.

cerinman: actually that is false statement, we don't use a ref anymore but we do pass close as an argument to children of the drawer if they are a function(function as child)

cerinman: example...

```javascript
<Drawer>
  {({ close }) => {
    <button onClick(close) />
  }}
</Drawer>
```

sambev:
:shipit:

PR #751, opened 2018-06-14
Fix date range picker button style

cerinman:
I just recently changed the selected date in the DateRangePicker from a link to a button and in doing so, I jacked up its styling.  This PR fixes that mistake to restore how it used to look.

### Broken

screen shot 2018-06-14 at 10 53 21 am

### Fixed

screen shot 2018-06-14 at 10 52 29 am


sambev:
:shipit:

PR #752, opened 2018-06-25
Date Range Picker a11y fix

cerinman:
DateRangePicker - a11y fix - Default Ranges not being read well in JAWS

- Add aria-label to give more details
- Cleans up JSX to use actual HTML button to simplify things

#### No Harm

screen shot 2018-06-25 at 2 15 48 pm


jmophoto:
:shipit:

PR #753, opened 2018-06-25
Calendar hidden on blur

jmophoto:
There is an issue with the Date Time Picker where the calendar was hidden whenever the containing div was blurred, making the scroll through months buttons useless. 

This changes the hide/show logic to a click, similar to the DatePicker pattern. It also hides the calendar when the time it focused.


cerinman:
:shipit:

PR #754, opened 2018-07-04
Mbd/update message box

marabd:
this is currently a work in progress. the styling is nearly there, but there's a possibility that we'll need to make a separate component, or at least move the message box text and button off props, and into the message box component. in the event that it needs to be worked on before I'm back on the 23rd, someone can pull this down.


marabd:
styling is finally done on this, and it is ready for review.

before:

screen shot 2018-08-09 at 1 36 57 pm

after:

screen shot 2018-08-09 at 1 35 57 pm

docs/css/reset.css
```diff
@@ -4,7 +4,7 @@ html;body;div;span;applet;object;iframe;h1;h2;h3;h4;h5,h6,p,blockquote,pre,a,abb
 body, html {
   height: 100%;
   -webkit-font-smoothing: antialiased;
-  font-family: "Lato", "Helvetica Neue", "Helvetica", Helvetica, Arial, sans-serif;;
+  font-family: "Helvetica Neue", "Helvetica", Helvetica, Arial, sans-serif;;
```
cerinman: Was this a request by Derek?

src/components/MessageBox.js
```diff
@@ -11,55 +12,51 @@ const StyleUtils = require('../utils/Style');
...
-              type={this.props.icon}
-            />
-            <div style={styles.title}>{this.props.title}</div>
+        <div style={styles.alertbar} />
+        <div style={styles.messageSection}>
+          <div style={styles.messageBody}>
+            <div>
+              <Icon
+                size={20}
+                style={Object.assign({}, styles.icon, { marginRight: theme.Spacing.SMALL })}
```
cerinman: I know this is already existing code but since you are here can we kill `Object.assign` in favor of object spread syntax?

src/components/MessageBox.js
```diff
@@ -69,31 +66,56 @@ class MessageBox extends React.Component {
...
+      },
+      messageContent: {
+        boxSizing: this.props.isSmall ? null : 'content-box',
+        textAlign: 'left',
+        width: this.props.isSmall ? '100%' : '75%',
+        paddingRight: this.props.isSmall ? 30 : null,
         fontSize: theme.FontSizes.MEDIUM
       },
-      icon: {
-        fill: theme.Colors.WHITE
```
cerinman: It looks like you removed this `icon` style but are still using it above for the `Icon` component.  Based upon your screen shots, it looks like the icons are defaulting to black which also points to this.  Can we add this back with the appropriate fill color?

docs/css/reset.css
marabd: yes, this was a change Derek wanted made

src/components/MessageBox.js
marabd: yep, I can absolutely do that

src/components/MessageBox.js
marabd: actually, we no longer want the icon fill to adjust to the color of the status anymore. we want them to remain solid Gray900 with a white icon, or solid white fill with Gray900 as the color

cerinman: But you're still applying `styles.icon` to the Icon component on line 37 but not providing the icon key here in the Styles object.  

This probably works internally because we're passing the icon key in the Styles prop object that we pass but here in the OS docs there is no icon style being applied so the fill of the icon becomes black, not `GRAY_900`.

marabd: thanks for clarifying that for me. I updated it so that the icon has `fill: GRAY_900`. does that look better?

also, I noticed two my keys I accidentally left in the docs, so I removed those too

src/components/MessageBox.js
```diff
@@ -11,55 +12,51 @@ const StyleUtils = require('../utils/Style');
...
-              type={this.props.icon}
-            />
-            <div style={styles.title}>{this.props.title}</div>
+        <div style={styles.alertbar} />
+        <div style={styles.messageSection}>
+          <div style={styles.messageBody}>
+            <div>
+              <Icon
+                size={20}
+                style={{ ...styles.icon, ...{ marginRight: theme.Spacing.SMALL } }}
```
cerinman: Hmm, now that I look at this closer, there is no conditional for applying `marginRight` here. It would probably be better if we move `marginRight` inside the `styles.icon` object directly to simplify things..

```javascript
      style={styles.icon}

      icon: {
        fill: theme.Colors.GRAY_900,
        marginRight: theme.Spacing.SMALL
      },

File: src/components/MessageBox.js [hunk 5]
```diff
@@ -11,55 +12,51 @@ const StyleUtils = require('../utils/Style');
...
+            </div>
+            <div style={styles.messageContent}>
+              <div style={styles.title}>
+                {this.props.title}
+              </div>
+              <div style={styles.mainMessage}>
+                {!_isNil(this.props.message) &&
+                  <div>
+                    {this.props.message}
+                  </div>
```
cerinman: Can we remove this wrapping div from around `message`?

src/components/MessageBox.js
```diff
@@ -11,55 +12,51 @@ const StyleUtils = require('../utils/Style');
...
       <div className='mx-message-box' style={styles.component}>
-        <div onClick={this.props.children ? this._toggleMessageBox : () => {}} style={styles.header}>
-          <div style={styles.leftHeader}>
-            <Icon
-              size={20}
-              style={Object.assign({}, styles.icon, { marginRight: theme.Spacing.SMALL })}
-              type={this.props.icon}
-            />
-            <div style={styles.title}>{this.props.title}</div>
+        <div style={styles.alertbar} />
```
cerinman: I actually think we should try and replace all these divs with semantic HTML elements for a11y reasons.  We should also probably try and make the title a header but we should think through the level required there.

src/components/MessageBox.js
marabd: good call, I just pushed that change

src/components/MessageBox.js
marabd: I tried that, but it needs a div for it to be a JSX element

src/components/MessageBox.js
```diff
@@ -11,56 +12,55 @@ const StyleUtils = require('../utils/Style');
...
+                style={styles.icon}
+                type={this.props.icon}
+              />
+            </section>
+            <section style={styles.messageContent}>
+              <p style={styles.title}>
+                {this.props.title}
+              </p>
+              { this.props.message ?
+                (<p role='status' style={styles.message}>
```
cerinman: 😢 sorry to comment again....

I think we better move the `role='status'` here back up to the `messageContent` section and verify that voice over still works.

marabd: I should have caught this myself, so no worries. I updated this, and tested.

cerinman:
:shipit:

PR #755, opened 2018-07-10
Add a styles prop to Button that can override any style in the button

sambev:
I didn't move `style` prop around since that would be a major breaking change to the button's api. That prop only gets merged into the `component` style key.

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -34,6 +34,7 @@ class Button extends React.Component {
     elementProps: {},
     onClick () {},
     isActive: false,
+    styles: {},
     type: 'primary'
   };
 
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -25,6 +25,7 @@ class Button extends React.Component {
     onClick: PropTypes.func,
     primaryColor: PropTypes.string,
     style: PropTypes.object,
+    styles: PropTypes.object,
     theme: themeShape,
     type: PropTypes.oneOf(buttonTypes)
   };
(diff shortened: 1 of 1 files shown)


src/components/Button.js
```diff
@@ -101,7 +105,7 @@ class Button extends React.Component {
     const windowSizeIsSmall = this._windowSizeIsSmall(theme);
 
     return {
-      component: Object.assign({
+      component: {
```
cerinman: Thanks for killing Object.assign here.

src/components/Button.js
```diff
@@ -68,6 +69,9 @@ class Button extends React.Component {
     const mergedTheme = StyleUtils.mergeTheme(theme, primaryColor);
     const styles = this.styles(mergedTheme);
 
+    // We need to remove the styles prop from rest so we don't pass it to children.
+    delete rest.styles;
```
cerinman: Nice, I didn't think `delete` was fully supported in all browsers but it is.  👍 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete

src/components/Button.js
```diff
@@ -245,7 +250,8 @@ class Button extends React.Component {
       },
       buttonText: {
         marginLeft: (this.props.isActive && this.props.actionText) ? 10 : 0
-      }
+      },
+      ...this.props.styles
```
cerinman: Should the the styles prop have a default value of `{}`? Or does spread do fine if a value is null/undefined?

sambev: I think it works okay:
screen shot 2018-07-11 at 1 34 28 pm

sambev: This now has a default of `{}`

cerinman:
:shipit:

PR #756, opened 2018-07-11
remove event from window and put it into top level div

hokakede1:


src/components/SimpleSelect.js
```diff
@@ -44,15 +44,9 @@ class SimpleSelect extends React.Component {
...
-
-    window.addEventListener('keydown', this._handleKeyDown);
-  }
-
-  componentWillUnmount () {
-    window.removeEventListener('keydown', this._handleKeyDown);
   }
 
-  _handleKeyDown = (e) => {
+  _handleKeyUp = (e) => {
```
paniclater: What's the need to change from keyDown to keyUp?

src/components/SimpleSelect.js
```diff
@@ -44,15 +44,9 @@ class SimpleSelect extends React.Component {
     deprecatePrimaryColor(this.props, 'hoverColor');
     deprecateProp(this.props, 'iconStyles', 'styles', 'simple-select');
     deprecateProp(this.props, 'menuStyles', 'styles', 'simple-select');
-
-    window.addEventListener('keydown', this._handleKeyDown);
```
paniclater: 😨 Oh my.... 😨

src/components/SimpleSelect.js
hokakede1: Naming convention, since the method now binding to `onKeyUp` listener, it will cause confusions if I left it as `handleKeyDown`.

paniclater:
:shipit:

cerinman:
:shipit:

PR #757, opened 2018-07-11
added Glamour and remove Radium in Listbox

hokakede1:
it seems like this is all I have to do, but is it ? Am I missing anything ?

### DIFF
--- a/src/components/accessibility/Listbox.js
+++ b/src/components/accessibility/Listbox.js
@@ -106,10 +106,9 @@ class Listbox extends React.Component {
     return (
       <div
         aria-label={this.props['aria-label']}
-        className='mx-listbox'
+        className={'mx-listbox' + css({ ...this.props.style })}
         ref={ref => this.component = ref}
         role='listbox'
-        style={css(...this.props.style)}
       >
         {React.Children.map(this.props.children, (child, index) =>
           React.cloneElement(child, {
--- a/src/components/accessibility/Listbox.js
+++ b/src/components/accessibility/Listbox.js
@@ -106,7 +106,7 @@ class Listbox extends React.Component {
     return (
       <div
         aria-label={this.props['aria-label']}
-        className={'mx-listbox' + css({ ...this.props.style })}
+        className={`mx-listbox ${css(this.props.style)}`}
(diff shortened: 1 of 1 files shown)


paniclater:
I believe that this is all you need to do in most components - however there may be some trickier issues around the media queries for breakpoints that may pop up. Most components should be similar to this though.

paniclater:
@phantomxc @cerinman @sambev  am I correct here?

paniclater:
sorry accidentally clicked close 😳

sambev:
I think this is the general idea, but we do want to be taking a close look at differences, sometimes it isn't always a pure 1:1.

sambev:
Actually I think you need to change it from `style` to `className`

hokakede1:
@sambev Oh my god, thanks Sam. I was stuck for 30 minutes wondering why the styles are jacked up. 🤦‍♂️

cerinman:
Yes, 1:1 for most cases but as an example, the Drawer component uses break points so we'll have to make sure that is covered.

I would also suggest looking for merged styles.  In the case of radium, there where places where we used the radium syntax that will have to be updated to object spread.

Radium syntax
```javascript
<div style={[styles.item, isSelected && styles.selectedItem]}>
```

Object spread
```javascript
<div style={{ ...styles.item, ...isSelected ? styles.selectedItem : {} }}
```

src/components/accessibility/Listbox.js
```diff
@@ -106,10 +106,9 @@ class Listbox extends React.Component {
     return (
       <div
         aria-label={this.props['aria-label']}
-        className='mx-listbox'
+        className={'mx-listbox' + css({ ...this.props.style })}
```
cerinman: You shouldn't have to spread `this.props.style` in a new object here.  You should be able to just do.

```javascript
className={'mx-listbox' + css(this.props.style)}
```

cerinman: You might even be able to use string interpolation as well

```javascript
className={`mx-listbox ${css(this.props.style)}`}
```

cerinman:

PR #758, opened 2018-07-11
DateRangePicker - a11y Fixes and clean up

cerinman:
### Problems

1. Using the keyboard to select a date on the calendar at mobile resolutions did not close the calendar like clicking on a date does.
2. Using the enter key to open the DateRangePicker at mobile resolutions would immediately select the `FROM` link and open the calendar. 
3. After selecting a date at mobile resolutions, focus was lost as we were not returning focus to the `From` or `To` button that was used to open the calendar
4. The logic for the `_handleDateSelect` function was convoluted and had a lot of setState calls. We were also setting state in one case in a way that was problematic due to the async nature of setState.
5. The docs were not on the same version of React as the components.  This caused some weird stuff with setState callbacks that was hard to trouble shot.
6. Updating the docs React version to 16 broke everything due to using React Router v 2.
7. The logic in the `MonthTable` component for focusing days was confusing and bug prone.
8. Aria labels don't call out which end of the range you are selecting.
9. Initial values for the `currentDate` and `focusedDay` state properties in `DateRangePicker` are not set to start of day to match days in the calendar causing weird initial behavior.

### Solutions

1. Pass along the selected date and event to key event so we can pass the right thing and prevent default/stop propagation so that the calendar closes when using enter to select a date.
2. Convert `From` and `To` links into buttons so that button key press event from opening the DateRangePicker does not overlap with selecting the `From` or `To`. Also corrected the styling due to changing to buttons.
3. Use refs for `From` and `To` buttons to return focus to the correct button after selecting a date.
4. Clean up code and consolidate setState calls. Also switches to using a function as the first arg to setState so that using existing state to update a state property doesn't become problematic.
5. Update the docs dependencies to be in line with the components
6. Update React Router to v 3(going all the way to 4 would require more work than I wanted to deal with at the moment)
7. Refactored the weird ref callback logic for focusing days on the calendar to using straight componentDidMount/componentDidUpdate lifecycle methods
8. Improved aria-labels to indicate which end of the range you are currently selecting.
9. Init values for the `currentDate` and `focusedDay` state properties in `DateRangePicker` to start of day.

### No Harm screen shots

screen shot 2018-07-11 at 4 22 36 pm
screen shot 2018-07-11 at 4 22 46 pm
screen shot 2018-07-11 at 4 23 10 pm
screen shot 2018-07-11 at 4 23 28 pm
screen shot 2018-07-11 at 4 23 46 pm

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -5,8 +5,8 @@
   "author": "MX",
   "license": "MIT",
   "devDependencies": {
-    "babel-cli": "^6.3.17",
-    "babel-core": "^6.3.21",
+    "babel-cli": "^6.6.5",
+    "babel-core": "^6.0.0",
     "babel-eslint": "7.2.1",
     "babel-loader": "^6.0.0",
     "babel-plugin-transform-object-assign": "^6.5.0",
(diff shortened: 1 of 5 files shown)


jmophoto:
🍪 :shipit:

jmophoto:
:shipit:

PR #759, opened 2018-07-12
Apply glamour and remove radium to Button Group

hokakede1:
- So I move only `styles.disabled` into glamor, the rest of it I use regular in-line style. 

- For some reasons, if I put it all into glamor, the style will be overwritten.


cerinman:

PR #762, opened 2018-07-12
remove radium in DatePicker

hokakede1:
- remove radium and use regular inline style.


cerinman:

PR #764, opened 2018-07-17
Moore/updating simple input in search input

paniclater:
Updates search input to use prefix and suffix instead of deprecated left and right icon biz.

- deals with search input reset
- deals with focus on search icon

Its been a while since I've done anything of consequence here so please review carefully!


jmophoto:
:shipit:

cerinman:
Double :shipit:

PR #767, opened 2018-07-26
Focus Trap Fix via React portals

cerinman:
Fixes FocusTrap issue in Android Talkback and iOS voice over where a user could focus content behind a modal/drawer due to DOM events not firing.

- Uses React Portals to render MXFocusTrap as a sibling next to a selected DOM element via a query string. Renders in place if no query string or query string results in no DOM element found.
- Manages aria-hidden values for traps and sibling of traps so that content behind modals/drawers is hidden properly when portaling.
- Adds tests for new functionality in MXFocusTrap wrapper.(Attempted testing via DOM but ran into a lot of problems so scaled back to just check the traps for aria hidden attributes for now.)

I've spent a lot of time testing this internally in several widgets for modals and drawers and have found no issues.

### Drawer

Portal to App Wrapper

screen shot 2018-07-28 at 4 05 46 pm
screen shot 2018-07-28 at 4 06 07 pm

### Modal

Portal to App wrapper

screen shot 2018-07-28 at 4 10 53 pm

### SimpleSelect

Rendered in place

screen shot 2018-07-28 at 4 11 58 pm


jmophoto:
:shipit:

jmophoto:
Lot of work and research. Many 🍪

PR #769, opened 2018-08-02
Focus Trap - Manage zIndex level of traps

cerinman:
When we use React Portals to render Focus Trapped content(Drawers, Modals) the content is appended to the DOM location chosen.  This causes issues where Multiple Drawers/Modals are rendered at once since all of the Drawer/Modals have the same z index value.

This PR solves the issue by passing a `trapNumber` to the children of the `MXFocusTrap` component so that they can use it to appropriately set zIndex values.  I also added extra tests for the `trapNumber` being passed to children.

FYI: I researched React Portals to see if we could prepend instead of append but it doesn't currently take any additional options to do so. :(

screen shot 2018-08-02 at 12 13 56 pm
screen shot 2018-08-02 at 12 14 38 pm
screen shot 2018-08-02 at 12 15 02 pm

### DIFF
--- a/src/components/Drawer.js
+++ b/src/components/Drawer.js
@@ -238,92 +238,97 @@ class Drawer extends React.Component {
   };
 
   render () {
-    const { theme } = this.state;
-    const styles = this.styles(theme);
-    const { closeButtonAriaLabel, headerMenu, focusTrapProps, navConfig, portalTo } = this.props;
     const mergedFocusTrapProps = {
       focusTrapOptions: {
         clickOutsideDeactivates: true,
-        portalTo
+        portalTo: this.props.portalTo
       },
       paused: false,
-      ...focusTrapProps
+      ...this.props.focusTrapProps
     };
-
-    let menu = null;
-
-    // If headerMenu is a function then we want to pass the Drawer's
-    // exposed functions to the call.
-    if (typeof headerMenu === 'function') {
-      menu = headerMenu(this._getExposedDrawerFunctions());
-    // If headerMenu is a normal node/element then use directly.
-    } else if (headerMenu) {
-      menu = headerMenu;
-    // If no headerMenu and navConfig passed then use Drawer's
-    // _renderNav function to generate the menu.
-    } else if (navConfig) {
-      menu = this._renderNav(navConfig, styles, theme);
-    }
-    const titleUniqueId = _uniqueId('mx-drawer-title-');
 
     return (
       <StyleRoot>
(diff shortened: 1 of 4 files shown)


src/components/Drawer.js
```diff
@@ -238,92 +238,97 @@ class Drawer extends React.Component {
...
-                      type={'base'}
-                    />
-                  }
-                </span>
-                <h2 id={titleUniqueId} style={styles.title}>
-                  {this.props.title}
-                </h2>
-                <div className='mx-drawer-header-menu' style={styles.headerMenu}>
-                  {menu}
+          {trapNumber => {
```
cerinman: The method for passing the trapNumber is to make the children a function.  This is the same in Drawer and Modal

src/components/Drawer.js
```diff
@@ -238,92 +238,97 @@ class Drawer extends React.Component {
...
-  styles = (theme) => {
+  styles = (theme, trapNumber = 1) => {
     const HEADER_HEIGHT = this._getHeaderHeight();
 
     return _merge({}, {
       component: {
         border: '1px solid ' + theme.Colors.GRAY_300,
         boxSizing: 'border-box',
-        zIndex: 1001,
+        zIndex: trapNumber * 1002,
```
sambev: I am not sure multiplication is what we want here. Do you want it to increment by the thousands for each number?

cerinman: If I do addition then I have to know the zIndex of the first trap rendered to make sure we don't have overlap with numbers.  This is due to the fact that the wrapping div, scrim, and content all have individual z index values because of how they are positioned.

Multiplication solves that problem as the zIndex value will always be unique.

sambev: I am not sure I understand how multiplication ensures the zIndex value is always unique. Can you give me an example of the difference?

src/components/MXFocusTrap.js
```diff
@@ -83,7 +85,7 @@ class MXFocusTrap extends React.Component {
       return ReactDOM.createPortal(
         (
           <FocusTrap {...this.props} paused={this.state.paused}>
-            {this.props.children}
+            {typeof this.props.children === 'function' ? this.props.children(this._trapNumber) : this.props.children}
```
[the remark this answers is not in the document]
sambev: Curious, Do you know if this breaks stateless functional components?

sambev:
z-index management is a pretty hard thing. Do we have any existing conventions around them?

cerinman:
Closing this.  It is a bull shit idea with no way of succeeding.  I hate my life.

PR #772, opened 2018-08-06
Focus Trap BS again

cerinman:
Reverts the revert that ended up not being needed because I'm just an idiot..... 
 https://github.com/mxenabled/mx-react-components/pull/770

There is nothing different here in this PR compared to the original here https://github.com/mxenabled/mx-react-components/pull/767


phantomxc:
:shipit:

PR #775, opened 2018-08-07
Add new waffle icon to docs

cerinman:
New waffle icon wasn't showing up in the docs because we neglected to add it to the app constants.


jmophoto:
:shipit:

PR #776, opened 2018-08-09
Use a button instead of div on the Select.

sambev:
A button is a more appropriate element and doesn't require all of
the custom click, keydown, focus state, or tabIndex.

Update the style to match the div look.

### DIFF
--- a/src/components/Select.js
+++ b/src/components/Select.js
@@ -196,15 +196,13 @@ class Select extends React.Component {
 
     return (
       <div className='mx-select' style={Object.assign({}, this.props.style, { position: 'relative' })}>
-        <div className='mx-select-custom'
+        <button className='mx-select-custom'
           onClick={haltEvent(this._open)}
-          onKeyDown={this._handleKeyDown}
           ref={ref => {
             this.elementRef = ref;
             if (typeof this.props.elementRef === 'function') this.props.elementRef(ref);
           }}
           style={styles.component}
-          tabIndex='0'
         >
           {this._renderScrim(styles)}
           <div className='mx-select-selected' key='selected' style={styles.selected}>
@@ -222,7 +220,7 @@ class Select extends React.Component {
             />
           </div>
           {this.props.options.length || this.props.children ? this._renderOptions(styles) : null}
-        </div>
+        </button>
       </div>
     );
   }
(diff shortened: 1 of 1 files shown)


src/components/Select.js
```diff
@@ -245,7 +243,9 @@ class Select extends React.Component {
           fontSize: theme.FontSizes.MEDIUM,
           padding: '8px 10px',
           position: 'relative',
-          boxSizing: 'border-box'
+          boxSizing: 'border-box',
+          width: '100%',
+          textAlign: 'left'
```
cerinman: Is there any need to update border styling as well?

cerinman:
Screen shot from the docs showing changes?

src/components/Select.js
cerinman: Scratch this question.  I expanded the code and realized there is already border styling in place.

src/components/Select.js
```diff
@@ -196,15 +196,13 @@ class Select extends React.Component {
 
     return (
       <div className='mx-select' style={Object.assign({}, this.props.style, { position: 'relative' })}>
-        <div className='mx-select-custom'
+        <button className='mx-select-custom'
           onClick={haltEvent(this._open)}
-          onKeyDown={this._handleKeyDown}
```
cerinman: It looks like `_handleKeyDown` was checking for the `esc` key to close the select.  We might still have to provide a key event handler to preserve that functionality.  We can also remove the `_handleKeyDown` function as it wasn't being used anywhere else.

sambev: hmmmm. Good catch. Maybe I just go through the effort of making this div work with roles and tabIndex.

sambev: @cerinman any downsides to just adding the keydown back to the button?

cerinman: nope, it is a valid thing to do if you have to handle something outside of matching the click handler.

sambev:
@cerinman added the handler back.

screen shot 2018-08-14 at 9 53 10 am
screen shot 2018-08-14 at 9 53 15 am

cerinman:
👍 ❤️

cerinman:
:shipit:

cerinman:
Queued for 5.2.25 release

PR #777, opened 2018-08-09
Icon - background fill

cerinman:
### Problem

We have a situation internally where content behind an icon shows through but we can't necessarily change the position of the icon.  We've also have valid design use cases for certain icons having separate background fill colors.

### Solution

This MR adds more path data to a handful of icons and provides a new prop `backgroundFillColor` so that we can effect it's color separately from the main icon.

### Things I'm not happy with

The mechanism for changing the fill of the main icon is via the style prop.  The mechanism for styling the background path is a separate prop.  I tried reconciling everything into the style prop for consistency but that required using a key name that is not a valid css property and after talking to Sam we both felt that it was better to be explicit with a new prop rather than doing something funky with the style prop.

#### Background Fill Demo

backgroundfill

#### Icon Fill Demo

iconfill

#### Icon Size Demo

iconsize


lukeschunk:
oooh i dig this

phantomxc:
:shipit:

cerinman:
Queued for release in 5.2.25

phantomxc:
:shipit:

PR #779, opened 2018-08-13
DisplayInput disabled fix

cerinman:
Before, we would hide the input via the `aria-hidden` attribute if `disabled: true` was passed int he `elementProps` prop.  This was not meeting a11y standards as visually impaired users became completely unaware of the field rather than being notified of it but that is was disabled.

This PR fixes the issue by desctructuring `disabled` and `onChange` from the `elementProps` prop and applying them correctly to the input.

I also updated the docs with an example of a disabled DisplayInput.

### Screen shot

Notice the `aria-disabled` attribute on the input

screen shot 2018-08-13 at 2 33 18 pm

### DIFF
--- a/docs/components/DisplayInputDocs.js
+++ b/docs/components/DisplayInputDocs.js
@@ -68,6 +68,18 @@ class DisplayInputDocs extends React.Component {
           valid={true}
         />
         <DisplayInput
+          elementProps={{
+            disabled: true,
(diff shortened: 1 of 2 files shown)


jmophoto:
:shipit:

cerinman:
Queued for the 5.2.25 release

Issue #778, opened 2018-08-10
mx-react-components version 6 release


cerinman:
We are nearing the release of version 6 of `mx-react-components` due to some needed breaking changes in the `MessageBox` component.  I'm opening this issue to outline what I would like to get done for this major release to keep us all on the same page.

cc: @marabd @sambev @paniclater @phantomxc @jmophoto @hokakede1 

- [x] Updated `MessageBox` component https://github.com/mxenabled/mx-react-components/pull/754
- [x]  Remove deprecated icons  https://github.com/mxenabled/mx-react-components/pull/782
- [x] Remove `DatePickerFullScreen` component. It was to solve a problem we don't have any more and it is behind the curve on styling and maintenance.  https://github.com/mxenabled/mx-react-components/pull/780
- [x] Remove `RajaIcon` component. It is no longer used and we have no documentation for it anyway. https://github.com/mxenabled/mx-react-components/pull/781
- [x] `SearchInput` needs to be updated to use the correct `SimpleInput` props. https://github.com/mxenabled/mx-react-components/pull/790
- [x] There are several components that have props that have been deprecated.  We need to remove those deprecated props as part of this release.  A simple search in the src directory for `deprecate` will find most of them.
https://github.com/mxenabled/mx-react-components/pull/783
https://github.com/mxenabled/mx-react-components/pull/784
https://github.com/mxenabled/mx-react-components/pull/785
https://github.com/mxenabled/mx-react-components/pull/786
https://github.com/mxenabled/mx-react-components/pull/787
https://github.com/mxenabled/mx-react-components/pull/788
https://github.com/mxenabled/mx-react-components/pull/789
- [x] DateRangePicker footer styling fix https://github.com/mxenabled/mx-react-components/pull/795
- [x] Search for `TODO` pertaining to deprecated items and address as needed. All TODO comments where removed in the above deprecated props clean up PRs
- [x] Do an audit of the docs and update/add documentation where appropriate. (Mara)
~Auto Change log/Docs based upon squashed commit messages.~
~New branching structure. Master/Stable with updated README info on development/releasing.~

Open to suggestions/feed back here.

daveguymon:
Verified resolved. Closing issue.

PR #780, opened 2018-08-14
Removes DatePickerFullScreen

cerinman:
https://github.com/mxenabled/mx-react-components/issues/778

As part of the 6.0 release we want to clean some stuff up.  This Deletes the old unsupported component `DatePickerFullScreen`.  It is not used and far behind the curve for being up to date/maintained.


jmophoto:
:shipit:

PR #781, opened 2018-08-14
Remove RajaIcon component

cerinman:
https://github.com/mxenabled/mx-react-components/issues/778

Removes old unsupported icons that we no longer use or want to maintain.


jmophoto:
:shipit:

PR #791, opened 2018-08-15
Fix Tabs docs

cerinman:
We were getting a theme error when trying to view the docs for the Tabs component because we missed updating the `PillTabs` component to merge theme when we updated a while back.  This PR fixes that issue.

screen shot 2018-08-15 at 1 22 24 pm


marabd:
looks good!

marabd:
:shipit:

PR #794, opened 2018-08-21
Select - Switch back to a div :(

cerinman:
Firefox and IE 11 don't like nested content in a button to be absolutely positioned apparently. Options in the select are not selectable via mouse in Firefox and the options don't even show up in IE 11.

This PR fixes the issue by switching back to a div but providing a role of `button` to maintain a11y.  Basically reverts https://github.com/mxenabled/mx-react-components/pull/776


sambev:
😢 :shipit:

lukeschunk:
😢

PR #795, opened 2018-08-27
Fix spacing for the date range picker.

phantomxc:
Fixes the spacing for the date range picker. Notice the bottom left corner. 

Before:

image

After:

image


cerinman:
:shipit:

PR #798, opened 2018-09-05
MBD: updated docs with typo fixes

marabd:
this fixes typos found during QA review of the OS 6.0 release


paniclater:
Typos look good, but tests failed?

jmophoto:
Looks like the calendar test is set up assuming every month will show 5 weeks and checks the current month. Now that it is September, the calendar would show six weeks.

jmophoto:
You could hard code a unix timestamp into the `selectedDate` prop in the test.

https://github.com/mxenabled/mx-react-components/blob/master/src/components/__tests__/Calendar-test.js#L25

Change
`selectedDate={moment().unix()}`

To
`selectedDate={'1533124800'}`

That would set it to always be August 1, 2018.

jmophoto:
:shipit:

paniclater:
:shipit:

PR #799, opened 2018-09-06
MBD/updated prop type from string to include node

marabd:
in the message box redesign released in the 6.0.0 release, the message prop was only a string. however, due to certain instances where this component is utilized, it makes more sense to change this prop type to oneOfType node or string


cerinman:
:shipit:

PR #802, opened 2018-09-21
Remove global handler prop to ListBox in SimpleSelect and Select.

sambev:
These global event listeners (particularly space) prevent inputs from allow the space character when the simple select is open but not focused.

I am not really sure why they were applied globally in the first place. @jtanner do you happen to know if this could cause regressions? I am not sure what situation would require the global listener.


cerinman:
:shipit:

cerinman:
:shipit:

PR #805, opened 2018-10-08
fix path to bundle

jmophoto:
The bundle file was being put in `dist/docs/bundle` instead of `docs/bundle`.

This changes that path to the correct location.


webpack.config.js
```diff
@@ -44,7 +44,7 @@ module.exports = {
     noParse: [/autoit.js/]
   },
   output: {
-    filename: './docs/bundle.js'
+    filename: '../docs/bundle.js'
```
cerinman: Do we need to fix this on the entry point as well?

cerinman: Actually, scratch this comment.  I get it now.  Output of webpack is the dist folder so `./` refers to dist.  Entry should be fine the way it is.

webpack.config.js
```diff
@@ -10,9 +12,9 @@ module.exports = {
     port: 8080,
     progress: true
   },
-  devtool: 'eval',
+  devtool: !isProd && 'eval',
```
cerinman: So this was the source of the bundle ballooning in size?

jmophoto: Yes. Adding this dropped it to 2.5MG

cerinman: I think we might still need to add the uglify plugin here.  I think the bundle can be smaller still.

jmophoto: Possibly. That is back down to what the bundle was with uglify.

jmophoto: And it does run Uglify. When you make the production build and watch the progress, it goes through uglify.

cerinman: Yes, but with webpack 4 I would expect better. Everything I've read on it says that people have been decreasing bundle size, not maintaining.

Also, it looks like this might be increasing our bundle size internally so we need to go the extra mile to make sure we're covering our bases and doing everything we can.

jmophoto: I've added the `ContextReplacementPlugin` that removes the moment locales and it dropped to 2MB. 

I agree that it would be ideal to lower the bundle size.

I also think that having an open source project that we use in production that an NPM audit shows having 11 low, 1 moderate, 8 high and 1 critical vulnerabilities is far more of a concern than the bundle size for the development docs. 

Any other suggestions are welcome.

jmophoto:
There were no changes in this MR that touched the packages included, so I am not sure why there are so many changes in the package-lock. Let me try deleting it and re-running just to be sure.

webpack.config.js
sambev: @cerinman The webpack bundle shouldn't be related to the increase in bundle size for babel that I was seeing in the other repo.
 
I totally agree that the audit is a more important issue. But I don't think that we need to treat bundle size and security as mutually exclusive. Before we were ignoring the audit issues, which was wrong to do. 3x-ing our bundle size without knowing why is also a problem we should get to the bottom of. Unfortunately we all made this bed of updating everything at once and we need to sort it out. We can get it figured out!

I'll try to run some tests to see what if I can figure out anything about bundle size. I am not totally convinced that the webpack numbers are what we see in the deployed version, so I am going to confirm that real quick and report back.

jmophoto: Where are you seeing a bundle increase of 3X?

jmophoto: Maybe we are talking about completely different things. I was looking at the size of the bundle.js for the docs.

sambev: The 6.0.1 docs had a bundle size of 677kb:

screen shot 2018-10-09 at 9 38 30 am


If webpack is reporting 2MB, that is quite a disparity, but I want to make sure that they are one in the same before casting verdict on that.

sambev:
:shipit:

PR #808, opened 2018-10-09
Kelliegreen/image upload behavior

kellie-green:
This is for MR: https://gitlab.mx.com/mx/robinhood/merge_requests/1763


cerinman:
Closing this. After discussing with Kellie we've decided that this change is too specific to our needs internally and not valid for the OS component. We'll most likely just create a new FileUpload component internally for our needs and deprecate this one over time.

PR #809, opened 2018-10-09
Export Icon Paths

cerinman:
This moves the icon paths used in the `Icon` component out to a constant file with a helper function for getting the correct path.

I've done this because I have a need to access the path data internally for a new budget icon based upon path data used here in the `Icon` component.

I've had to use function here because we recently added path data to be used with a backgroundFillColor prop so that we can color the background of some icons.

screen shot 2018-10-09 at 1 04 13 pm


jmophoto:
:shipit:

jmophoto:
:shipit:

PR #810, opened 2018-10-09
Sb/bundle sizer

sambev:
Check our bundle sizes when testing.

screen shot 2018-10-09 at 2 42 36 pm


sambev:
I think this is failing because of the package.json consolidation. The CI can't find mx-react-components dep of the docs. Not sure how to fix this but we need it to be built for the bundlesize checker. Works locally though :(

cc @jmophoto @cerinman

sambev:
https://github.com/mxenabled/mx-react-components/pull/813

Issue #811, opened 2018-10-09
Split up package.jsons for docs and lib again


sambev:
Bringing them into one package is convenient for, well having just one, but I wonder if that cost is worth it.

So far there are two issues I think one is causing that I don't know how to fix otherwise:

1. CI bundle size checking (docs bundle vs lib bundle)
2. Minifying the lib build.

Webpack already minify's the doc build, but babel isn't minifying the lib. It probably should for production. Can webpack do both separately?

cc @jmophoto @cerinman  @phantomxc

cerinman:
We can define separate entry and output points in the webpack config file but having one `package.json` file can have unintended side effects when adding dependencies if they are placed in the wrong location in the file. The single webpack config file would also be more complicated. 

I know it is more work to keep two separate `package.json` files in sync but I feel that it is less problematic to separate the concerns and have two files and if we want to minify the lib build then we can handle that with a separate config file as well.

jmophoto:
Being the one that merged the two, I am thinking it would be best to break them apart again. Seeing issues with the hot reload.

PR #812, opened 2018-10-10
Separate Packages

jmophoto:
This separates the package.json files out again and moves the webpack config into the docs directory.

Issue: https://github.com/mxenabled/mx-react-components/issues/811 

This pulls in Sam's bundle size check on the docs bundle.

It also removes the `.npmignore` file. From what I have read, this can lead to issues where `npm` honors this over the `.gitignore` file. This can lead to files ignored in git being added accidentally to the npm package. Since we are specifying the `files: 'dist'` now, those are the only ones included and the ignore isn't needed.

#### .travis.yml
- Add ci and auditing for the package.json in the docs directory

#### Changelog.js
- Update changelog to latest version

#### docs/package.json
- Add in package.json for the docs

#### webpack.config.js
- Remove `hot` and `inline` since they are in the script
- Change devTools to `eval-source-map` to use original code instead of generated
- Fix paths
- Add babel options back in. Since this file is in the docs directory, the `.babelrc` isn't picked up so they need to be added here, also.
- Specify path along with filename for bundle package


sambev:
I am not sure the package.json files and scripts are lining up. A fresh install of this for me yields:

```
 ~/p/mx-react-components/jmo/separate-packages > npm run dev                                                                    Wed Oct 10 13:52:54 2018

> mx-react-components@6.1.3 dev /Users/sam.beveridge/projects/mx-react-components
> cd docs && NODE_ENV=development webpack-dev-server --mode=development

sh: webpack-dev-server: command not found
```

Even if I have installed the docs package. I think this makes sense, given that the dev server is not installed in the lib's dev dependencies, so it wouldn't be able to find it in it's environment.

jmophoto:
Did you do a fresh install on both the package.json and the docs/package.json?

I have been running `npm ci` and then `npm run docs:ci`.

I tried to pare them down to only what was needed, but I may have missed something.

jmophoto:
Looking now, I am missing the `webpack-dev-server` in the main package. I can try adding that.

sambev:
Yeah that is with a fresh install in both and yeah, it is missing the dev server in the main. This has me thinking about what should be where. 🤔

cerinman:
Closing this in favor of https://github.com/mxenabled/mx-react-components/pull/813

PR #814, opened 2018-10-24
Remove radium and use glamor for the Drawer component

sambev:
This removes Radium from the drawer component in favor of glamor.

Profile with Radium:

screen shot 2018-10-24 at 2 36 23 pm

Profile without Radium:

screen shot 2018-10-24 at 2 35 24 pm

I see noticeable differences in opening speed and the drawer is only closing to about halfway on larger screens, would love for someone to test as well.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "mx-react-components",
-  "version": "6.1.1",
+  "version": "6.1.2",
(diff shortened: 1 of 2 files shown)


cerinman:
Putting this one on hold until we have more time to regression test.  Will shoot for the next release in the near future.

PR #815, opened 2018-10-29
apply focus to Select component

hokakede1:


sambev:
Did you run an npm update? Seems weird for you to have a ton of package-lock file changes with this.

src/components/Select.js
```diff
@@ -240,7 +241,11 @@ class Select extends React.Component {
           fontSize: theme.FontSizes.MEDIUM,
           padding: '8px 10px',
           position: 'relative',
-          boxSizing: 'border-box'
+          boxSizing: 'border-box',
+          ':focus': {
+            outline: 'dotted thin #333',
```
sambev: This should probably use a constant instead of #333. Maybe GRAY_900?

hokakede1:
@sambev I restore the old package-lock and apply GRAY_900

cerinman:
:shipit:

PR #817, opened 2018-11-15
Fix drawer swallowing events and spin not clearing the interval

phantomxc:
@cerinman The dreaded drawer. Also fixed something else that was annoying me.


lukeschunk:
🐉 ⚔️

cerinman:
:shipit:

Issue #818, opened 2019-01-29
Add boolean to toggle defaultLabelText changes on slice hover


ballen27:
Currently, the default action is to change the `defaultDataLabelText` and `defaultDataLabelValue` to the data of the slice on hover. This functionality is disabled when the DonutChart has children, however. Adding children should not be the solution to toggling the hover on slice action. 

**Change:**
Add an additional boolean value that allows you to toggle this hover functionality on or off. For specific cases where there is no relevant data associated with a slice and you want to hover over a slice without animation **AND** without updating the data label in the center of the chart.

PR #819, opened 2019-01-30
Add bool and function to toggle data label changing on hover

ballen27:
Issue: https://github.com/mxenabled/mx-react-components/issues/818

**Change:**

Add an additional boolean value that allows you to toggle this hover functionality on or off. For specific cases where there is no relevant data associated with a slice and you want to hover over a slice without animation AND without updating the data label in the center of the chart.

Update docs to describe new bool.

### DIFF
--- a/docs/components/DonutChartDocs.js
+++ b/docs/components/DonutChartDocs.js
@@ -160,7 +160,7 @@ class DonutChartDocs extends React.Component {
 
         <h5>shouldToggleDataLabelOnHover <label>Boolean</label></h5>
         <p>Default: true</p>
-        <p>When set to true, on slice hover, the defaultLabelText with change to display the data corresponding with the slice.</p>
+        <p>When set to true, on slice hover, the defaultLabelText will change to display the data corresponding with the slice.</p>
 
         <h5>width <label>Number</label></h5>
         <p>Default: 150</p>
--- a/docs/components/DonutChartDocs.js
+++ b/docs/components/DonutChartDocs.js
@@ -153,10 +153,14 @@ class DonutChartDocs extends React.Component {
 
         <h5>showDataLabel <label>Boolean</label></h5>
         <p>Default: true</p>
-        <p>If set to true, the defaultLabelText will be displayed in the center of the chart. On slice hover, the defaultLabelText with change to display the data corresponding with the slice.</p>
+        <p>If set to true, the defaultLabelText will be displayed in the center of the chart.</p>
 
         <h5>theme <label>Object</label></h5>
         <p>Customize the component&apos;s look. See <Link to='/components/theme'>Theme</Link> for more information.</p>
+
+        <h5>toggleDataLabelOnHover <label>Boolean</label></h5>
+        <p>Default: true</p>
+        <p>When set to true, on slice hover, the defaultLabelText with change to display the data corresponding with the slice.</p>
 
         <h5>width <label>Number</label></h5>
         <p>Default: 150</p>
--- a/docs/components/DonutChartDocs.js
+++ b/docs/components/DonutChartDocs.js
@@ -158,7 +158,7 @@ class DonutChartDocs extends React.Component {
         <h5>theme <label>Object</label></h5>
         <p>Customize the component&apos;s look. See <Link to='/components/theme'>Theme</Link> for more information.</p>
 
-        <h5>toggleDataLabelOnHover <label>Boolean</label></h5>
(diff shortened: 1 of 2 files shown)


src/components/DonutChart.js
```diff
@@ -38,6 +38,8 @@ class DonutChart extends React.Component {
     showBaseArc: PropTypes.bool,
     showDataLabel: PropTypes.bool,
     theme: themeShape,
+    toggleDataLabelOnHover: PropTypes.bool,
```
paniclater: this name makes it look like a function or similar to me. I prefer booleans to start with the prefix is, are, should or similar. In this case I think something like "shouldTriggerSliceHoverEvents" or "shouldToggleDataLabelOnHover" would be better options.

ballen27: I'll use shouldToggleDataLabelOnHover. Thanks, this was a difficult thing to try and name

cerinman:
The failing build here is due to an npm dependency in the docs having a vulnerability. If you pull master and merge it into your branch the build should succeed.

docs/components/DonutChartDocs.js
```diff
@@ -153,11 +153,15 @@ class DonutChartDocs extends React.Component {
...
         <p>Default: true</p>
-        <p>If set to true, the defaultLabelText will be displayed in the center of the chart. On slice hover, the defaultLabelText with change to display the data corresponding with the slice.</p>
+        <p>If set to true, the defaultLabelText will be displayed in the center of the chart.</p>
 
         <h5>theme <label>Object</label></h5>
         <p>Customize the component&apos;s look. See <Link to='/components/theme'>Theme</Link> for more information.</p>
 
+        <h5>shouldToggleDataLabelOnHover <label>Boolean</label></h5>
+        <p>Default: true</p>
+        <p>When set to true, on slice hover, the defaultLabelText with change to display the data corresponding with the slice.</p>
```
cerinman: grammar here. I think you meant `with` to be `will`.

src/components/DonutChart.js
```diff
@@ -35,9 +35,11 @@ class DonutChart extends React.Component {
     onMouseLeave: PropTypes.func,
     opacity: PropTypes.number,
     padAngle: PropTypes.number,
+    shouldToggleDataLabelOnHover: PropTypes.bool,
     showBaseArc: PropTypes.bool,
     showDataLabel: PropTypes.bool,
     theme: themeShape,
+
```
cerinman: Doesn't look like we need this line break. Can you remove it?

src/components/DonutChart.js
```diff
@@ -62,6 +64,7 @@ class DonutChart extends React.Component {
     padAngle: 0.02,
     showBaseArc: true,
     showDataLabel: true,
+    shouldToggleDataLabelOnHover: true,
```
cerinman: By defaulting this to true, you are introducing unexpected behavior for users of this library. It would probably be better to default to false and then set it to true in your usage of it.

cerinman: Scratch this comment. I just read the code below. Defaulting to true preserves the existing behavior. Disregard this comment. 👍

src/components/DonutChart.js
```diff
@@ -323,8 +326,9 @@ class DonutChart extends React.Component {
       } else {
         const activeDataSet = this.props.data[this.state.activeIndex] || {};
         const color = this.state.activeIndex === -1 ? colors[0] : colors[this.state.activeIndex];
-        const text = this.state.activeIndex === -1 ? this.props.defaultLabelText : activeDataSet.name;
-        const value = this.state.activeIndex === -1 ? this.props.formatter(this.props.defaultLabelValue) : this.props.formatter(activeDataSet.value);
+        const activeAndDataLabelHover = this.state.activeIndex !== -1 && this.props.shouldToggleDataLabelOnHover
+        const text = activeAndDataLabelHover ? activeDataSet.name : this.props.defaultLabelText
+        const value = activeAndDataLabelHover? this.props.formatter(activeDataSet.value): this.props.formatter(this.props.defaultLabelValue)
```
cerinman: You can simplify this.

```javascript
const value = this.props.formatter(activeAndDataLabelHover ? activeDataSet.value : this.props.defaultValue)
```

src/components/DonutChart.js
ballen27: True was the default behavior of the chart before this implementation. In our usage, we want to set it to false so there is no hover effect changing the text inside the chart.

cerinman:
:shipit:

PR #821, opened 2019-01-30
DateRangePicker - DefaultRange cleanup for a11y

cerinman:
- Replace radium with glamor to kill additional renders
- Adds aria-live polite to wrapper around default ranges options so that toggling of options gets called out immediately by screen readers.
- removed selected status from aria label and move it to aria-pressed attribute
- Remove object assign usage in favor of object spread for icon style
- Remove empty role on wrapping div around default range options.
- Fixes package vulnerability in docs


lukeschunk:
🏀

phantomxc:
:shipit:

PR #826, opened 2019-05-13
Remove expanded aria label from date range picker

DrewWarrenTIY:
Having both "aria-expanded" and "aria-popup" was being misread by Android TalkBack. This commit will remove the expanded label to make it more clear to the user that the element is a popup overlay and not expanded content.

### DIFF
--- a/src/components/DateRangePicker.js
+++ b/src/components/DateRangePicker.js
@@ -317,8 +317,7 @@ class DateRangePicker extends React.Component {
       <div className='mx-date-range-picker' style={styles.component}>
         <button
           aria-controls='calendarMenu'
-          aria-expanded={this.state.showSelectionPane}
-          aria-haspopup='menu'
+          aria-haspopup={true}
           aria-label={`${placeholderText}${
             this.props.selectedStartDate && this.props.selectedEndDate
               ? `, ${selectedStartDateFromPropsAsMoment.format('MMMM Do, YYYY')} to ${selectedEndDateFromPropsAsMoment.format('MMMM Do, YYYY')} currently selected`


src/components/DateRangePicker.js
```diff
@@ -317,8 +317,7 @@ class DateRangePicker extends React.Component {
       <div className='mx-date-range-picker' style={styles.component}>
         <button
           aria-controls='calendarMenu'
-          aria-expanded={this.state.showSelectionPane}
```
hokakede1: I am not too familiar with aria-expanded. May you explain why we remove it here?

src/components/DateRangePicker.js
```diff
@@ -317,8 +317,7 @@ class DateRangePicker extends React.Component {
...
-          aria-haspopup='menu'
```
hokakede1: aria-haspopup='menu'  🤦‍♂

hokakede1:
:shipit:

src/components/DateRangePicker.js
hokakede1: is aria-haspopup='menu' valid? I see only boolean for it.

cerinman: `aria-haspopup` does allow the string `menu` in addition to `true` and `false`.  While `menu` is a valid value, it does have unexpected consequences in JAWS where it causes that screen reader to not call out a button as a button.

Also the standard does not make clear that you use the string `menu`. The only place that can it even explained is under the menu button section here https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html

In general I think we want to stick to just using `true` and `false` for `aria-haspopup` because of how JAWS deviates in it's interpretation of what it means.

PR #827, opened 2019-05-14
Remove expanded aria label from date picker popup element

DrewWarrenTIY:


DrewWarrenTIY:
Comments from forked Repo PR: (Closing that PR but didn't want to lose the discussion)

Having both "aria-expanded" and "aria-popup" was being misread by Android TalkBack. This commit will remove the expanded label to make it more clear to the user that the element is a popup overlay and not expanded content

hokakede: is aria-haspopup='menu' valid? I see only boolean for it.

cerinman: aria-haspopup does allow the string menu in addition to true and false. While menu is a valid value, it does have unexpected consequences in JAWS where it causes that screen reader to not call out a button as a button.

Also the standard does not make clear that you can use the string menu. The only place that I can even find it explained is under the menu button section here https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html

In general I think we want to stick to just using true and false for aria-haspopup because of how JAWS deviates in it's interpretation of what it means.

me to hokakaede:
I think you removed a comment, but here is the answer to "what is aria-expanded" and "Why was it removed here"

Sure thing. aria-expanded label accepts true/false and basically tells the reader "This element is interactive and will toggle some expanded content" as well as the "expanded" or "collapsed" state.

The date picker element is an overlay that sits 'on top' of the UI, rather than expanding in line, so the expanded tag was already a bit inaccurate to use. However the major issue was that the Android TalkBack screen reader would ignore the "haspopup" tag if the "expanded" tag was also present. (JAWS and iOS VoiceOver would read both, which works fine) This meant that users were not being notified that the button would trigger a pop up.

Because the popup is wrapped in a "Focus Trap" element that would prevent the user from getting back to the 'toggle' button, it was misleading/problematic and needed to be removed.

cerinman:
:shipit:

PR #830, opened 2019-07-26
Add search to dropdown component

cgregs32:
https://github.com/mxenabled/mx-react-components/issues/832
Adds the ability to include search component to select component via `withSearch` prop 

Utilizes `Search` component within `select` component to be able to filter longer lists of options.
`withSearch` default to false, so this change will not affect existing calls to the `select` component

Screen Shot 2019-07-25 at 2 52 50 PM

### DIFF
--- a/docs/components/Changelog.js
+++ b/docs/components/Changelog.js
@@ -8,6 +8,11 @@ class Changelog extends React.Component {
 
         <h2>MX React Components V 6.1.0</h2>
 
+          <h3>6.1.11</h3>
+          <ul>
+            <li>Select - Add ability to search dropdown options via withSearch prop. (<a href='https://github.com/mxenabled/mx-react-components/issues/832'>#832</a>)</li>
+          </ul>
+
           <h3>6.1.10</h3>
           <ul>
             <li>Modal - Revert tabindex change due to Focus Trap error</li>
--- a/docs/components/SelectDocs.js
+++ b/docs/components/SelectDocs.js
@@ -6,6 +6,7 @@ const { Link } = require('react-router');
 const { Select } = require('mx-react-components');
 
 const Markdown = require('components/Markdown');
+const Code = require('components/Code');
 
 const options = [
   {
@@ -108,6 +109,7 @@ class SelectDocs extends React.Component {
         <p>If set to 'false', then the element will be marked as invalid and a red border will be placed around the element.</p>
         
         <h5>withSearch <label>Boolean</label></h5>
+        <p>Default: <Code>false</Code></p>
         <p>If set to true, the component renders with the Search component to support longer lists of options</p>
 
         <h3>Example</h3>
--- a/docs/components/SelectDocs.js
+++ b/docs/components/SelectDocs.js
@@ -106,6 +106,9 @@ class SelectDocs extends React.Component {
 
         <h5>valid <label>Boolean</label></h5>
         <p>If set to 'false', then the element will be marked as invalid and a red border will be placed around the element.</p>
+        
+        <h5>withSearch <label>Boolean</label></h5>
+        <p>If set to true, the component renders with the Search component to support longer lists of options</p>
 
         <h3>Example</h3>
         <Markdown>
--- a/docs/components/SelectDocs.js
+++ b/docs/components/SelectDocs.js
@@ -62,10 +62,20 @@ class SelectDocs extends React.Component {
         </h1>
 
         <h3>Demo</h3>
+
+        <h5>Default:</h5>
         <Select
           key='default'
           options={options}
           valid={true}
+        />
+
+        <h5>With prop: <Code>withSearch</Code></h5>
+        <Select
+          key='withSearch'
+          options={options}
+          valid={true}
+          withSearch={true}
         />
 
         <h3>Usage</h3>
--- a/src/components/Select.js
+++ b/src/components/Select.js
@@ -1,4 +1,5 @@
 const _isEqual = require('lodash/isEqual');
+const _includes = require('lodash/includes');
 const keycode = require('keycode');
 const PropTypes = require('prop-types');
 const Radium = require('radium');
@@ -164,7 +165,7 @@ class Select extends React.Component {
                 onChange={this._onSearchInputChange}
               />}
             {this.props.options
-              .filter(option => this.state.searchTerm ? option.displayValue.toLowerCase().includes(this.state.searchTerm) : true)
+              .filter(option => this.state.searchTerm ? _includes(option.displayValue.toLowerCase(), this.state.searchTerm) : true)
               .map(option => {
                 return (
                   <Option
--- a/src/components/Select.js
+++ b/src/components/Select.js
@@ -50,7 +50,8 @@ class Select extends React.Component {
     onChange () {},
     options: [],
     placeholderText: 'Select One',
-    valid: true
+    valid: true,
+    withSearch: false,
   };
 
   constructor (props) {
@@ -59,16 +60,13 @@ class Select extends React.Component {
     this.state = {
       isOpen: false,
       selected: props.selected,
-      options: props.options,
+      searchTerm: "",
     };
   }
 
   componentWillReceiveProps (newProps) {
     if (!_isEqual(newProps.selected, this.props.selected)) {
       this.setState({ selected: newProps.selected });
-    }
-    if(!_isEqual(newProps.options, this.props.options)) {
-      this.setState({ options: newProps.options })
     }
   }
 
@@ -90,7 +88,7 @@ class Select extends React.Component {
   };
 
   _close = () => {
-    this.setState({ isOpen: false, options: this.props.options });
+    this.setState({ isOpen: false, searchTerm: "" });
     this.elementRef.focus();
   };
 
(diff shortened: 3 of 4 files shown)


src/components/Select.js
```diff
@@ -41,7 +42,8 @@ class Select extends React.Component {
     selected: optionShape,
     selectedStyle: PropTypes.oneOfType([PropTypes.object, PropTypes.array]),
     theme: themeShape,
-    valid: PropTypes.bool
+    valid: PropTypes.bool,
+    withSearch: PropTypes.bool,
```
cerinman: Can you add a default state of false for this in `defaultProps`. I know it technically doesn't matter but it's nice to have some clarity around prop states for stuff like this.

src/components/Select.js
```diff
@@ -56,14 +58,18 @@ class Select extends React.Component {
...
     };
   }
 
   componentWillReceiveProps (newProps) {
     if (!_isEqual(newProps.selected, this.props.selected)) {
       this.setState({ selected: newProps.selected });
     }
+    if(!_isEqual(newProps.options, this.props.options)) {
+      this.setState({ options: newProps.options })
+    }
```
cerinman: We're not big fans of handing props off to state. Is there a specific reason you had for doing this? I would think you would be able to just filter the prop directly for mapping over it to render them rather than doing this.

src/components/Select.js
```diff
@@ -133,6 +139,14 @@ class Select extends React.Component {
     }
   };
 
+  _handleSearch = (e) => {
+    const filteredOptions = this.props.options.filter((item) => {
+      return item.displayValue.toLowerCase().includes(e.target.value)
```
paniclater: There's no support in Internet Explorer for includes - need to make sure this is tested inside of sabotage and moneymap in ie11 to make sure our babel configs are handling this.

cerinman: Good catch. I've gotten use to the poly fill.

src/components/Select.js
cerinman: You'll also need to make sure you update the associated docs with this new prop as well as the change log in the docs directory.

src/components/Select.js
paniclater: I think in this case it is probably ok. There needs to be a key change event here, and it either gets handled in state here or by component state or redux somewhere else

src/components/Select.js
cgregs32: Absolutely!

src/components/Select.js
cgregs32: Thanks for pointing this out.
I thought this would be the most simple solution for updating options.

I was (still am) struggling to determine how to update the props for mapping after the `searchInput` callback is fired and props are filtered.  Without triggering by setting state.

paniclater: Actually on further thought I think this should be handled differently.
The props coming in should initialize state. From there on out state should handle the options. 
It will be super weird if the person is typing to filter, and new props come in.

src/components/Select.js
cgregs32: Thanks, I will test this.

src/components/Select.js
cerinman: Why can't it be something like this?

```javascript
onSearchInputChange = e => {
  this.setState({ searchTerm: e.target.value })
}

render(){
  return (
    <searchInput onChange={this.onSearchInputChange} />
    {
      this.props.options
        .filter(option => {
          return searchTerm ? option.displayValue.toLowerCase().includes(searchTerm) : true
        })
        .map(option => JSX)
    }
  )
}

Comment (cgregs32) [reply src/components/Select.js]: @paniclater 
If I am not mistaken..
By declaring `options: props.options,` in the constructor, we are letting state handle options.

In `componentWillReceiveProps` we should be comparing this.state.options.
And we should refactor the use of `this.props.options` for `this.state.options`.

Comment (cgregs32) [reply src/components/Select.js]: @cerinman 
As you suggested that was another way of handling this.  Adding`searchTerm` as state in this component. 
But It felt that this component should not be responsible for keeping track of `searchTerm` as state, and it should be handled within `searchInput`. 
🤷‍♂

Comment (cerinman) [reply src/components/Select.js]: I would prefer the suggested method over handing props off to state. Equality checks in life cycle methods tend to be more costly than just running a filter on every render. Equality checks also have a tendency to introduce bugs because of their complexity and/or just not realizing something isn't a deep equality check.

Examples being objects within objects that have arrays of more objects or a check just checking the objects id rather than checking actual key/value pairs within an object. That might not be the case currently but this type of implementation opens the door for such things if some one passes an option object that isn't laid out the way you would expect. 

And even if you ensure deep equality checking, you usually do so at the cost of computational time. Also equality checks on props/state can also cause additional renders that you do not expect. The component may only use one or two key/values pairs  but the option object may be more complex. Changing of an unused key/value pair will still cause the component to re render and that can cause confusion when trying to debug.

All this adds up to a larger surface area to cause problems. The filter method suggested is not a guarantee against any of this of course but it tends to be way less risky.

File: docs/components/SelectDocs.js [hunk 4]
```diff
@@ -106,6 +106,9 @@ class SelectDocs extends React.Component {
 
         <h5>valid <label>Boolean</label></h5>
         <p>If set to 'false', then the element will be marked as invalid and a red border will be placed around the element.</p>
+        
+        <h5>withSearch <label>Boolean</label></h5>
+        <p>If set to true, the component renders with the Search component to support longer lists of options</p>
```
cerinman: Can you provide the default value here as well. You can look at other props to see how we provide the default value for props in the docs.

src/components/Select.js
```diff
@@ -149,32 +157,40 @@ class Select extends React.Component {
...
-                  ) : null}
-                  <div style={styles.optionText}>{option.displayValue}</div>
-                  {_isEqual(option, this.state.selected) ? <Icon size={20} type='check' /> : null }
-                </Option>
-              );
-            })}
+            {this.props.withSearch && 
+              <SearchInput
+                focusOnLoad={true}
+                onChange={e => this._onSearchInputChange(e)}
```
cerinman: You shouldn't have to wrap this in an anon arrow function.  `onChange={this._onSearchInputChange}`

src/components/Select.js
```diff
@@ -149,32 +157,40 @@ class Select extends React.Component {
...
+              />}
+            {this.props.options
+              .filter(option => this.state.searchTerm ? option.displayValue.toLowerCase().includes(this.state.searchTerm) : true)
+                .map(option => {
+                  return (
```
cerinman: Can you pull this map call back to be the same level as the filter?

cgregs32: 👍

docs/components/SelectDocs.js
cgregs32: Got it! Thanks.

src/components/Select.js
cgregs32: 👍

src/components/Select.js
cerinman: If you're having issues testing in IE 11 you can always just switch to the lodash includes function. https://lodash.com/docs/4.17.15#includes

src/components/accessibility/Listbox.js
```diff
@@ -74,6 +76,7 @@ class Listbox extends React.Component {
         break;
       case 'enter':
       case 'space':
+        if (this.props.withSearch && this.state.focusedIndex <= 0) return
```
cerinman: Nitpick but if the search input will always be index 0 then can you just `this.state.focusedIndex === 0`? Also, can you add a comment here that explains why you're bailing out on preventing default on space?

cgregs32: @cerinman
I observed two values the focusedIndex state could be when SearchInput is focused: -1 and 0. So this.state.focusedIndex === 0 will not work for every case.

I will add a comment to document why this line exists.

src/components/Select.js
cgregs32: I was able to confirm `includes()` is working in ie11 within sabotage.

includes_ie11

cerinman: I think it is important to see if if works in IE11 within the docs here. It's possible if works in Sabotage because Babel is providing a poly fill there but that it won't work here on it's own. You can run the local dev server in this repo to spin up the docs.

src/components/accessibility/Listbox.js
cerinman: sounds good.

src/components/Select.js
cgregs32: @cerinman 
I Updated the component to use lodash `includes` as you recommended.  I verified this works in the docs on ie11.

docs/components/Changelog.js
```diff
@@ -8,6 +8,11 @@ class Changelog extends React.Component {
 
         <h2>MX React Components V 6.1.0</h2>
 
+          <h3>6.1.9</h3>
```
cerinman: Looks like this version is now wrong because Drew merged and released while we were reviewing this. I think this will now need to be 6.1.11 but I would get with @DrewWarrenTIY  and make sure you guys have the change log updated correctly between the two of you. He'll have another MR that will probably go along with yours on this release.

cgregs32:
@cerinman 
Please take another look!

cerinman:
:shipit:

PR #831, opened 2019-07-26
Remove menu role that was causing a11y issues in JAWS

DrewWarrenTIY:
role="menu" affects keypress behavior of contained elements in JAWS screen reader. I noticed no side effects of removing the role.


cerinman:
:shipit:

PR #833, opened 2019-07-30
Remove extraneous package, default role, focus style override

DrewWarrenTIY:


src/components/Modal.js
```diff
@@ -218,7 +216,6 @@ class Modal extends React.Component {
                 onKeyUp={e => e.keyCode === 13 && this.props.onRequestClose()}
                 role='button'
                 style={styles.close}
-                tabIndex={0}
```
cerinman: While you're here.... Is the role='button' and onKeyUp really needed on this native button? Can we remove them too?

src/components/Modal.js
```diff
@@ -204,7 +203,6 @@ class Modal extends React.Component {
               ref={ref => this._modalContent = ref}
               role={this.props.role}
               style={Object.assign({}, styles.content, this.props.contentStyle)}
-              tabIndex={0}
```
cerinman: Just to be safe. You didn't run into any console errors around assigning undefined to the role here on line 204 above now that the default has been removed?

DrewWarrenTIY: Double checked with passing no role prop. No console errors

src/components/Modal.js
DrewWarrenTIY: Removed

src/components/Modal.js
```diff
@@ -204,7 +203,7 @@ class Modal extends React.Component {
...
+              tabIndex={-1} // DO NOT REMOVE: Focus will default to wrapper with focusOnLoad prop
```
cerinman: DO NOT REMOVE: ?

cerinman:
I don't see a :shipit:. Why did you merge?

cerinman:
Also, @cgregs32 will need to add his MR to the release you will do with this. https://github.com/mxenabled/mx-react-components/pull/830 Make sure you two work together on the changelog update.

src/components/Modal.js
cerinman: Never mind on this one. I refreshed and realized you added it.

PR #834, opened 2019-08-05
Jameson/fixed default direction prop and spinDirection to match docs

Jameson13B:
This MR fixes the Spin component. The docs state the default is counterclockwise but defaults to clockwise. The spinDirection was also backwards so if you choose clockwise it spins counter and vice versa.

### DIFF
--- a/src/components/Spin.js
+++ b/src/components/Spin.js
@@ -10,14 +10,14 @@ class Spin extends React.Component {
   };
 
   static defaultProps = {
-    direction: 'clockwise',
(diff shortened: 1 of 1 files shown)


phantomxc:
Hmmm, I think I would prefer to update the docs?

Jameson13B:
@phantomxc we can change the docs to show the default is clockwise, however, the ternary for the spinDirection will still be reversed and needs to be updated.

cerinman:
:shipit:

PR #837, opened 2019-08-23
Fix npm audit vulnerability

whiteinge:


whiteinge:
I don't think that CI is quite right. I can't replicate that failure locally (`npm ci && npm audit`) and this PR should fix the thing audit is failing on. Should we merge and see if that does it? Other thoughts?

phantomxc:
:shipit:

PR #839, opened 2019-09-04
Jameson/add children to message box

Jameson13B:
Added the ability to add arbitrary children to the `MessageBox` component. Also added the ability to style the Container in `DisplayInput`.


whiteinge:
:shipit:

PR #840, opened 2019-09-09
Jameson/Added ability to set aria-live and role on elements

Jameson13B:
There are a couple issues with this component where it reads the messagebox on repeat with current `aria-live` and `role` attributes. Added props to be able to specify depending on use case.


src/components/MessageBox.js
```diff
@@ -22,13 +24,18 @@ class MessageBox extends React.Component {
...
+
   render () {
     const theme = StyleUtils.mergeTheme(this.props.theme);
     const styles = this.styles(theme);
 
     return (
       <article
-        aria-live='polite'
+        aria-live={this.props.ariaLive}
         className='mx-message-box'
```
Firesheath: It looks like the default was 'polite' and it is now being set to 'region' for the aria-live attribute. I'm worried this may break something that was depending on 'polite' for this attribute?

src/components/MessageBox.js
```diff
@@ -115,7 +122,7 @@ class MessageBox extends React.Component {
       messageContent: {
         boxSizing: this.props.isSmall ? null : 'content-box',
         textAlign: 'left',
-        width: this.props.isSmall ? '100%' : '75%',
+        width: this.props.children ? '100%' : '75%',
```
Firesheath: Was it intentional to remove the `isSmall` prop from changing the width?

src/components/MessageBox.js
Jameson13B: Missed this completely. Nice catch.

src/components/MessageBox.js
Jameson13B: For backwards compatibility we should leave it. I have changed it to an or. When it small it should take 100%, but also when there are children it should be 100%.

PR #841, opened 2019-09-13
DateRangePicker Render Props

cerinman:
Allow for custom trigger element via render props to accommodate for new internal design for mobile.

I also removed Radium from the component(s) in favor of glamor and replaced `Object.assign` calls with object spread.

Example

```javascript
<DateRangePicker>
  {elementProps => (
    <button {...elementProps}> My Custom Button</button>
  )}
</DateRangePicker>
```

See docs changes for more information/better example.

Screen Shot 2019-09-13 at 4 26 46 PM


src/components/DateRangePicker.js
```diff
@@ -313,42 +312,51 @@ class DateRangePicker extends React.Component {
...
-            ) : placeholderText}
-          </div>
-          <Icon
-            size={20}
-            style={styles.selectedDateCaret}
-            type={this.state.showSelectionPane ? 'caret-up' : 'caret-down'}
-          />
-        </button>
+        {
+          children && typeof children === 'function' ? this.props.children(buttonProps) :
```
cerinman: This is the main change here. If children and children are a function then we call it and pass the button Props that would have been spread across the normal button.

src/components/DateRangePicker.js
```diff
@@ -695,4 +704,4 @@ class DateRangePicker extends React.Component {
   };
 }
 
-module.exports = withTheme(Radium(DateRangePicker));
```
cerinman: And here is the second change. I've removed Radium in favor of using glamor in the related components.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "mx-react-components",
-  "version": "6.5.1",
+  "version": "6.5.2",
```
sambev: I think semantically speaking, this isn't a bug fix, but a new feature, so `6.6.0` would be slightly more appropriate.

DrewWarrenTIY:
:shipit:

src/components/DateRangePicker.js
sambev: I think it is a bit odd that you are using `children` and `this.props.children` here. It reads like they could be different things but they are the same thing. I would prefer that you use one or the other.

DrewWarrenTIY: Looks good and well documented

package.json
cerinman: Updated

DrewWarrenTIY:
:shipit: New version 6.6.0 looks good 👍

PR #842, opened 2019-09-17
Adds new prop elementPropsForChartSVG

cerinman:
To allow for adding element properties to the wrapping SVG element such as aria hidden for accessibility.


DrewWarrenTIY:
:shipit:

PR #843, opened 2019-11-04
feat: added the disabled attribute to the Display Input

TayteStokes:
Added the `disabled` attribute to the Display Input component to avoid having users be able to focus into the input when the `disabled` attribute is present


sambev:
:shipit:

cerinman:
Closing this and opening a new one due to issues in Taytes Forked copy

PR #844, opened 2019-11-05
Move jest to dev dependencies and fix npm audit issues

cerinman:
Some how we ended up with jest as a normal dependency in package.json. This means jest and all it's dependencies were being included in the bundle internally. This MR fixes the issue by moving jest to the devDependencies object instead.

I also ran `npm audit fix` to resolve two high warnings with dependencies around jest.

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -4574,9 +4574,9 @@
       "dev": true
     },
     "handlebars": {
-      "version": "4.1.2",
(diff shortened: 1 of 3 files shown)


sambev:
:shipit:

cerinman:
Closing and then reopening to trigger the CI again. It's saying there is an audit failure but locally there is no audit issue.

PR #845, opened 2019-11-05
Bump version of numeral to v 2.0.6

cerinman:
We're out of sync with numeral version used internally. This causes webpack to include two copies of numeral in the final bundle.

Tested with all components via the docs. The major breaking changes in numeral v 2 deal with localization which we are not doing anything with here. v 2 also no longer throws if an invalid number is supplied to a numeral call.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5714,9 +5714,9 @@
       "dev": true
     },
     "numeral": {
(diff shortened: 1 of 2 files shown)


sambev:
:shipit:

PR #851, opened 2019-11-21
Reset focused day matches originally set focused day

DrewWarrenTIY:
When the focusedDay piece of state is originally set, it is calculated from props.selectedEndDate. For some reason when it was set when the picker is closed via the _resetToPropValuesAndClose method, it was being set to props.selectedStartDate which was leading to inconsistent behavior specifically with keyboard navigation between the first time the picker was opened and subsequent openings.

and dependency vulnerability fixes


TayteStokes:
:shipit:

phantomxc:
:shipit:

PR #853, opened 2019-11-25
Final casing update

DrewWarrenTIY:
Adds a command in the build script that will remove dist before building, which should prevent any old instances of `Index.js` from persisting.


cerinman:
:shipit:

PR #854, opened 2019-11-26
Input error labels

DrewWarrenTIY:
When there is an invalid input in the DisplayInput component, the aria-label of the input will now announce the error message, if provided, to indicate to the user when they focus the input that they need to make corrections.


cerinman:
:shipit:

PR #855, opened 2019-12-04
Fix: Improved RadioButton component accessibility

TayteStokes:
Replaced the wrapping component div to a native button element as well as adjusted some styling for the button element.


cerinman:
Tayte, we may need to add a role of radio as well. Can you do some research and verify that role of radio is needed. I'll slack you an internal MR where we did role of check box on buttons for a similar reason.

cerinman:
:shipit:

cerinman:
:shipit:

PR #860, opened 2020-01-14
Jameson/Updated react and react-dom versions

Jameson13B:
This MR updates react and react-dom versions to latest


DrewWarrenTIY:
:shipit:

PR #863, opened 2020-01-28
Jameson/add style prop to loader

Jameson13B:
This MR add the `style` prop to the Loader component to override styles.

### DIFF
--- a/src/components/Loader.js
+++ b/src/components/Loader.js
@@ -1,5 +1,6 @@
 const PropTypes = require('prop-types');
 const React = require('react');
+const _merge = require('lodash/merge')
 
 import { withTheme } from './Theme';
 const Spin = require('./Spin');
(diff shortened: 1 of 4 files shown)


src/components/Loader.js
```diff
@@ -53,7 +55,7 @@ class Loader extends React.Component {
   styles = (theme) => {
     const color = this.props.color || theme.Colors.PRIMARY;
 
-    return {
+    return _merge({}, {
```
phantomxc: You can wrap this in an if checking for the prop so we aren't merging nothing if we aren't passing in any styles.

Jameson13B: @phantomxc Great idea. I have that updated and found a typo. Should be good to go. Thanks!

src/components/Loader.js
```diff
@@ -53,47 +55,90 @@ class Loader extends React.Component {
...
+          borderTop: '3px solid ' + color,
+          borderRight: '3px solid transparent',
+          borderBottom: '3px solid transparent',
+          borderLeft: '3px solid transparent'
+        },
+        text: {
+          marginTop: '10px',
+          fontSize: '10px'
+        }
+      }, this.props.styles)
```
phantomxc: Lets move the styles to a const so it's not duplicated in your ternary.

phantomxc:
:shipit:

PR #864, opened 2020-02-03
Adds iconSize to the button component

phantomxc:
Adds the ability to control the icon size in the button component with the prop `iconSize`.


cerinman:
:shipit:

PR #865, opened 2020-02-04
Added learn-more page components for Finstrong

jacob-pettingill:
I just used the names from figma for the naming of the icons.


DrewWarrenTIY:
Do the SVG names need to be added to `src/constants/App.js`? Other SVG names are in there. I'm not sure if that's for the docs or the Icon component.

jacob-pettingill:
> Do the SVG names need to be added to `src/constants/App.js`? Other SVG names are in there. I'm not sure if that's for the docs or the Icon component.

Good call. Added the values. Checked for other places that need it too. Looks good.

DrewWarrenTIY:
:shipit:

PR #867, opened 2020-03-04
Removed active styles and focus styles that clashed

Jameson13B:
On the `Select` component we had some active and focus styles that clashed. I have removed some styles that cause this and have tested to make sure it still functions as should.

### DIFF
--- a/src/components/Select.js
+++ b/src/components/Select.js
@@ -177,8 +177,7 @@ class Select extends React.Component {
                     onClick={haltEvent(this._handleOptionClick.bind(null, option))}
                     style={Object.assign({},
                       styles.option,
-                      this.props.optionStyle,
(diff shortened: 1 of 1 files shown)


src/components/Select.js
```diff
@@ -177,8 +177,7 @@ class Select extends React.Component {
                     onClick={haltEvent(this._handleOptionClick.bind(null, option))}
                     style={Object.assign({},
                       styles.option,
-                      this.props.optionStyle,
-                      _isEqual(option, this.state.selected) ? styles.activeOption : null
```
TayteStokes: Is `_isEqual` being used anywhere else or could we remove this import?

Jameson13B: @TayteStokes it is used in a couple other spots on this file.

TayteStokes:
Looks good! :shipit:

PR #868, opened 2020-03-11
Spin & Loader | Updating the default direction of spin

garrowp:
#### Description

The direction of our `Spin` and `Loader` components have gone from counterclockwise to clockwise. In Serenity, there are many instances where they are going opposite directions.

#### Solution

Rather than changing each individual instances of the `Spin` and `Loader` components, I've changed them both to default to spin in a clockwise direction. I've also added `direction` to the `Loader` API, accepting either 'clockwise' or 'counterclockwise' as valid inputs. This gets passed on to the `Spin` component inside of Loader.

Docs have been updated to note that the default direction is now clockwise, details about the new `direction` prop of the `Loader` API, and a demo showing the `Loader` spinning counterclockwise.

Also ran `npm audit fix` to update vulnerable dependencies because there were so many that the pipeline was failing.

Note: A bunch of the changes to this file were prettier changes

### DIFF
--- a/docs/components/LoaderDocs.js
+++ b/docs/components/LoaderDocs.js
@@ -1,11 +1,11 @@
 // eslint-disable react/jsx-indent rule added for proper <Markdown /> formatting
 /* eslint-disable react/jsx-indent */
-const React = require('react');
-const { Link } = require('react-router');
+const React = require('react')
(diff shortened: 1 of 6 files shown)


phantomxc:
:shipit:

PR #873, opened 2020-06-12
Fix Travis CI

HappyViki:
This PR is successful if Travis passes. This shouldn't actually get merged though.


HappyViki:
Changes:
- updated root and docs
- fixed all eslint errors

Very helpful to me:
https://itnext.io/fixing-security-vulnerabilities-in-npm-dependencies-in-less-than-3-mins-a53af735261d

src/components/DateRangePicker/DefaultRanges.js
```diff
@@ -28,27 +34,25 @@ class DefaultRanges extends React.Component {
...
-                handleDefaultRangeSelection(range);
-                this.setState({ selectedOption: index });
+                handleDefaultRangeSelection(range)
+                this.setState({ selectedOption: index })
               }}
             >
               <div>
                 <Icon
                   size={20}
-                  style={{ ...styles.rangeOptionIcon,
```
DrewWarrenTIY, line 38: As far as I can tell, removing the `rangeOptionIcon` style rules is the only actualy change in this PR. Is that right?

DrewWarrenTIY: Everything else seems to be the package updates for the CI warnings or prettier changes to formatting.

HappyViki: No, there are changes in other files too. I just forgot to not save with Prettier again.

HappyViki: There, go rid of Prettier formatting.

HappyViki:
> This shouldn't actually get merged though.

It should now. I wrote that when I was frustrated with my donut pr and wasn't sure if I was going to eventually merge this.

src/components/DateRangePicker/DefaultRanges.js
DrewWarrenTIY: Was it intentional to get rid of the `rangeOptionIcon` style rules?

DrewWarrenTIY: Answered in Slack - variable was undeclared.

DrewWarrenTIY:
:shipit:

src/components/DateRangePicker/DefaultRanges.js
HappyViki: We discussed in private, and it's all just updating code. Should work the same in production.

src/components/Tabs.js
```diff
@@ -22,8 +22,8 @@ const TabsFactory = ({ type = 'standard', ...props }) => {
   );
 };
 
-TabsFactory.propTypes = Object.assign({
-  type: PropTypes.oneOf(Object.keys(TabsTypes))
-}, Tabbable.PropTypes);
```
sambev, line 27: I don't think you need to do this conversion and I also don't think this is correct. Before we were merging one object into another:
```
Object.assign({type: ....}, Tabble.PropTypes)
```
Now you are flattening one object an array that is a property on an object.
```
{ type: PropTypes.oneOf([..., ...Tabable.PropTypes])}
```

PR #874, opened 2020-06-16
Fix structure of propTypes of TabsFactory

HappyViki:
This PR fixes the code that made docs crash. `propTypes` was refactored incorrectly.


DrewWarrenTIY:
:shipit:

PR #876, opened 2020-06-30
Revert Package Versions

HappyViki:
`Moment` broke after this merge: https://github.com/mxenabled/mx-react-components/pull/873

I'm reverting as much as possible, and only updating packages that need to be updated.


HappyViki:
@sambev Should we use older versions of `minimist` for for some dependancies?
https://www.npmjs.com/advisories/1179

PR #878, opened 2020-07-13
Sb/fix package updates

sambev:
This resets the package files back to what they were at 8.0.0.

Then we upgraded only the packages that needed audit fixes.

We can't yet upgrade jest-enzyme or babel/cli to the appropriate versions since they haven't updated yet either.

Fixed some lint issues.

The commit messages tell a good story if you read them commit by commit.

cc @HappyViki @cerinman @phantomxc


docs/package.json
```diff
@@ -36,15 +35,12 @@
     "babel-core": "^7.0.0-bridge.0",
     "babel-eslint": "^9.0.0",
     "babel-loader": "^8.1.0",
```
sambev, line 37: cc @cerinman @phantomxc I am pretty sure we don't need jest or enzyme in the docs project because there are no tests there.

package.json
```diff
@@ -36,38 +36,38 @@
     "focus-trap-react": "^3.1.4",
     "glamor": "^2.20.40",
     "keycode": "^2.1.8",
-    "lodash": "^4.17.15",
-    "moment": "^2.26.0",
-    "moment-timezone": "^0.5.31",
```
sambev, line 41: We upgraded lodash to 4.17.19, jest to 26.1.0, which requires a higher version of node, which we should be on anyway at that this point (I updated the travis file). We also ran npm audit fix which fixed some nested deps.

PR #882, opened 2020-09-14
Implements focus outline for accessibility to the RadioButton component

TayteStokes:
This change updates the outline style for the radio button when the user is focused on the element.

It also runs `npm audit fix` to fix the 4000+ vulnerabilities that was making the CI build fail. Tested each component in dev environment to make sure this didn't introduce breaking changes to the components.

### DIFF
--- a/src/components/RadioButton.js
+++ b/src/components/RadioButton.js
@@ -1,5 +1,6 @@
 const PropTypes = require('prop-types');
 const React = require('react');
+const Radium = require('radium');
 
 import { withTheme } from './Theme';
 const { themeShape } = require('../constants/App');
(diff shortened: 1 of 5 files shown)


src/components/RadioButton.js
```diff
@@ -46,15 +46,17 @@ class RadioButton extends React.Component {
 
   styles = (theme) => {
     return {
-      component: Object.assign({}, {
+      component: css(Object.assign({}, {
```
jacob-pettingill, line 49: Do we support the spread operator here?

jacob-pettingill (changes requested):

src/components/RadioButton.js
TayteStokes: Yeah, let me get that change pushed!

jacob-pettingill (approved):

PR #883, opened 2020-09-15
WIP: Fixes Color Adjustment

jacob-pettingill:
https://gitlab.mx.com/mx/moneymap/-/issues/2636

BECU was having an issue where they chose a new color and the adjustment was causing some sort of wrap around. I adjusted the colorAdjust function to be more verbose in it's implementation, resolving the errors found in the bug. 

Now when a color is adjusted e.g adjustColor('#123456', -15) 

the R, G, and B values are extracted and converted to decimal, the value is then added to each (which causes the light/dark values to increase or decrease), and the values are bounded at 'FF' or 255 and '00' or 0.

I tested this change on the colors that BECU described. I will make sure to test this on several other client colors to ensure that it works as intended.

### DIFF
--- a/docs/components/Changelog.js
+++ b/docs/components/Changelog.js
@@ -8,6 +8,11 @@ class Changelog extends React.Component {
 
         <h2>MX React Components V 8.0.0</h2>
 
+        <h3>8.2.0</h3>
+        <ul>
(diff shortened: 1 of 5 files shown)


TayteStokes:
:shipit:

TayteStokes (approved):

PR #885, opened 2020-09-18
Revert RadioButton styles to use correct class

TayteStokes:
This change reverts the class name change and removes glamor from being used in component.


jacob-pettingill:
:shipit:

jacob-pettingill (approved):

PR #888, opened 2020-10-09
add statusRole prop to displayinput component for a11y

DrewWarrenTIY:


DrewWarrenTIY:
After talking with @sambev it has been determined that the low severity vulnerability identified in the Travis check can be ignored for now. We may fork `glamor` at some point to update this dependency on our own, or try to submit a PR that will do so.

TayteStokes (approved):
LGTM :shipit:

PR #889, opened 2020-10-22
Adds checks to .blur() executions

TayteStokes:
This change adds checks to make sure the elements exist and have the blur method available before executing the method.


sambev:
:shipit:

PR #892, opened 2021-01-04
Add Aria Label to Toggle Switch

jacob-pettingill:
This extends the current ToggleSwitch API to allow for an ariaLabel input.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "d3": "^3.5.6",
     "focus-trap-react": "^3.1.4",
     "glamor": "^2.20.40",
+    "js-logger": "^1.6.1",
     "keycode": "^2.1.8",
     "lodash": "^4.17.20",
     "moment": "^2.26.0",
--- a/package.json
+++ b/package.json
@@ -35,7 +35,6 @@
     "d3": "^3.5.6",
     "focus-trap-react": "^3.1.4",
     "glamor": "^2.20.40",
-    "js-logger": "^1.6.1",
     "keycode": "^2.1.8",
     "lodash": "^4.17.20",
     "moment": "^2.26.0",
--- a/src/components/ToggleSwitch.js
+++ b/src/components/ToggleSwitch.js
@@ -11,7 +11,7 @@ const StyleUtils = require('../utils/Style');
 
 class ToggleSwitch extends React.Component {
   static propTypes = {
-    ariaLabel: PropTypes.string, 
+    buttonProps: PropTypes.object,
     checked: PropTypes.bool,
     elementRef: PropTypes.func,
     falseIcon: PropTypes.string,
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -35,6 +35,7 @@
     "d3": "^3.5.6",
     "focus-trap-react": "^3.1.4",
     "glamor": "^2.20.40",
+    "js-logger": "^1.6.1",
```
sambev, line 38: Why was this added? Doesn't seem related.

src/components/ToggleSwitch.js
```diff
@@ -11,6 +11,7 @@ const StyleUtils = require('../utils/Style');
 
 class ToggleSwitch extends React.Component {
   static propTypes = {
+    ariaLabel: PropTypes.string,
```
sambev, line 14: Do you think it would be worth the time to add a more generic 'pass through' prop like `buttonProps` that get spread onto the button? That way we don't have to keep adding individual props like this.

sambev:
would you mind checking out the audit failure and seeing if it is easily fixed?

package.json
jacob-pettingill: So whenever I try to do an `npm link` without it installed, the build fails on the serenity side. This was just a simple fix and felt like it didn't harm to have it in the dev-deps. I know that other people have run into the same issue.

I can dig deeper however and see if there is something going on that I don't realize.

src/components/ToggleSwitch.js
jacob-pettingill: Yeah I had thought to do that but went this way on the first run. I like the idea of the generic pass through. I will get that implemented.

package.json
jacob-pettingill: @TayteStokes has run into this issue before. Our solution in the past has been to install it, then remove it. However @DrewWarrenTIY says that he hasn't run into the issue with `js-logger` while linking.

Essentially what I do
- run `npm link` in `mx-react-components`
- run `npm dev-release` in `mx-react-components`
- run `npm link mx-react-components` in `serenity`
- run `npm run dev` in `serenity.

Then I get these errors: 
```
ERROR in ../mx-react-components/dist/constants/IconPaths.js
Module not found: Error: Can't resolve 'js-logger' in '/Users/jacob.pettingill/repos/mx-react-components/dist/constants'
 @ ../mx-react-components/dist/constants/IconPaths.js 3:13-33
 @ ../mx-react-components/dist/index.js
 @ ./src/components/app/IEDeprecationModal.js
 @ ./src/components/app/AppWrapper.js
 @ ./src/widgetLoader.js
 @ multi ./src/widgetLoader.js
```

I can't seem to nail down why this error is occurring. @sambev @phantomxc any ideas on why this might be happening?

The simple solution would be to either install `js-logger` as a `dev-dependency` or else install it while I test and then remove it. But I would like to nail down a longer term solution.

jacob-pettingill: I have tried `npm ci`, and removing the modules. As well as a complete refresh on the linking several times.

jacob-pettingill: It looks like it might have something to do with webpack configs. I guess symlink resolution is a difficult problem to resolve. https://github.com/webpack/webpack/issues/985. I am going to just install temporarily for now.

jacob-pettingill:
@sambev looks like this pull request https://github.com/threepointone/glamor/pull/395/files, that has been opened, would resolve that audit check. The audit is failing because `node-fetch` a 3rd level dependency of `fbjs` has a denial of service vulnerability. I don't know if there is anything we can do at the moment.

jacob-pettingill:
@sambev since we can't do anything about the audit check for now, is this good to go based on the changes made? Or any other thoughts?

package.json
sambev: I think it would be best to investigate removing js-logger entirely. Is it being used?

sambev:
:shipit:

PR #900, opened 2021-03-17
Switch from Glamor to our own cssinjs wrapper lib

### LINKED ISSUE #895: Documenting audit failure in node-fetch package
Currently when the Travis CI Pipeline runs there is a vulnerability detected that looks like this:

```
┌──────────────────────────────────────────────────────────────────────────────┐
│                                Manual Review                                 │
│            Some vulnerabilities require your attention to resolve            │
│                                                                              │
│         Visit https://go.npm.me/audit-guide for additional guidance          │
└──────────────────────────────────────────────────────────────────────────────┘
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Low           │ Denial of Service                                            │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ node-fetch                                                   │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >=2.6.1 <3.0.0-beta.1|| >= 3.0.0-beta.9                      │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ glamor                                                       │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ glamor > fbjs > isomorphic-fetch > node-fetch                │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://nodesecurity.io/advisories/1556                      │
└───────────────┴──────────────────────────────────────────────────────────────┘
found 1 low severity vulnerability in 960 scanned packages
  1 vulnerability requires manual review. See the full report for details.
  ```

I did some digging around and found that glamor only imports the following functions from the `fbjs` package:

```
node_modules/glamor/lib/CSSPropertyOperations/index.js:9:var _camelizeStyleName = require('fbjs/lib/camelizeStyleName');
node_modules/glamor/lib/CSSPropertyOperations/index.js:17:var _hyphenateStyleName = require('fbjs/lib/hyphenateStyleName');
node_modules/glamor/lib/CSSPropertyOperations/index.js:21:var _memoizeStringOnly = require('fbjs/lib/memoizeStringOnly');
node_modules/glamor/lib/CSSPropertyOperations/index.js:25:var _warning = require('fbjs/lib/warning');
node_modules/glamor/lib/CSSPropertyOperations/dangerousStyleValue.js:11:var _warning = require('fbjs/lib/warning');
```

and then further went into each of those functions to ensure that none of them are calling the `isomorphic-fetch` function.

This shows me that this vulnerability does not exist in glamor and we are safe to ignore it in our Travis CI output... which is good because glamor is no longer being maintained and hasn't been published for 3 years. 😂

### LINKED ISSUE #897: Switch from Glamor to another (API-compatible) lib
We have a months-old `npm audit` (low) vulnerability which is a dep-of-a-dep-of-a-dep-of-a-dep (#895) and has been fixed upstream that we need to resolve.

# Background

We opted for [Glamor](https://github.com/threepointone/glamor) due to the simple, straightforward `const className = css({...styles})` API and because the API was effectively "done" and would not see churn or changes. (Plus it had a little momentum already across some of our projects.) Another benefit is that same API is also used across several other CSS-in-JS libraries.

There are two noteworthy drawbacks to Glamor:
```
% curl --compressed -Ssw '%{size_download}\n' -o /dev/null https://unpkg.com/glamor@2.20.40/umd/index.min.js | numfmt --to=si
9.2K

% npm ls
└─┬ glamor@2.20.40
  ├─┬ fbjs@0.8.17
  │ ├── core-js@1.2.7
  │ ├─┬ isomorphic-fetch@2.2.1
  │ │ ├─┬ node-fetch@1.7.3
  │ │ │ ├─┬ encoding@0.1.13
  │ │ │ │ └─┬ iconv-lite@0.6.2
  │ │ │ │   └── safer-buffer@2.1.2
  │ │ │ └── is-stream@1.1.0
  │ │ └── whatwg-fetch@3.5.0
  │ ├─┬ loose-envify@1.4.0
  │ │ └── js-tokens@4.0.0
  │ ├── object-assign@4.1.1 deduped
  │ ├─┬ promise@7.3.1
  │ │ └── asap@2.0.6
  │ ├── setimmediate@1.0.5
  │ └── ua-parser-js@0.7.23
  ├─┬ inline-style-prefixer@3.0.8
  │ ├── bowser@1.9.4
  │ └─┬ css-in-js-utils@2.0.1
  │   ├── hyphenate-style-name@1.0.4
  │   └── isobject@3.0.1
  ├── object-assign@4.1.1
  ├─┬ prop-types@15.7.2
  │ ├── loose-envify@1.4.0 deduped
  │ ├── object-assign@4.1.1 deduped
  │ └── react-is@16.13.1
  └── through@2.3.8
```

The huge dependency chain makes a huge surface area for problems like the current `npm audit` failing. Even if Glamor was updated to resolve the `node-fetch` vulnerability it would still be valuable to explore replacements -- this is just a good catalyst.

# Alternatives

Below are three libraries for consideration. There are others that also make use of the same API but these are the three I thought were most compelling. The suggestions below are purposefully minimalist; an example of a non-minimalist lib that also meets these requirements is [styled-components](https://styled-components.com/docs/api#css) (once you fish the relevant APIs out of their docs), however our current problem is largely a result of a lib with an unnecessarily large dependency tree so I have not included that here for comparison.

Needed features:

- Idempotent invocations.
  We call `css()` inside our `render()` methods in a few places. While we _should_ change that just to avoid the unnecessary performance overhead, it should also not cause any duplicate styles or memory leaks. Glamor uses deterministic hashing so we'd need something similar.

- Parent references.
  The '&' symbol must reference the parent style in a nested object. E.g.

  ```js
  css({
      backgroundColor: 'black',
      '& .foo': { color: 'white' },
  })
  // .asdf { background-color: black; }
  // .asdf .foo { color: white; }
  ```

- Child selectors, pseudo-selectors.
  Dictionary keys must allow complex CSS selectors and arbitrary nesting. E.g.:

  ```js
  css({
      '& > div': {
          '& > ul > li:hover': {
              cursor: 'pointer',
          },
      },
  })
  // .asdf > div > ul > li:hover { cursor: pointer; }
  ```

- Media queries.
  Namespaced media query support is required. E.g.:

  ```js
  css({
      '@media(min-width: 300px)': {
          '& h2': { fontSize: '30px' }
      }
  })
  ```

- Keyframes.
  Not required as there's not an obvious universal way to do it. I don't think we're currently making heavy use of keyframes but if we are this point of discussion should be re-prioritized. E.g.:

  ```js
  keyframes('bounce', {
      '0%': { transform: 'scale(0.1)', opacity: 0 },
      '60%': { transform: 'scale(1.2)', opacity: 1 },
      '100%': { transform: 'scale(1)' },
  })
  ```

- Global styles.
  Not required, as we can always add global styles the old fashioned way, but is a nice-to-have. The API for this _may_ vary between libraries -- Glamor uses a separate `global({...styles})` function -- because there's probably not an obvious universal way to do it. These calls should be infrequent enough that porting to a new API won't be too big a lift. E.g.:

  ```js
  global({
      fontFamily: `"Comic Sans MS", "Comic Sans", cursive`,
  })
  // font-family: "Comic Sans MS", "Comic Sans", cursive;
  ```

## [Typestyle](https://github.com/typestyle/typestyle)

```
% curl --compressed -Ssw '%{size_download}\n' -o /dev/null https://unpkg.com/typestyle@2.1.0/dist/umd/typestyle.min.js | numfmt --to=si
3.6K

% npm ls
└─┬ typestyle@2.1.0
  ├── csstype@2.6.9
  └── free-style@3.1.0
```

Meets all the requirements above and is identical to Glamor for that desired API subset. It does not have a `global()` function but it does have an identical `keyframes()` function. Like Glamor, it does have additional APIs that we will want to avoid. The project has a reasonable number of stars. The number and frequency of releases is high -- my initial impression is those are mostly benign and not a lot of API change but there is also a lot of changes around the extraneous API surface area.

Pros: Mostly drop-in replacement for Glamor. Mature & battle-tested. Half the size. Only two deps.

Cons: Larger API than we need/want. Some API churn.

## [free-style](https://github.com/blakeembrey/free-style)

```
% npm ls
└── free-style@4.1.0

% curl --compressed -Ssw '%{size_download}\n' -o /dev/null https://unpkg.com/free-style@4.1.0/dist/index.js | numfmt --to=si
3.1K (compressed but not minified)
```

Used by Typestyle above. It's a mature, tiny, featurefull library that we would create a wrapper around that has just the API surface area that we use and want to promote in our projects.

Pros: Tiny, no deps, mature, & battle tested. We could expose only the API that we want to use.

Cons: Would require us to create a maintain a small wrapper.

## [g-style](https://github.com/lucasrainett/g-style)

```
% curl --compressed -Ssw '%{size_download}\n' -o /dev/null https://unpkg.com/g-style@1.2.8/dist/index.umd.js | numfmt --to=si
1.6K

% npm ls
└── g-style@1.2.8
```

Meets the requirements subset detailed above.

Pros: Ridiculously tiny (263 SLOC). No deps. So small we could maintain it ourselves if necessary.

Cons: Brand new lib, not battle tested. Has an OOP API that we would need to ignore.

whiteinge:
Closes #895
Closes #897

### DIFF
--- a/docs/components/Changelog.js
+++ b/docs/components/Changelog.js
@@ -8,6 +8,11 @@ class Changelog extends React.Component {
 
         <h2>MX React Components V 8.0.0</h2>
 
+        <h3>8.4.0</h3>
+        <ul>
+          <li>Switch from Glamor to @mxenabled/cssinjs free-style wrapper. No user-facing changes expected.</li>
+        </ul>
+
         <h3>8.3.0</h3>
         <ul>
           <li>Updates moment package to fix a dependency security vulnerability</li>
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,10 +9,10 @@
       "version": "8.3.0",
       "license": "MIT",
       "dependencies": {
+        "@mxenabled/cssinjs": "^0.6.0",
         "bowser": "^1.9.4",
         "d3": "^3.5.6",
         "focus-trap-react": "^3.1.4",
-        "glamor": "^2.20.40",
         "keycode": "^2.1.8",
         "lodash": "^4.17.21",
         "moment": "^2.29.3",
@@ -3077,6 +3077,15 @@
         "node": ">= 6"
       }
     },
+    "node_modules/@mxenabled/cssinjs": {
+      "version": "0.6.0",
+      "resolved": "https://registry.npmjs.org/@mxenabled/cssinjs/-/cssinjs-0.6.0.tgz",
+      "integrity": "sha512-pj4/+iqyEGS1zg4i7jo7jLX+tS9h1YA5W1vLfUipuLWbxzEmoI6K2JvfhP1jT3eKTOJE3MTH+K53/t9KDyUu3Q==",
+      "dependencies": {
+        "free-style": "^4.1.0",
+        "insert-css": "^2.0.0"
+      }
+    },
     "node_modules/@nicolo-ribaudo/chokidar-2": {
       "version": "2.1.8-no-fsevents.2",
       "resolved": "https://registry.npmjs.org/@nicolo-ribaudo/chokidar-2/-/chokidar-2-2.1.8-no-fsevents.2.tgz",
@@ -3433,11 +3442,6 @@
       "funding": {
         "url": "https://github.com/sponsors/ljharb"
       }
-    },
-    "node_modules/asap": {
-      "version": "2.0.6",
-      "resolved": "https://registry.npmjs.org/asap/-/asap-2.0.6.tgz",
-      "integrity": "sha1-5QNHYR1+aQlDIIu9r+vLwvuGbUY="
     },
     "node_modules/asn1": {
       "version": "0.2.4",
@@ -4597,15 +4601,6 @@
         "node": ">=4.8"
       }
     },
-    "node_modules/css-in-js-utils": {
-      "version": "2.0.1",
-      "resolved": "https://registry.npmjs.org/css-in-js-utils/-/css-in-js-utils-2.0.1.tgz",
-      "integrity": "sha512-PJF0SpJT+WdbVVt0AOYp9C8GnuruRlL/UFW7932nLWmFLQTaWEzTBQEx7/hn4BuV+WON75iAViSUJLiU3PKbpA==",
-      "dependencies": {
-        "hyphenate-style-name": "^1.0.2",
-        "isobject": "^3.0.1"
-      }
-    },
     "node_modules/cssom": {
       "version": "0.3.8",
       "resolved": "https://registry.npmjs.org/cssom/-/cssom-0.3.8.tgz",
@@ -4852,14 +4847,6 @@
       "dev": true,
       "engines": {
         "node": ">= 4"
-      }
-    },
-    "node_modules/encoding": {
-      "version": "0.1.12",
-      "resolved": "https://registry.npmjs.org/encoding/-/encoding-0.1.12.tgz",
-      "integrity": "sha1-U4tm8+5izRq1HsMjgp0flIDHS+s=",
-      "dependencies": {
-        "iconv-lite": "~0.4.13"
       }
     },
     "node_modules/end-of-stream": {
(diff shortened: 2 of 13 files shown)


DrewWarrenTIY:
I think I broke your branch because I just merged my updates from yesterday in... Sorry! 😓

DrewWarrenTIY:
And just so it's clear - I updated to 8.3.0 in mine since it had a minor bump for moment. I think you'll want to go to 8.4.0 here now

DrewWarrenTIY:
I haven't used github in too long - was unable to inline review but in the messagebox component just confirming I'm reading correctly, it was setting border to grey 300 and then specifically passing in a color prop to the top border. Now with the shorthand you're passing color into border for the first value and the rest are being set to grey300. The only thing I can think of is if `color` is optional there then the old way would have defaulted to the grey 300. Not sure if there's a concern there or not.

DrewWarrenTIY:
Other than confirming the behavior in MessageBox - this looks really good. Thank you for taking the time to do this. If you don't think there's a problem with the border color then please consider this approval and a shipit :)

whiteinge:
Good question (stop me if I misunderstood). The top border was always overridden even when the `color` prop was omitted. There's a ['color' variable outside the diff](https://github.com/mxenabled/mx-react-components/blob/0d02e80514df20eee09a09e5d2a09af0dddbbf13/src/components/MessageBox.js#L76) so `props.color` and `color` are different and the former has a fallback value. So I think this has the same end-result.

DrewWarrenTIY:
Good call - I didn't see where `color` was falling back to the theme.Colors.PRIMARY value when defined. Should be good!

Ship it!

PR #902, opened 2021-03-26
Removes radium from the Drawer component

TayteStokes:
This change removes using radium to apply styles to the Drawer component and implements using glamor to apply those styles instead.

// @garrowp @DrewWarrenTIY @jacob-pettingill @phantomxc

### DIFF
--- a/src/components/Drawer.js
+++ b/src/components/Drawer.js
@@ -1,22 +1,22 @@
-const _get = require('lodash/get');
-const _isEqual = require('lodash/isEqual');
-const _isNumber = require('lodash/isNumber');
-const _merge = require('lodash/merge');
-const _throttle = require('lodash/throttle');
-const _uniqueId = require('lodash/uniqueId');
-const keycode = require('keycode');
-const PropTypes = require('prop-types');
-const React = require('react');
-const Velocity = require('velocity-animate');
-const { StyleRoot } = require('radium');
+const _get = require('lodash/get')
+const _isEqual = require('lodash/isEqual')
+const _isNumber = require('lodash/isNumber')
+const _merge = require('lodash/merge')
+const _throttle = require('lodash/throttle')
+const _uniqueId = require('lodash/uniqueId')
+const keycode = require('keycode')
+const PropTypes = require('prop-types')
+const React = require('react')
+const Velocity = require('velocity-animate')
+import { css } from 'glamor'
 
-import { withTheme } from './Theme';
-const Button = require('../components/Button');
-const MXFocusTrap = require('../components/MXFocusTrap');
+import { withTheme } from './Theme'
+const Button = require('../components/Button')
+const MXFocusTrap = require('../components/MXFocusTrap')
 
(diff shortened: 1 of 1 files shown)


src/components/Drawer.js
```diff
@@ -8,7 +8,7 @@ const keycode = require('keycode');
 const PropTypes = require('prop-types');
 const React = require('react');
 const Velocity = require('velocity-animate');
-const { StyleRoot } = require('radium');
+import { css } from 'glamor';
```
jacob-pettingill, line 11: Was @whiteinge trying to remove glamor from the OS components? Where did we get on that? Is there an package we are trying to get to?

jacob-pettingill:
Other than the Glamor question, this looks good to me! Do we have a screenshot or vid showing this working?

src/components/Drawer.js
```diff
@@ -227,7 +227,7 @@ class Drawer extends React.Component {
     const titleUniqueId = _uniqueId('mx-drawer-title-');
 
     return (
-      <StyleRoot>
+      <div>
```
phantomxc, line 230: Do we still need this div?

src/components/Drawer.js
jgerlach: I had the same question.

TayteStokes: I think that is the plan, but it hasn't been implemented yet.

src/components/Drawer.js
TayteStokes: Nope, I should have removed that. Just pushed that change.

TayteStokes:
@jacob-pettingill I provided screen recordings in the Gitlab issue, but here they are as well.

Before:

before

After:

after

phantomxc (approved):

src/components/Drawer.js
whiteinge: PR #900 makes the switch off Glamor. Looks like there's a new npm audit failure :roll_eyes: so I'll get that PR updated and then I think it's ready for prime-time.

That said, this PR looks fine and I don't see any reason the Glamor switch should hold this one up.

src/components/Drawer.js
```diff
@@ -240,12 +240,15 @@ class Drawer extends React.Component {
             <div
               aria-describedby={this.props['aria-describedby']}
               aria-labelledby={this.props['aria-labelledby'] || titleUniqueId}
+              className={css({ ...styles.component, ...this.props.style })}
```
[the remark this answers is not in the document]
TayteStokes: `style` is actually the name of the prop that we are using. If you take a look at the `FlexDrawer` component in Raja, you can see an example of how it's being used. Do you think this should be updated to `styles`? We would have to update everywhere that is passing styles through the style prop as well.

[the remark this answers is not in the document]
DrewWarrenTIY: we talked about this. It probably "should" be styles with an s, but is currently being passed the object in props as style without an s so we should leave it to minimize chances of anything breaking

PR #905, opened 2021-04-08
DateRangePicker | Pass displayValue as a param to onDateRangeSelect 

garrowp:
Add state in DateRangePicker to keep track of `displayValue` to pass into `onDateRangeSelect`. In MM, we're expecting to receive a 3rd parameter in `_handleDateRangeSelect`, which we end up passing to the DateRangePicker as `onDateRangeSelect`. We check for the value of `"This Month"` or `"Last Month"` to determine if the `dateRangeType` should be monthly or not. I added this piece of state to DateRangePicker so we could keep track of the `displayValue` and pass it into `onDateRangeSelect` when it is called.

Also updated the docs to specify that there is an extra argument that will be passed into `onDateRangeSelect`. The other two arguments are maintaining the same order as before.

#### Issue

https://gitlab.mx.com/mx/moneymap/-/issues/3023

### DIFF
--- a/docs/components/DateRangePickerDocs.js
+++ b/docs/components/DateRangePickerDocs.js
@@ -143,7 +143,7 @@ class DateRangePickerDocs extends React.Component {
         <p>A function to be called when the user presses the cancel button or clicks on the scrim to cancel the date selection. Can be useful for analytics.</p>
 
         <h5>onDateRangeSelect <label>Function</label></h5>
-        <p>A function to be called when the user has applied their selected date range. The range's start and end dates are passed as arguements to the call.</p>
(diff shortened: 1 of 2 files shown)


jacob-pettingill:
I'm guessing you tested this locally?

Also isn't there a way to test components in isolation in mx-react-components?

TayteStokes:
> I'm guessing you tested this locally?
> 
> Also isn't there a way to test components in isolation in mx-react-components?

Yeah, these can be tested out locally but most of the time they get tested through linking to Serenity.

TayteStokes (approved):
LGTM

Issue #896, opened 2021-01-20
Documenting audit failure in ini package


DrewWarrenTIY:
Currently when the Travis CI Pipeline runs there are 100 fo the same vulnerability detected that looks like this:

```
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Low           │ Prototype Pollution                                          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ ini                                                          │
├───────────────┼──────────────────────���───────────────────────────────────────┤
│ Dependency of │ webpack-dev-server [dev]                                     │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ webpack-dev-server > chokidar > fsevents > node-pre-gyp > rc │
│               │ > ini                                                        │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://nodesecurity.io/advisories/1589                      │
└───────────────┴──────────────────────────────────────────────────────────────┘
```

There is currently a PR submitted for the `rc` package here: 

https://github.com/dominictarr/rc/pull/121

Since the assessment is `low` and there is an upstream PR open, we are comfortable 'ignoring' this for now. Hopefully it is resolved by the maintainer of `rc` soon.

PR #914, opened 2021-05-20
DisplalyInput | Replaced radium with glamor

garrowp:
Replace radium with glamor in the DisplayInput component. Relates to some issues we were having with Q2 custom user agent and Radium.

I'm using glamor here because we have a couple of instances in our styling where we're using pseudo selectors (focus). With radium, they would get applied when any of the elements inside the selector were focused. However, that is not equivalent for when we're using glamor. To get the same effect, we can use the `focus-within` selector, but it's not supported by IE 11. MoneyMap is close to ending support for IE 11 (August), so this may be a moot point soon.

If anyone has any ideas that aren't too heavy of a lift to support IE 11 while removing radium and maintaining the same styles, I'd be happy to hear them out.

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -5995,8 +5995,9 @@
           "optional": true
         },
         "ini": {
-          "version": "1.3.5",
-          "resolved": "",
+          "version": "1.3.6",
+          "resolved": "https://registry.npmjs.org/ini/-/ini-1.3.6.tgz",
+          "integrity": "sha512-IZUoxEjNjubzrmvzZU4lKP7OnYmX72XRl3sqkfJhBKweKi5rnGi5+IUdlj/H1M+Ip5JQ1WzaDMOBRY90Ajc5jg==",
           "optional": true
         },
         "is-fullwidth-code-point": {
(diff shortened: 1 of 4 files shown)


DrewWarrenTIY:
I'm seeing the pipeline reporting a ton of "High" severity audit failures. I know we had a few we were ignoring like from https://github.com/mxenabled/mx-react-components/issues/896 but these seem to be new. Do you have any thoughts about whether we need to do something about that?

The code looks good, IMO

garrowp:
> I'm seeing the pipeline reporting a ton of "High" severity audit failures. I know we had a few we were ignoring like from #896 but these seem to be new. Do you have any thoughts about whether we need to do something about that?

It looks like there have been quite a few dependabot PRs open in the past 2 weeks or so. My assumption is that there were some vulnerabilities that triggered both the dependabot and the new audit failures. 

We've just been ignoring them for now. Do you think we should start fixing this audit failures and/or merging these dependabot PRs so our pipeline is cleaned up?

DrewWarrenTIY:
Awesome, much better. Only "Low" vulnerabilities left. Thanks for doing that.

:shipit:

PR #917, opened 2021-05-25
Fixes Axe auditing issue when using role="header" on the Selector component

TayteStokes:
This change updates the Selector component to also use the attribute of aria-level where we are setting the elements role attribute to header.

This will resolve issue: https://gitlab.mx.com/mx/money-experiences/moneymap/moneymap-issues/-/issues/3215


garrowp (approved):
LGTM!

PR #918, opened 2021-05-25
Updates the aria-controls being used in DateRangerPicker

TayteStokes:
This change adds a conditional to only apply the value of "calendarMenu" to the buttons aria-controls when the element with the corresponding id is available.

This will resolve https://gitlab.mx.com/mx/money-experiences/moneymap/moneymap-issues/-/issues/3214


DrewWarrenTIY:
The CI pipeline failures don't look to match the "Low" severity ones that we have been okay with ignoring. Would you try to update some of the packages and see if those warnings go away? peter just did this a few days ago but it looks like a new vulnerability or two crept in since then

DrewWarrenTIY:
Other than the pipeline, the code looks good. Let's see what we can do about that and then if we find resolution this is a :shipit: from me :)

TayteStokes:
@DrewWarrenTIY Sorry, these PR's kind of got lost at sea while I have been focusing on a few other things in GitLab. I'm wondering if maybe we should create a separate PR to update those packages. Thoughts around that?

garrowp (approved):
👍🏼

PR #920, opened 2021-06-03
Get end of last day of last year, not the beginning of the last day of last year

garrowp:
Rather than using `startOf('year').subtract(1, 'd')`, I switched it to `endOf('year').subtract(1, 'y')` so that it would get the end of the last day of last year rather than the beginning of the last day of last year.


TayteStokes (approved):
LGTM 👍 

It looks like there's a few packages that we need to go through and update due to vulnerabilities, but that can be handled in another PR.

PR #921, opened 2021-06-11
Drewwarren/update date picker text color a11y

DrewWarrenTIY:
A few instances of text in our date picker were consistently failing an axe audit due to contrast. I've updated those so that the ones that were not brand color related would always pass:

* the greyed out "carry over" dates from last month and next month on the first and last calendar rows
* the "current day" when not selected was white on light gray - is now dark gray on light gray

Also the likelihood that our 50% transparency applied to a brand color will fail is much much lower as I've changed the white text to dark grey.

Screenshots:

Before: 

Screen Shot 2021-06-11 at 11 15 33 AM

After:

Screen Shot 2021-06-11 at 11 15 07 AM

Also I had to run `npm audit fix` for some pipeline vulnerabilities but it looks safe from what I can tell, updating `babel-cli` shouldn't be a risky package I don't think but if you're reviewing this please take a look :)

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -4,20 +4,148 @@
   "lockfileVersion": 1,
   "dependencies": {
     "@babel/cli": {
-      "version": "7.10.1",
-      "resolved": "https://registry.npmjs.org/@babel/cli/-/cli-7.10.1.tgz",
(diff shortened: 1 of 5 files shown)


TayteStokes (approved):
Sweet, this looks good! I don't think we should worry too much about the minor change for babel/cli, but let's keep an eye out for anything.

garrowp:
LGTM 👍🏼

PR #922, opened 2021-06-24
DateRangePicker | Use end of day for selected end date

garrowp:
#### Description

We have an issue in MoneyMap with the date range picker showing the date before the selected end date. https://gitlab.mx.com/mx/money-experiences/moneymap/moneymap-issues/-/issues/3328#note_1482946

#### Solution

We've been switching to date-fns in MoneyMap over moment recently. This has caused some issues with conversion with timezones. Because of this, we need to make sure that the end date that is getting passed from the mx-react-components `DateRangePicker` is actually the end of the day instead of the start of the day. My changes just use moment to pass the unix timestamp for the end of the day.

I've tested this in the date range picker for the both the Spending widget and the Transactions widget and the dates have been showing up correctly


DrewWarrenTIY:
Looks good to me! :shipit:

PR #948, opened 2022-06-02
Fix deps

whiteinge:
There was a peer dep conflict that CI didn't catch when I upgraded npm to the 7+ format. This resolves that so `npm ci` and `npm i` run correctly.


DrewWarrenTIY:
I don't see anything concerning here! Should be good to ship.

PR #949, opened 2022-06-10
Fix missing parent selector syntax for pseudo-classes

whiteinge:
Radium also uses the `:hover` syntax so determining which styles are
passed to Radium and which are passed to cssinjs is a little tricky.
Hopefully I got this right...

### DIFF
--- a/docs/components/MessageBoxDocs.js
+++ b/docs/components/MessageBoxDocs.js
@@ -115,12 +115,12 @@ class MessageBoxDocs extends React.Component {
         border: Styles.Colors.GRAY_300,
         color: Styles.Colors.GRAY_900,
 
-        ':hover': {
+        '&:hover': {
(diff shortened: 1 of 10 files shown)


whiteinge:
@DrewWarrenTIY I _think_ this fixes the hover states. I'm testing by building the docs rather than in an upstream project so I would appreciate a second pair of :eyes: (none of my regularly-scheduled upstream projects use these components anymore) -- but I'd be (more than) happy to pair on pulling this MR down and testing it in a real project locally.

The second commit is the same peer dep fix as last week so this is best reviewed by-commit rather than by-PR.

DrewWarrenTIY:
Code looks good. I checked this branch out locally and linked it to `serenity` and it looks like the hover and focus styles are applying correctly. I think this is good to ship and publish! Thanks!

PR #954, opened 2022-08-01
update some hover and focus styles that were not correctly applied po…

DrewWarrenTIY:
The hover styles were not applying to the "Select a Date Range" section of the Date Range Picker component. This update fixes that.

Screen Shot 2022-08-01 at 1 07 11 PM

(very feint grey on hover)


DrewWarrenTIY:
This is all related to removing `glamor` in favor of `cssinjs`

smstamm:
As discussed in Slack, we don't need to import `css` from @mxenabled/cssinjs here since the styles are passed as props from another component that *does* import that.

This looks good to me! 🚀

PR #988, opened 2023-10-18
Add localization to DateRangePicker

smstamm:
Issue: https://mxcom.atlassian.net/browse/MEW-452

#### Description
* Adds a required `getTranslation()` function prop to allow the consumer to translate strings

#### Testing
This can't really be tested until it's integrated into Serenity, but once that's complete, the visible text, aria and dates should all be localized to Spanish (and French)

### DIFF
--- a/src/components/DateRangePicker/SelectionPane.js
+++ b/src/components/DateRangePicker/SelectionPane.js
@@ -16,6 +16,8 @@ class SelectionPane extends React.Component {
     defaultRanges: PropTypes.array,
     getFromButtonRef: PropTypes.func,
     getToButtonRef: PropTypes.func,
+    getTranslation: PropTypes.func,
+    locale: PropTypes.string,
     onDateBoxClick: PropTypes.func,
     selectedBox: PropTypes.string,
     selectedEndDate: PropTypes.number,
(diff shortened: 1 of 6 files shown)


src/components/DateRangePicker/SelectionPane.js
```diff
@@ -24,43 +26,53 @@ class SelectionPane extends React.Component {
...
           >
-            {selectedStartDate ? moment.unix(selectedStartDate).format('MMM D, YYYY') : 'Select Start Date'}
+            {selectedStartDate ? this._formatDate(selectedStartDate) : getTranslation('Select Start Date')}
           </button>
 
-          <label style={styles.boxLabel}>To</label>
+          <label style={styles.boxLabel}>{getTranslation('To')}</label>
           <button
-            aria-label={`Select End Date, ${selectedEndDate ? 'Current end date is ' + moment.unix(selectedEndDate).format('MMM D, YYYY') : ''}`}
+            aria-label={getTranslation(`Select End Date${selectedEndDate ? ', Current end date is' : ''} %1`, this._formatDate(selectedEndDate))}
```
jgerlach, line 64: Does this need a space before `selectedEndDate`? Not sure if the screenreader will read it out funny with it smashed together or not.

smstamm: I smooshed it because sometimes that string `Current end date` is there, and sometimes it isn't. I moved the space and the comma to inside that optional string, so that the screen reader won't show a hanging comma. 

previous without extra string:  `Select End Date, ` (note comma and extra space, even though this string is whole)
previous with extra string: `Select End Date, Current end ...`

updated without extra string: `Select End Date`
updated with extra string: `Select End Date, Current end ...`

csconly:
This MR looks good to me

PR #990, opened 2023-10-19
Fix missed strings for localization

smstamm:
Issue: https://mxcom.atlassian.net/browse/MEW-452

#### Description
* Fixes translation for previously-missed aria labels
image

#### Testing
This can't really be tested until it's integrated into Serenity, but once that's complete, the visible text, aria and dates should all be localized to Spanish (and French)


jgerlach:
Looks good to me.

PR #995, opened 2023-11-27
Calendar | translates strings, uses localized dates

smstamm:
Issue: https://mxcom.atlassian.net/browse/MEW-562

#### Description
* Translates all calendar-related strings
* Localizes dates based on language

| language | before | after |
| ------ | ------ | ------ |
|    Spanish    | image |  image  |
|    French    | image    |  image |
| English |  image |  image |

#### Testing (must be completed in Serenity)
1. Load Serenity > Transactions with the client profile language set to `en-US`, `es` and `fr-CA`
2. Click a transaction to open the drawer
3. Click the Date row to open the calendar
4. Ensure the calendar's days of the week and all related text is correctly localized for different languages
5. Check that Voice Over announces everything in the correct language

### DIFF
--- a/src/components/Calendar.js
+++ b/src/components/Calendar.js
@@ -47,6 +47,7 @@ export const getNewDateStateChange = ({
 
 class Calendar extends React.Component {
   static propTypes = {
+    getTranslation: PropTypes.func,
     locale: PropTypes.string,
     minimumDate: PropTypes.number,
     onDateSelect: PropTypes.func,
@@ -66,6 +67,10 @@ class Calendar extends React.Component {
     focusedDay:
       this.props.selectedDate || this.props.minimumDate || moment().unix()
   };
+
+  componentDidMount() {
+    moment.locale(this.props.locale)
+  }
 
   UNSAFE_componentWillReceiveProps (newProps) {
     if (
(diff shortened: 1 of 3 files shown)


src/components/Calendar.js
```diff
@@ -67,6 +68,10 @@ class Calendar extends React.Component {
       this.props.selectedDate || this.props.minimumDate || moment().unix()
   };
 
+  componentDidMount() {
+    moment.locale(this.props.locale)
```
jgerlach, line 72: I'm assuming we don't need to import the locale files from moment because that is happening in `raja` now?

smstamm: Unfortunately we do. The locale does not get picked up in mx-react-components from Serenity, strangely. But Serenity _does_ pick up the locale from mx-react-components, which is why we needed to set that at the top level in Serenity.

smstamm: Sorry, I misunderstood your initial question - I thought it was about `componentDidMount()`. I assume we don't need to import those, because it's all working... 🤔

jgerlach: Ok. In that case I think this MR is good to go.

